PR #1, opened 2021-05-08
Fix markdown linting issues

munichpavel:
Just some small things that showed up in my editor (VS Code) with markdown linting enabled.

* MD025/single-title/single-h1: Multiple top-level headings in the same document
* MD009/no-trailing-spaces: Trailing spaces

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,33 +2,33 @@
 
 ----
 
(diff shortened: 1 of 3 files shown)


munichpavel:
I am not sure if I should add to the change log and bump the version, or if you want to do that.

richardangell (approved):

PR #10, opened 2021-08-23
NearestImputer Update : Remove median imputer when no null

ClaireF57:
Use Case : When a column has no null when fitting NearestMeandResponseImputer 
- Previously: have the possibility to have a median imputer 
- Now: This triggers an error. Goal is that in case there is no null, the NearestMeandResponseImputer is not the right one to use but directly a median imputer. 

Code update : 
- tubular/imputers.py : removed median imputation
- tests/imputers/test_NearestMeandResponseImputer.py : removed the test concerning use case with median imputation , updated other test to remove default init concerning the use of median imputer as default
- example/imputers/NearestMeandResponseImputer.ipyng: removed the code associated to the use of median imputer in notebook and put the example to "deprecated"


richardangell (changes requested):
all looks good, thanks, just very minor tidying of github action file needed

PR #13, opened 2021-09-27
modified GroupRareLevelsTransformer

bissoligiulia:
We modified the transformer in such a way the transformation returns the columns selected without changing in type. The transformation works for int, float, datetime, str,... columns.

- Replaced the constraint type of rare_level_name==string with type of rare_level_name==type of columns selected
- Modified and added tests
- Modified the .add_categories() for categorical columns. Now the transformer works even for pandas 1.3.0+


tests/nominal/test_GroupRareLevelsTransformer.py
```diff
@@ -234,6 +234,54 @@ def test_learnt_values_weight_2(self):
...
+
+            x = GroupRareLevelsTransformer(columns = ['c'])
+            
+            x.fit(df)
+
+    def test_rare_level_name_same_col_type(self):
+        """Test that checks if output columns are of the same type with respect to the input label."""
+
+        df = d.create_df_10()
+
```
richardangell, line 258: you can avoid repetition in this test by using `pytest.mark.parametrize` e.g. 

```
@pytest.mark.parametrize(
     "label,col",
     [
          (2.0, "a"),
          ("zzzz", "b"),
          ....
    ]
)
def test_rare_level_name_same_col_type(self, label,col):

    df = d.create_df_10()
    x = GroupRareLevelsTransformer(columns = [col], rare_level_name=label)
    x.fit(df)
    df_2 = x.transform(df)
    assert pd.Series(label).dtype == df_2[col].dtypes, "column type should be the same as label type"
```

richardangell (changes requested):

richardangell (approved):

PR #26, opened 2021-11-07
Feature/separate x y

richardangell:
Update relevant transformers to use `X` and `y` in the `fit` methods rather than passing a `response_column` variable when initialising the objects


nedwebster (approved):
Everything looks good!

PR #28, opened 2021-11-08
Updated test to no longer use ordered arg in pd.cut

nedwebster:
Updated `tests/numeric/test_CutTransformer.py::TestTransform::test_expected_output` to no longer use the "ordered" argument for `pd.cut()`.

This test is now compatible with pandas versions <1.1.0.


richardangell (approved):
Thanks @nedwebster !

PR #30, opened 2021-11-09
Feature/attribute fix

lsumption:
* Define attributes for date and numeric transformers to allow printing
* Fix unit tests with new attributes
* Add tests for printing and sklearn.base.clone for all transformers


richardangell (approved):
Thanks @lsumption looks good!

PR #31, opened 2021-11-09
Feature/version change 0.3.1

merve-alanyali:
- Removed pytest and pytest-mock from requirements.txt
- Fixed min pandas version as 1.0.0 in requirements.txt, requirements-dev.txt and docs/requirements.txt
- Updated version from 0.3.0 to 0.3.1


richardangell (approved):
Thanks @merve-alanyali !

PR #38, opened 2021-12-16
Interaction transformer

ClaireF57:
- create interactionTransformer class in numeric.py
- create test and example for interaction transformer
- Update forked repo, add pre-commit


richardangell (approved):
Thanks!

PR #41, opened 2022-05-27
add list_transformers call

nedwebster:
Draft Pull Request

Contains initial code structure for the `list_transformers()` call. I had three suggestions for how to code the call:

1) Raw string manually coded
- Pro: Can allow for full flexibility
- Con: Lengthy to do, quite ugly, and would require a lot of upkeep for every new transformer that gets added

2) Looping through nested dirs and applying `eval()`
- Pro: Full flexibility as well as automated updating
- Con: Requires use of unsafe python `eval()` call

3) Current approach, manually adding modules but automatically grabbing transformers
- Pro: Semi-automated and can prompt warnings if new modules are added and forgotten about here
- Con: Without using `eval()` I'm not sure there is an easy way to get args/descriptions for the transformers

I would be interested to know what your opinions are. Once we land on a solution, I can flesh the code out and add unit tests.

### DIFF
--- a/tubular/__init__.py
+++ b/tubular/__init__.py
@@ -8,3 +8,5 @@ from tubular import misc
 from tubular import nominal
 from tubular import numeric
 from tubular import strings
+
+from tubular.helpers import list_transformers
--- a/tubular/helpers.py
+++ b/tubular/helpers.py
@@ -1,0 +1,88 @@
+import tubular
+from tubular import base
+from tubular import capping
+from tubular import dates
+from tubular import imputers
+from tubular import mapping
+from tubular import misc
+from tubular import nominal
+from tubular import numeric
+from tubular import strings
+
+from pprint import pprint
+
+
+class Helper:
+
+    def __init__(self):
+
+        self.module_dict = {
+            "base": self._format_module_transformers(base),
+            "capping": self._format_module_transformers(capping),
+            "dates": self._format_module_transformers(dates),
+            "imputers": self._format_module_transformers(imputers),
+            "mapping": self._format_module_transformers(mapping),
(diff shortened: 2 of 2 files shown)


DWdapengwang:
Hi @nedwebster,   thank you for the suggestions. 

I agree, option 3 sounds the best. The main aim for this functionality is too allow the user to quickly find a list of transformers and where to import them. And then if more details for the transformer is needed, then it is easy to know where to find it.  

Option 1 would be too manual, if user wants that much detail, then there is always readthedoc, or the code. 
Option 2 using eval() would be risky, so it is not a viable option.

adamsardar:
What's wrong with relying on the [documentation site](https://tubular.readthedocs.io/en/latest/api.html#mapping-module)? It's automatically generated, contains arguments (alongside their typehints eventually) and will continue to be extended. What is the expected user behaviour where a list of all of the transformers is needed? The docs will almost always be the best port of call, with IDE autocomplete a close second.

This is a very old PR and I would probably suggest closing it unless a clear user story for the feature is present?

davidhopkinson26:
> What's wrong with relying on the [documentation site](https://tubular.readthedocs.io/en/latest/api.html#mapping-module)? It's automatically generated, contains arguments (alongside their typehints eventually) and will continue to be extended. What is the expected user behaviour where a list of all of the transformers is needed? The docs will almost always be the best port of call, with IDE autocomplete a close second.
> 
> This is a very old PR and I would probably suggest closing it unless a clear user story for the feature is present?

I agree.  Closing down as a user can use the documentation/ IDE autocomplete and this has been stale for a long time.

PR #44, opened 2022-11-17
Fixes #43: Debug kwargs removal

### LINKED ISSUE #43: [Bug]: Unexpected arguments are passed silently without raising errors
### What happened?

While using the 'nominal.py::GroupRareLevelsTransformer' we had the problem that when the 'columns' argument is not passed, the transformer takes automatically all columns of type  'object 'or 'category', which is fine. However, if for some reason there was a typo in the 'columns' attribute (for instance 'coulumns'), the transformer does not raise any error regarding unknown/unexpected argument, but rather goes with the automatic behaviour explained above. This generate error going silent if anything happens and can generate quite nasty unexpected behaviours.

While looking a the parent classes, one can see that its parents do not expect any other column.

* BaseNominalTransformer: does not defines an __init__, so parameters go to parents
* BaseTransformer: it does expect argument 'columns' among others, further arguments go to the parents
* TransformerMixin: does not defines an __init__, no parent
* BaseEstimator: does not defines an __init__, no parent

That means that one could stop the bug from happening, if the argument '**kwargs' is removed from the class 'BaseTransformer' which is the last in the inheritance that accepts some arguments, also it does not cal the __init__ of the super at any moment. This would cause that all unexpected arguments would raise errors. https://github.com/lvgig/tubular/blob/f54424948608a1d8cc489aeefaef75ce7a602c53/tubular/base.py#L15

That would solve the problem with the 'GroupRareLevelsTransformer'. However, it would make several other tests fail, since there are many calls to some classes where non expected arguments are passed. For instance, in https://github.com/lvgig/tubular/blob/f54424948608a1d8cc489aeefaef75ce7a602c53/tests/test_transformers.py#L49 there is the call to 'NearestMeanResponseImputer' with an argument that it dies not expect: 'response_column'.
After fixing that, there are other 19 test that fail, since several tests pass unexpected arguments at some moment or another.

### Environment

Python 3.9.5
--- 
pytest==7.1.2
pandas>=1.0.0
scikit-learn>=1.1.1
tubular == 0.3.2

### Minimum reproducible code

```shell
# This code should raise an error, note the argument 'im_not_expected', does not break the code.

def test_tubular_args():
    with pytest.raises(TypeError) as excinfo:
        GroupRareLevelsTransformer(im_not_expected=['a', 'b', 'c'])
```

### Relevant error output

```shell
FAILED test_tubular_args.py::test_tubular_args - Failed: DID NOT RAISE <class 'TypeError'>
```

### Code of Conduct

cagonza6:
Fixes issue #43 where 'base.BaseTransformer' was accepting kwargs even when its parent classes did not expect any. This was propagating as a bug where any unexpected/wrong argument passed to child classes did not raise any error.

There were some unit tests that where failing because of old used arguments that went silently and one that even had a wrong argument passed to.

Some notes on the PR:

* By executing black I had an issue and had to update to black==22.3.0.  Since the contributions guide says that there should be one feature per PR
* I did not fix any version in the Changelog, since I'm not sure when you plan to make a release and also because of the previous point.
* There were few places when lists and dictionaries where defined as default parameters in the __init__ functions, so I changed them for None and and changing to the original one in the code. That might generate problems with mutation in some cases.

Overall it was fun to solve, but a bit too long. A recommendation would be to remove the possibility to have the arguments 'columns', 'copy' and 'verbose' from the 'BaseTransformer', but that is just taste/design choice and not a bug or issue.

### DIFF
--- a/tubular/base.py
+++ b/tubular/base.py
@@ -13,7 +13,7 @@ from tubular._version import __version__
 
 
 class BaseTransformer(TransformerMixin, BaseEstimator):
-    """Base tranformer class which all other transformers in the package inherit from.
+    """Base transformer class which all other transformers in the package inherit from.
 
     Provides fit and transform methods (required by sklearn transformers), simple input checking
     and functionality to copy X prior to transform.
@@ -25,22 +25,19 @@ class BaseTransformer(TransformerMixin, BaseEstimator):
         in columns is saved in the columns attribute on the object.
 
     copy : bool, default = True
-        Should X be copied before tansforms are applied?
+        Should X be copied before transforms are applied?
 
     verbose : bool, default = False
         Should statements be printed when methods are run?
 
-    **kwds
-        Arbitrary keyword arguments.
-
     Attributes
     ----------
     columns : list or None
-        Either a list of str values giving which columns in a input pandas.DataFrame the transformer
+        Either a list of str values giving which columns in an input pandas.DataFrame the transformer
         will be applied to - or None.
 
     copy : bool
-        Should X be copied before tansforms are applied?
+        Should X be copied before transforms are applied?
 
     verbose : bool
         Print statements to show which methods are being run or not.
@@ -50,7 +47,7 @@ class BaseTransformer(TransformerMixin, BaseEstimator):
 
     """
 
-    def __init__(self, columns=None, copy=True, verbose=False, **kwargs):
+    def __init__(self, columns=None, copy=True, verbose=False):
 
         self.version_ = __version__
 
@@ -189,9 +186,11 @@ class BaseTransformer(TransformerMixin, BaseEstimator):
         return X_y
 
     def transform(self, X):
-        """Base transformer transform method; checks X type (pandas DataFrame only) and copies data if requested.
+        """Base transformer transform method; checks X type (pandas DataFrame only)
+        and copies data if requested.
 
-        Transform calls the columns_check method which will check columns in columns attribute are in X.
+        Transform calls the columns_check method which will check columns in
+        columns attribute are in X.
 
         Parameters
         ----------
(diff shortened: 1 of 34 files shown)


cagonza6:
Hi,
I think I'm doing something very wrong, so I prefer to ask

- Should this PR go to the development branch? 
- Since the test are crucial part of the PR, should I update the required black version as part of the PR? If not, the main branch will be broken.

davidhopkinson26:
Hi Cristian,

Thank you very much for the contribution! Yes,  please make the PR into develop rather than main.  I have made another PR into develop to fix the black issue separately #45.

On versioning -  we will probably put this in a small release with #45  which will be 0.3.3.  Version 0.4 is however likely to be ready in December with lots of new transformers so we may roll in these changes into a single release.  For the purposes of this PR please set version to 0.3.3 and update the change log with this too for now.

You went a bit further than I expected by removing kwargs from all transformers and changing to explicitly pass copy and verbose.  For the issue in #43 need it would be sufficient to remove **kwargs from BaseTransformer only and continue to have those two arguments passed using **kwargs from its children.  If unexpected args were passed they would still be picked up by BaseTransformer and raise an error.  Apologies if this wan't clear in my comments on the issue.

This structure was intended to help with maintainability since it allows us to add new options at the BaseTransformer level without needing to change every inheriting transformer and their tests.  It also reduces clutter. This is however at a slight trade off with ease of use for these two features, and the code is possibly harder to understand.

I can see you have put a lot of work into the changes so I am keen to hear your thoughts on removing **kwargs just from the base transformer or from all children too before we decide which direction to take here.

tubular/base.py
```diff
@@ -367,12 +369,16 @@ def __init__(
         new_column_name,
         pd_method_name,
         columns,
-        pd_method_kwargs={},
+        pd_method_kwargs=None,
```
davidhopkinson26, line 372: Can you explain your reasoning for changing this?

davidhopkinson26:
Hi, thanks for your contribution here!  It has helped identify a number of tests which were not functioning properly.  I have taken the key components of the fix into #56 and merged into develop for the next release.  Closing this for now.

PR #45, opened 2022-11-22
Fixes #45: updated black version to 22.3.0 to fix compatibility issue with click

davidhopkinson26:
updated black version to 22.3.0 to fix compatibility issue with click 8.1.0.  Without this fix black fails with:
`ImportError: cannot import name '_unicodefun' from 'click'`


davidhopkinson26:
Also updated flake8 as the old version was throwing an error in the build pipeline

Sarah-TaylorKnight (approved):
Looks good

PR #48, opened 2022-12-06
prefix errors with transformer classname

davidhopkinson26:
Added classname() method to BaseTransformer
Prefixed all errors with classname call for easier debugging


Ritunjai-Sharma (approved):
Reviewed Changes! Looks good!

PR #50, opened 2022-12-08
feature/DatetimeSinusoidCalculator

davidhopkinson26:
A transformer that calculates sin or cos of any time unit.

### DIFF
--- a/tests/dates/test_DatetimeSinusoidCalculator.py
+++ b/tests/dates/test_DatetimeSinusoidCalculator.py
@@ -1,0 +1,326 @@
+import pytest
+import tests.test_data as d
+import test_aide as ta
+import re
+
+import tubular
+from tubular.dates import DatetimeSinusoidCalculator
+import pandas as pd
+import numpy as np
+
+
+@pytest.fixture(scope="module", autouse=True)
+def example_transformer():
+
+    return DatetimeSinusoidCalculator("a", "cos", "hour", 24)
+
+
+class TestDatetimeSinusoidCalculatorInit(object):
+    """Tests for DateDifferenceTransformer.init()."""
+
+    def test_arguments(self):
+        """Test that init has expected arguments."""
+
+        ta.functions.test_function_arguments(
+            func=DatetimeSinusoidCalculator.__init__,
+            expected_arguments=[
+                "self",
+                "columns",
+                "method",
+                "units",
+                "period",
+            ],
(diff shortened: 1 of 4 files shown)


tubular/dates.py
```diff
@@ -663,3 +665,152 @@ def transform(self, X):
...
+            raise TypeError(
+                "{}: units must be a string but got {}".format(
+                    self.classname(), type(units)
+                )
+            )
+
+        if (
+            (not isinstance(period, int))
+            and (not isinstance(period, float))
+            or isinstance(period, bool)
```
TommyMatthews, line 736: Think the bool option here is left over from when I had the option to set `period=False`

tubular/dates.py
```diff
@@ -663,3 +665,152 @@ def transform(self, X):
         X[self.new_column_name] = lower_comparison & upper_comparison
 
         return X
+
+
+class DatetimeSinusoidCalculator(BaseTransformer):
+
+    """
+    Transformer to derive a feature in a dataframe by calculating the
```
TommyMatthews, line 673: Looking at this transformer again, I think it seems a bit inconsistent that you can specify multiple input columns but only one unit or period - might be a bit much for now but could be worth adding a refactor to the backlog (e.g. to dictionaries or lists that are assumed to have matching inputs)

tests/dates/test_DatetimeSinusoidCalculator.py
```diff
@@ -0,0 +1,326 @@
...
+                "columns",
+                "method",
+                "units",
+                "period",
+            ],
+            expected_default_values=(2 * np.pi,),
+        )
+
+    def test_class_methods(self, example_transformer):
+        """Test that DateDifferenceTransformer has a transform methods."""
```
TommyMatthews, line 37: method

davidhopkinson26: https://github.com/lvgig/tubular/pull/50/commits/eff8469d47e048a8634eb8227f5491c0a048c531

tubular/dates.py
davidhopkinson26: removed this line and relevant test case: 
https://github.com/lvgig/tubular/pull/50/commits/eff8469d47e048a8634eb8227f5491c0a048c531

TommyMatthews (approved):

PR #54, opened 2022-12-13
feature/warning to mapping transformer for unexpected dtype changes

davidhopkinson26:
Due to the way pd.Series.map works, mappings can result in column dtypes changing, sometimes unexpectedly. If the result of the mappings is a dtype that doesn't match the original dtype, or the dtype of the values provided in the mapping a warning will be raised. This normally results from an incomplete mapping being provided, or a mix of dtypes causing pandas to default to the object dtype.

For columns with a 'category' dtype, conventional type comparisons do not work as they category dtypes reflect the possible values on the category. This means that although a mapping can maintain a category dtype, if the values in the column change it will be registered as a different dtype. To address this a check is performed before the warning is raised to ensure that it does not get raised in these cases.


tubular/mapping.py
```diff
@@ -163,13 +166,25 @@ def __init__(self, mappings, **kwargs):
...
+        Due to the way pd.Series.map works, mappings can result in column dtypes changing,
+        sometimes unexpectedly. If the result of the mappings is a dtype that doesn't match
+        the original dtype, or the dtype of the values provided in the mapping a warning
+        will be raised. This normally results from an incomplete mapping being provided,
+        or a mix of dtypes causing pandas to default to the object dtype.
+
+        For columns with a 'category' dtype, conventional type comparisons do not work as they
+        category dtypes reflect the possible values on the category. This means that although
+        a mapping can maintain a category dtype, if the values in the column change it will
+        be registered as a different dtype. To address this a check is performed before the
```
TommyMatthews, line 185: I think just: "The warning will not be raised in this case" is clearer - this description led me to look for a pre-transform check for category dtype

tubular/mapping.py
```diff
@@ -181,9 +196,49 @@ def transform(self, X):
...
+
+                warnings.warn(
+                    f"{self.classname()}: There are values in the mapping for {col} that are not present in the dataframe"
+                )
 
         X = BaseMappingTransformMixin.transform(self, X)
 
+        mapped_dtypes = X[mapped_columns].dtypes
+
+        for col in mapped_columns:
```
TommyMatthews, line 223: I would add a `suppress_dtype_warning` parameter (default `False`) to this transformer - if you're applying a mapping from e.g. integer to bool, you know that the dtype is going to change and this warning would be quite annoying

tubular/mapping.py
```diff
@@ -163,13 +166,25 @@ def __init__(self, mappings, **kwargs):
         BaseMappingTransformer.__init__(self, mappings=mappings, **kwargs)
 
     def transform(self, X):
```
TommyMatthews, line 168: One for the backlog probably but should all of these checks also be applied to the CrossColumnMapping transformer? and perhaps the others in mapping.py? (or put into BaseMappingTransformer)

davidhopkinson26: good point, will put that on the list

tubular/mapping.py
davidhopkinson26: good call, have added this in [e9c8d66](https://github.com/lvgig/tubular/pull/54/commits/e9c8d6687626a87102295d53ffb1f34ccd916362)

tubular/mapping.py
davidhopkinson26: agreed and changed to suggested simpler explanation in [4632f52](https://github.com/lvgig/tubular/pull/54/commits/4632f527a938d1962110eda5af823b52e7434caa)

TommyMatthews (approved):

PR #56, opened 2022-12-13
Fixes #43: Remove kwargs from base transformer.  init  

### LINKED ISSUE #43: [Bug]: Unexpected arguments are passed silently without raising errors

davidhopkinson26:
Removing **kwargs argument from BaseTransformer so that unexpected arguments passed to it will throw an error.  

without fix doing may result in a silent error and unexpected behavior if an argument is misspelt.  Revealed several tests in need of fixes due to incorrect arguments being used.

credit to @cagonza6 for identifying the issue and making initial PR in #44.  Recreating minimum fix here due to lack of activity on that PR.


tubular/base.py
```diff
@@ -50,7 +50,7 @@ class BaseTransformer(TransformerMixin, BaseEstimator):
 
     """
 
-    def __init__(self, columns=None, copy=True, verbose=False, **kwargs):
+    def __init__(self, columns=None, copy=True, verbose=False):
```
Ritunjai-Sharma, line 57: Since we have decided to do away with **kwargs, shouldn't the above docstring be updated to not have **kwds Arbitrary keyword arguments.

davidhopkinson26: fixed in [10b7c5f](https://github.com/lvgig/tubular/pull/56/commits/10b7c5fae0743ac2c042520a456d3a7aec241207)

Ritunjai-Sharma (approved):

PR #63, opened 2023-01-19
V0.3.3 tidy up

davidhopkinson26:
Tidy up of changelog and docs ahead of release

### DIFF
--- a/docs/source/quick-start.rst
+++ b/docs/source/quick-start.rst
@@ -54,7 +54,7 @@ The standard `OutOfRangeNullTransformer <https://tubular.readthedocs.io/en/lates
 Dates
 ^^^^^
 
-This module contains transformers to deal with date columns.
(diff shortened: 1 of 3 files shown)


docs/source/quick-start.rst
```diff
@@ -54,20 +54,24 @@ The standard `OutOfRangeNullTransformer <https://tubular.readthedocs.io/en/lates
 Dates
 ^^^^^
```
TommyMatthews, line 55: All these additions are fine but there are a few other errors/improvments in the quickstart guide that it might be worth fixing in this tidy up:

- In Numeric and Nominal transformers inconsistently have brackets around the links
- in "Installation" I think "important" is meant to format the text but is appearing as text
- I think it would look a lot nicer with tubular in the first line as text with the picture underneath, rather than trying to use the logo as text

TommyMatthews (approved):

PR #66, opened 2023-02-01
Adds workaround for sklearn ohe get_feature_names deprecation

### LINKED ISSUE #65: [Bug]: get_feature_names method referenced but doesn't exist
### What happened?

When using the OneHotEncodingTransformer transform method the get_feature_names is called but is not defined. This seems to be the only reference to this method in the code, so seems to have been left over in an update.

### Environment

n/a

### Minimum reproducible code

```shell
from tubular.nominal import OneHotEncodingTransformer

ohe = OneHotEncodingTransformer(
    columns="feature", copy=True, verbose=True
)

ohe.fit(pd.DataFrame(data["feature"]))

data_ohe = ohe.transform(pd.DataFrame(data["feature"]))
```

### Relevant error output

```shell
AttributeError: 'OneHotEncodingTransformer' object has no attribute 'get_feature_names'
```

### Code of Conduct

TommyMatthews:
suggested solution for the get_feature_names deprecation by sklearn

Fixes #65

### DIFF
--- a/tubular/nominal.py
+++ b/tubular/nominal.py
@@ -947,7 +947,10 @@ class OneHotEncodingTransformer(BaseNominalTransformer, OneHotEncoder):
         # Apply OHE transform
         X_transformed = OneHotEncoder.transform(self, X[self.columns])
 
-        input_columns = self.get_feature_names(input_features=self.columns)
+        if hasattr(self, "get_feature_names"):
+            input_columns = self.get_feature_names(input_features=self.columns)
+        else:
+            input_columns = self.get_feature_names_out(input_features=self.columns)
 
         X_transformed = pd.DataFrame(
             X_transformed, columns=input_columns, index=X.index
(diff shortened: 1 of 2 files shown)


tubular/nominal.py
```diff
@@ -947,7 +947,10 @@ def transform(self, X):
         # Apply OHE transform
         X_transformed = OneHotEncoder.transform(self, X[self.columns])
 
-        input_columns = self.get_feature_names(input_features=self.columns)
+        if hasattr(self, "get_feature_names"):
```
davidhopkinson26, line 950: Can we add an explanatory comment here? could be confusing otherwise.

something like 'sklearn.OneHotEncoder.get_feature_names changed to sklearn.OneHotEncoder.get_feature_names_out in v1.2.  Logic to accommodate old and newer versions'

davidhopkinson26 (approved):

PR #67, opened 2023-02-07
Feature/multi class mre

TommyMatthews:


tubular/nominal.py
```diff
@@ -994,3 +1028,177 @@ def transform(self, X):
...
+    def __init__(self, columns: Union[str, list[str]], response_column: str, ohe_kwargs: dict = {}, mre_kwargs: dict = {}, process_all_levels_the_same: bool = True):                
+
+        super().__init__(columns = columns)
+
+        # dict checks
+        
+        ohe_kwargs = self._process_ohe_kwargs(ohe_kwargs)
+
+        if not isinstance(mre_kwargs, dict):
+            raise TypeError(f"mre_kargs should be a dict but got type {type(mre_kwargs)}")
```
davidhopkinson26, line 1080: *mre_kwargs

tubular/nominal.py
```diff
@@ -994,3 +1028,177 @@ def transform(self, X):
         X_transformed = pd.concat((X, X_transformed), axis=1)
 
         return X_transformed
+
+
+class MultiLevelMeanResponseTransformer(BaseNominalTransformer):
+    """
+    Create mean response encoded columns for a multi-level response. Consider using this transformer for non-ordinal multi-level responses. 
+    
+    An n-level response will be one hot encoded into a n binary responses. For each of the m features to be transformed, an encoded column will be
```
davidhopkinson26, line 1037: 'each of the m feature **columns**' ?

tubular/nominal.py
```diff
@@ -994,3 +1028,177 @@ def transform(self, X):
...
+
+class MultiLevelMeanResponseTransformer(BaseNominalTransformer):
+    """
+    Create mean response encoded columns for a multi-level response. Consider using this transformer for non-ordinal multi-level responses. 
+    
+    An n-level response will be one hot encoded into a n binary responses. For each of the m features to be transformed, an encoded column will be 
+    generated for each level, adding n x m new columns to the dataframe.
+
+    It is possible to customise the kwargs passed to each intermediate transformer, for both the one hot encoding step and the n mean response encoders. 
+    In each case it is possible to provide arguments for the Tubular transformmer and the underlying SciKit Learn transformer. Kwargs for either should
```
davidhopkinson26, line 1041: be specific about which tubular and scikit learn transformers

tubular/nominal.py
```diff
@@ -627,6 +618,62 @@ def fit(self, X, y):
...
+                    raise ValueError("Levels contains a level to encode against that is not present in the response.")
+
+                self.response_levels = self.level
+                
+            self.transformer_dict = {}
+            mapped_columns = []
+
+            for level in self.response_levels:
+                columns = [column + '_' + level for column in self.columns]
+                y_temp = y.apply(lambda x: x == level)
```
davidhopkinson26, line 664: if we create y_temp for each level do we need to pass a level to fit_binary_response?

tubular/nominal.py
```diff
@@ -647,15 +694,23 @@ def transform(self, X):
...
-        """
-
+        """  
+        
+        if self.level:
+            for response_level in self.response_levels:
+                for column in self.columns:
+                    X[column + '_' + response_level] = X[column]
+            drop_columns = self.columns
+            self.columns = self.mapped_columns
```
davidhopkinson26, line 704: Changing self.columns here would mean that following calls to transform are broken.  We need to at least set it back to original values after transforming data.  Preferable would be not changing transformer attributes during transform call.

TommyMatthews: I agree with this, given that we have;
- self.columns is to be set to columns to encode in init
- mapping generated for column copies that don't exist in the original dataframe
- mapping accessed by iterating over self.columns in `BaseMappingTransform Mixin.transform()`

We can't set self.columns differently in init because we need to learn the response levels from the data if `level = 'all' has been passed`

A far as I can see that leaves us with either _temporarily_ overwriting in transform then setting back, or writing a new method on `BaseMappingTransfomMixin`

tubular/nominal.py
TommyMatthews: Passing level is to control the generation of new columns in (X) temp in _fit_binary_response - I could probably change how this works, although I can't see immediately how without creating an X_temp before the method call

TommyMatthews: Done in this commit, unfortunately I can't see a way of removing the columns argument to make it just of the form fit(X,y). There's maybe a way involving an e.g. dictionary attribute that stores the new columns for each level, but I think that telling fit which element to access would be pretty convoluted.

https://github.com/lvgig/tubular/pull/67/commits/bf53838f22cb21e17ba1514d4a95c7eadda68e9a

tubular/nominal.py
TommyMatthews: Done here: https://github.com/lvgig/tubular/pull/67/commits/4b7baab553f7e7489b1d7d7e9e7c6844865ad72c

tubular/nominal.py
davidhopkinson26: I think it would also work to take a single column at a time and iterate over columns for line 663, but I'm happy enough with the approach taken here.

tubular/nominal.py
```diff
@@ -471,8 +472,20 @@ class MeanResponseTransformer(BaseNominalTransformer, BaseMappingTransformMixin)
     """Transformer to apply mean response encoding. This converts categorical variables to
     numeric by mapping levels to the mean response for that level.
 
+    For a continuous or binary response there is an natural mean to calculate and the columns
```
davidhopkinson26, line 475: I think this could be explained in simpler terms given it is the default usage.

'For a continuous or binary response the categorical columns specified will have values replaced with the mean response for each category.'

tubular/nominal.py
```diff
@@ -627,6 +675,73 @@ def fit(self, X, y):
...
+                        "Levels contains a level to encode against that is not present in the response."
+                    )
+
+                self.response_levels = self.level
+
+            self.transformer_dict = {}
+            mapped_columns = []
+
+            for level in self.response_levels:
+                columns = [column + "_" + level for column in self.columns]
```
davidhopkinson26, line 724: small thing - can we change this variable name? - too similar to self.columns

davidhopkinson26 (approved):

PR #72, opened 2023-02-15
value and type checks

alphanumericmale:
value and type checks for new dictionary functionality as well as unit tests

### DIFF
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -1027,8 +1027,8 @@ class DatetimeSinusoidCalculator(BaseTransformer):
         self,
         columns: Union[str, List[str]],
         method: Union[str, List[str]],
-        units: str,
-        period: Union[int, float] = 2 * np.pi,
+        units: Union[str, dict[str, str]],
+        period: Union[int, float, dict[str, int], dict[str, float]] = 2 * np.pi,
     ):
 
         super().__init__(columns, copy=True)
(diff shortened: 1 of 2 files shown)


tubular/dates.py
```diff
@@ -1040,20 +1040,56 @@ def __init__(
...
                     self.classname(), type(units)
                 )
             )
 
-        if (not isinstance(period, int)) and (not isinstance(period, float)):
+        if (
+            (not isinstance(period, int))
+            and (not isinstance(period, float))
+            and (not isinstance(period, dict))
+            or (isinstance(period, bool))
```
davidhopkinson26, line 1054: explicit check for bool dtype unnescessary?

alphanumericmale: apparently bool is a subclass of int so if a bool entered it sneaks through - did in the tests i made anyway

tubular/dates.py
```diff
@@ -1040,20 +1040,56 @@ def __init__(
...
+        if isinstance(period, dict):
+            if (
+                not all(isinstance(item, str) for item in list(period.keys()))
+                or (
+                    not all(isinstance(item, int) for item in list(period.values()))
+                    and not all(
+                        isinstance(item, float) for item in list(period.values())
+                    )
+                )
+                or any(isinstance(item, bool) for item in list(period.values()))
```
davidhopkinson26, line 1083: why the explicit check for bools?

alphanumericmale: see above

tubular/dates.py
```diff
@@ -1040,20 +1040,56 @@ def __init__(
...
+            ):
+                raise TypeError(
+                    "{}: period dictionary key value pair must be str:int or str:float but got {} {}".format(
+                        self.classname(),
+                        set(type(k) for k in period.keys()),
```
davidhopkinson26, line 1088: error message might be hard to read as dtype for all keys presented then dtype of all values rather than in pairs as they will ahve been input.  Either specify keys: {} values: {} in the error message or find a way to present the dtypes in a format more similar to input.

davidhopkinson26:
moved to #74

PR #75, opened 2023-03-03
Preserving dtypes - Arbitrary Imputer

JagadishSivakumar:
* Reading the dtype of the target column and casting the impute values to the same dtype before applying imputing.


davidhopkinson26:
Hi Jagadish, could you add some testing of the dtype preservation to the test pack please?

tubular/imputers.py
```diff
@@ -90,40 +85,41 @@ def __init__(self, impute_value, columns, **kwargs):
...
 
-            # for categorical dtypes have to set new category for the impute values first
-            if "category" in X[c].dtype.name:
+                    if self.impute_value not in X[c].cat.categories:
 
-                if self.impute_value not in X[c].cat.categories:
+                        X[c] = X[c].cat.add_categories(self.impute_value) # add new category
 
-                    X[c] = X[c].cat.add_categories(self.impute_value)
+                dtypes = X[c].dtype # preserve dtype
```
davidhopkinson26, line 115: `dtypes` variable only holds one dtype value at a time so the name is a little misleading.  Please change to something like `dtype` or `col_dtype`

JagadishSivakumar:
Hi, @davidhopkinson26 . Thanks for your comments. I have added a unit testing block for preserving the datatypes in arbitrary imputer. Additionally, I have also pushed code to downcast columns to the lowest possible dtypes, which can be found in the data_type_casting.py file.

This function is used to downcast dtypes in the testing and example notebooks. Also, I have created a PR for the develop branch. Kindly check and update reg.

tests/imputers/test_ArbitraryImputer.py
```diff
@@ -172,3 +175,20 @@ def test_super_columns_check_called(self, mocker):
         ):
 
             x.transform(df)
+
+    # Testing if the datatypes are preserved after imputation
+    def test_data_types_preserved(self):
```
davidhopkinson26, line 180: This test needs to to explicitly set the downcast dtypes.  Otherwise we need to fully test data_type_casting.py too. If that did not function as expected this test is compromised at present.

data_type_casting.py
```diff
@@ -0,0 +1,29 @@
+import pandas as pd
```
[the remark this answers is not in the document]
JagadishSivakumar: Thanks for your comments. I have added the downcast function in the example file. And have explicitly set the dtype of columns in unit_testing of Arbitrary Imputer. Hope this helps :)

PR #76, opened 2023-03-08
Preserve dtypes in the Arbitrary Imputer

JagadishSivakumar:
Reading the dtype of the target column and casting the impute values to the same dtype before applying imputing.

The unit testing and example for ArbitraryImputer are updated to show that the data types are preserved on imputing.


davidhopkinson26 (changes requested):
Please remove the .DS_store file

davidhopkinson26 (approved):

PR #78, opened 2023-03-14
0.3.4 changes into main

davidhopkinson26:
- added support for passing multiple columns and periods/units parameters to DatetimeSinusoidCalculator #74 

- added support for handling a multi level response to MeanResponseTransformer #67 

- changed ArbitraryImputer to preserve the dtype of columns (previously would upcast dtypes like int8 or float32) #76 

- fixed issue with OneHotencodingTransformer use of deprecated sklearn.OneHotEencoder.get_feature_names method #66

PR #79, opened 2023-03-15
Revert "v0.3.3 changes into main"

davidhopkinson26:
Reverts lvgig/tubular#64

correcting for a squashed commit from develop


davidhopkinson26:
failing linting checks due to reverting bugfix from #45.  #78 should fix this and it is only linting so we should be safe to override the checks in this case

adamsardar:
I agree that, whilst it isn't ideal to override the linter, commit history train wrecks can make for an unfortunate tangle. Agree that we can await #78 for restoration of lints  (currently passing).

adamsardar:
Practically impossible to review the squashed commits here as the individual commits have been lost. We should disable the ability to 'squash and merge' is possible; I don't see why you would do this at the stage of merge.

See: https://dev.to/wesen/squash-commits-considered-harmful-ob1

PR #81, opened 2023-04-18
Feature/pandas2 update

davidhopkinson26:
updating dependencies to require pandas 2.0.0.  Changes in datetime handling are not backwards compatible between versions.

e.g. automatic down casting to int32


davidhopkinson26:
since updating to pandas 2.0.0 values produced by DateDifferenceTransformer have changed.  So far unsure of reasons for this.

davidhopkinson26:
Problematic area is DateDifferenceTransformer which uses np.timedelta to calculate differences between datetimes in specified units.  According to https://stackoverflow.com/questions/54794431/adding-one-month-to-datetime64-with-timedelta this has not been supported in pandas for some time for Y and M.

We get different values for Y and M difference calculations using pandas 2.0.0.  It would appear that for Months this is due to pandas 2.0.0 treating all months as 31 days long, so the difference= diff in days/ 31 (which is an underestimate in calendar months)

Given that fractions of months are a bit arbitrary the current solution just calculates integer month and year differences.   Tempted to just drop year form this transformer since we have DateDiffLeapYearTransformer anyway for years.

davidhopkinson26:
Another option here would be to simply add a warning to DateDifferenceTransformer (maybe just if using pandas 2.0.0) that Y/M are no longer supported.  This would hopefully avoid the need to update dependencies.

davidhopkinson26:
I have added a Pandas <2.0.0 dependency and a warning that M/Y units for DateDifference transformer may be removed/changed in the near future.

Sarah-TaylorKnight (approved):
Looks good, definitely simpler than trying to support compatibility for both major versions of pandas

PR #83, opened 2023-05-04
Feature/dev container

davidhopkinson26:
Adds devcontainer files to project so developers can automatically have an appropriate environment set up in codespaces.

Currently using a python 3.8 container image and then installing requirements-dev.txt. Also installs some default vscode extensions (pylance and gitlens).

Welcome suggestions for other extensions to add by default.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -12,7 +12,7 @@
 	// "forwardPorts": [],
 
 	// Use 'postCreateCommand' to run commands after the container is created.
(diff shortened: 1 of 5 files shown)


davidhopkinson26:
for some reason black/flake8 are passing on checks but still making changes when I run them locally despite versions being identical.  Unsure why and leaving without changes for now.

Sarah-TaylorKnight (changes requested):
All looks good and helps out with codespaces! There's a conflict left on `CHANGELOG.rst` but once that's resolved I'll approve

PR #84, opened 2023-05-10
59 unexpected dtype change warning moved to basemappingtransformer

alphanumericmale:
Changes made to mapping transformers - 

todo: update the tests for these


tubular/mapping.py
```diff
@@ -35,40 +34,58 @@ class BaseMappingTransformer(BaseTransformer):
...
 
         super().__init__(columns=columns, **kwargs)
 
-    def transform(self, X):
-        """Base mapping transformer transform method.  Checks that the mappings
+    def check_dtype_changes(self, X, original_dtypes, suppress_dtype_warning=False):
+        mapped_columns = self.mappings.keys()
+        mapped_dtypes = X[mapped_columns].dtypes
+
+        if not suppress_dtype_warning:
```
davidhopkinson26, line 60: It might be better to bring `if not suppress_dtype_warning` out of this function.  Since the check might slow the transformer down it would be good to both skip the logging of the mappings and the check itself if suppress_dtype_warning is True.  Also cleaner code to have less args passing.

tubular/mapping.py
```diff
@@ -150,97 +173,69 @@ class MappingTransformer(BaseMappingTransformer, BaseMappingTransformMixin):
...
-                        is_categorical_dtype(original_dtypes[col])
-                        and is_categorical_dtype(mapped_dtypes[col])
-                    ):
+    X = BaseMappingTransformMixin.transform(self, X)
 
-                        warnings.warn(
-                            f"{self.classname()}: This mapping changes {col} dtype from {original_dtypes[col]} to {mapped_dtypes[col]}. This is often caused by having multiple dtypes in one column, or by not mapping all values."
-                        )
+    # Call the check_dtype_changes method
+    self.check_dtype_changes(X, original_dtypes, suppress_dtype_warning)
```
davidhopkinson26, line 236: why are we doing the dtype changes check in two places? both in `BaseMappingTransformMixin.transform` and after it?

tubular/mapping.py
```diff
@@ -78,8 +95,15 @@ def transform(self, X):
...
         self.check_is_fitted(["mappings"])
 
+        # Store the original dtypes of the columns
+        mapped_columns = self.mappings.keys()
+        original_dtypes = X[mapped_columns].dtypes
+
         X = super().transform(X)
 
+        # Call the check_dtype_changes method
+        self.check_dtype_changes(X, original_dtypes, suppress_dtype_warning)
```
davidhopkinson26, line 105: I'm not sure we need the check here as super().transform() doesn't do any mappings?

adamsardar:
Is there a ticket to place this PR in context a bit? I looked, but couldn't see anything.

It might be worth exploring third-party functionality on this one. [pandera](https://pandera.readthedocs.io/en/stable/schema_inference.html) has a means of inferring a schema from existing dtypes and can perform a variety of smart checks on a new dataframe. But it's tricky to know if that's suitable without knowing what these changes exist to fix? Is there a fear of dtypes changing, or has some observed it with an example (which would create a good test to kick off with).

PR #92, opened 2023-06-29
Feature/tidy up

davidhopkinson26:
Tidy up of code quality mainly using automated tools (black and ruff)

### DIFF
--- a/examples/Data-Science-Festival-Workshop/Demo.ipynb
+++ b/examples/Data-Science-Festival-Workshop/Demo.ipynb
@@ -24,11 +24,9 @@
    "outputs": [],
    "source": [
     "try:\n",
-    "\n",
     "    import matplotlib\n",
-    "    \n",
-    "except ModuleNotFoundError: \n",
-    "    \n",
+    "\n",
+    "except ModuleNotFoundError:\n",
     "    %pip install matplotlib"
    ]
   },
@@ -59,6 +57,7 @@
    ],
    "source": [
     "import matplotlib\n",
+    "\n",
     "matplotlib.__version__"
    ]
   },
@@ -89,6 +88,7 @@
    ],
    "source": [
     "import tubular\n",
+    "\n",
     "tubular.__version__"
    ]
   },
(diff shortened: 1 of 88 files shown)


davidhopkinson26:
Need to work out how to specify rules list in github action to exclude E501 (88 char limit) or find a way to fix this en masse!

This rule is stricter than black as it applies to long strings which black ignores.

davidhopkinson26:
Adding quite a few rules in this PR.  Of the opinion that having a very opinionated linting ruleset is a good thing for code consistency when we have so many contributors but open to debate this.

Would aim to add D, AN and PT rules in future PR.  More work to manually fix current code for them though.

davidhopkinson26:
Audit of unused rules:

TODO:

- B - 21 errors
- DTZ - 146 DTZ001 errors 
- PT - 27 errors
- D - docstrings - 314 errors
- ANN - type annotations. 2547 errors. Big one
- BLE - 4 erros
- ICN - causes issue son a couple of tests which mock pandas
- T- 9 print statements
- SLF - private memeber access
- ARG - a few easy flags to fix
- PTH -  a few easy fixes
- PLR - lots of refactor suggestions
- TRY - few relativley simple fixes
- RUF - few flags to fix
- INP - wants __init__.py for test modules.

not sure:
- S - Bandit.  mainly flagging S101 - assert use due to potential to be ignored in optimised python run (-O).
- FBT - disallows boolean args having defaults.  breaks backwards compatibility.
- TD -  2 TDs flagging up
- FIX - as above
- ERA - deletes all comented out code, not sure we want to do this on commit but would be good for build pipleine.


not relevant:
- DJ - django
- EXE - executables
- G - logging
- AIR -airflow

Would aim to add D, AN, PT rules in future PR.  More work to manually fix current code for them though.

Also look to add C90 (code complexity) in future.  Only a few violations at the moment but quite complex to fix.  That is of course exactly why they are worth fixing though.

Sarah-TaylorKnight (approved):
All changes look good :)

I agree on the extra rules to add in once when we get the opportunity to - I'd also suggest we keep an eye on the functionality for notebooks as that would be also useful, though currently hit or miss as it's experimental.

PR #99, opened 2023-07-10
fixed bug in setup.py

davidhopkinson26:
bug introduced into setup.py when applying ruff rules.  tryign to open context manager with  

`while open(VERSION_FILE).read()`

led to an error because read()  does not have an _ enter _ method.


Sarah-TaylorKnight (approved):
Looks good

PR #123, opened 2023-07-17
122 datedifferencetransformer pandas 20 compatibility issues

davidhopkinson26:


tests/dates/test_DateDifferenceTransformer.py
```diff
@@ -154,7 +160,7 @@ def test_units_warning(self, unit):
         """Test that a warning is raised if M units passed to init."""
         with pytest.warns(
             Warning,
-            match="DateDifferenceTransformer: Y/M units will be changed or deprecated in a future version, consider using DateDiffLeapYearTransformer or D units instead",
+            match="DateDifferenceTransformer: Y/M will be deprecated ina future version, use D instead. Y/M units are not supported when using pandas>=2.0.0 and will produce inaccurate results. For an accurate age in years use DateDiffLeapYearTransformer.  M units will produce number of days/31.",
```
Sarah-TaylorKnight, line 163: I think this comment leaves some ambiguity - saying `Y/M units are not supported` implies they would cause an error, but they will run except with a different result with Pandas>2.0.0

davidhopkinson26: Yes it is ambiguous.  The Y/M units are not supported in pandas timedeltas anymore, even if they do still produce results it won't be reliable going forward (in fact it hasnt been for some time https://stackoverflow.com/questions/54794431/adding-one-month-to-datetime64-with-timedelta). 

On balance I think we are also taking on quite a bit of technical debt to try and support this piece of functionality when pre using pandas<=2.0.0.

Given that Y and M are not that useful anyway as units due to ambiguity I'm now of the opinion that we should just drop these units from the transformer to avoid ambiguity.

davidhopkinson26:
Warning too ambiguous. The Y/M units are not supported in pandas timedeltas anymore, even if they do still produce results it won't be reliable going forward (in fact it hasnt been for some time https://stackoverflow.com/questions/54794431/adding-one-month-to-datetime64-with-timedelta).

On balance I think we are also taking on quite a bit of technical debt to try and support this piece of functionality when pre using pandas<=2.0.0.

Given that Y and M are not that useful anyway as units due to ambiguity now dropping these units from the transformer to avoid ambiguity.

Sarah-TaylorKnight (approved):
I agree this is the easiest approach to get the pandas 2.0 support 👍

Issue #116, opened 2023-07-11
Apply ruff rulse ANN flake8-annotations


davidhopkinson26:
https://beta.ruff.rs/docs/rules/#flake8-annotations-ann

make code compliant with rule set and then add to ruleset in .ruff.toml. If appropriate identify specific rules or parts of code to exclude.

2547 errors found for missing annotations.

This is a big one, happy to take contributions by module or individual script rather than addressing entire package in one go.

adamsardar:
Closed by #128

PR #128, opened 2023-07-21
Adding type hints #116

Sarah-TaylorKnight:
Working through the repo to add in type hints across the code base in order to address ANN ruff rules:

ANN001: missing-type-function-argument - Missing type annotation for function argument
ANN002: missing-type-args - Missing type annotation for *{name}
ANN003: missing-type-kwargs - Missing type annotation for **{name}
ANN101: missing-type-self - Missing type annotation for {name} in method
ANN102: missing-type-cls - Missing type annotation for {name} in classmethod
ANN201: missing-return-type-undocumented-public-function - Missing return type annotation for public function {name}
ANN202: missing-return-type-private-function - Missing return type annotation for private function {name}
ANN204: missing-return-type-special-method - Missing return type annotation for special method {name}
ANN205: missing-return-type-static-method - Missing return type annotation for staticmethod {name}
ANN206: missing-return-type-class-method - Missing return type annotation for classmethod {name}
ANN401: any-type - Dynamically typed expressions (typing.Any) are disallowed in {name}

**Notes:**
I'd suggest we exclude ANN101 as it adds clutter without much information, and as ruff put it ["Note that many type checkers will infer the type of self automatically, so this annotation is not strictly necessary."](https://beta.ruff.rs/docs/rules/missing-type-self/)


Sarah-TaylorKnight:
Looking at ANN03 - type hinting kwargs - this is quite difficult to implement in general, and tubular in general does not have many params passed through as kwargs - unsure how to proceed with these

davidhopkinson26:
Have now made all the main tubular modules compliant with ANN rules except ANN101 and ANN003.  Still a mountain of type hints to add in test modules though.  Might need to merge what we have so far and return to adding type hints to tests later.

davidhopkinson26:
looks like we may be able to switch off specific ruff rules on specific files https://github.com/astral-sh/ruff/pull/2978, which would enable us to add the rules to CICD without making test modules compliant

davidhopkinson26:
> Looking at ANN03 - type hinting kwargs - this is quite difficult to implement in general, and tubular in general does not have many params passed through as kwargs - unsure how to proceed with these

I think this rule is best left ignored.  many **kwargs are present in tubular for using arbitrary pandas functions so do not have types we could easily specify.

tubular/nominal.py
```diff
@@ -978,13 +998,13 @@ class OneHotEncodingTransformer(BaseNominalTransformer, OneHotEncoder):
...
     def __init__(
         self,
-        columns=None,
-        separator="_",
-        drop_original=False,
-        copy=True,
-        verbose=False,
-        **kwargs,
-    ):
+        columns: str | list[str] | None = None,
```
TommyMatthews, line 1001: The functionality of the None option here isn't in the docstring - please can you quickly add something to say that in this case it will transform all of the columns.

tubular/nominal.py
```diff
@@ -554,7 +560,11 @@ def __init__(
 
         BaseNominalTransformer.__init__(self, columns=columns, **kwargs)
 
-    def _prior_regularisation(self, target_means, cat_freq):
+    def _prior_regularisation(
+        self,
+        target_means: pd.Series,
+        cat_freq: str,
```
TommyMatthews, line 566: I think this should be a pd.series, needs changing in the docstring as well.

tubular/nominal.py
TommyMatthews: This actually needs doing in a few places so happy to leave if you don't want to address this in this PR

tubular/numeric.py
```diff
@@ -164,7 +166,13 @@ class CutTransformer(BaseTransformer):
 
     """
 
-    def __init__(self, column, new_column_name, cut_kwargs=None, **kwargs) -> None:
+    def __init__(
+        self,
+        column: str,
+        new_column_name: str,
+        cut_kwargs: dict | None = None,
```
TommyMatthews, line 173: Think this can be dict[str,object]

tubular/numeric.py
```diff
@@ -263,11 +271,11 @@ class TwoColumnOperatorTransformer(DataFrameMethodTransformer):
...
         self,
-        pd_method_name,
-        columns,
-        new_column_name,
-        pd_method_kwargs=None,
-        **kwargs,
+        pd_method_name: str,
+        columns: list[str],
+        new_column_name: str,
+        pd_method_kwargs: dict | None = None,
```
TommyMatthews, line 277: Think this can be dict[str, object]

tubular/numeric.py
```diff
@@ -348,7 +356,13 @@ class ScalingTransformer(BaseTransformer):
 
     """
 
-    def __init__(self, columns, scaler_type, scaler_kwargs=None, **kwargs) -> None:
+    def __init__(
+        self,
+        columns: str | list[str] | None,
+        scaler_type: str,
+        scaler_kwargs: dict | None = None,
```
TommyMatthews, line 363: Think this can be dict[str, object]

tubular/strings.py
```diff
@@ -51,11 +53,11 @@ class SeriesStrMethodTransformer(BaseTransformer):
...
         self,
-        new_column_name,
-        pd_method_name,
-        columns,
-        pd_method_kwargs=None,
-        **kwargs,
+        new_column_name: str,
+        pd_method_name: str,
+        columns: str,
+        pd_method_kwargs: dict | None = None,
```
TommyMatthews, line 59: I think this can be dict[str, object]

tubular/comparison.py
```diff
@@ -1,7 +1,12 @@
-import pandas as pd
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
 
 from tubular.base import BaseTransformer
 
+if TYPE_CHECKING:
```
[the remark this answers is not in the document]
adamsardar: This will have been to appease the `TCH001` [typing-only-first-party-import lint](https://docs.astral.sh/ruff/rules/typing-only-first-party-import/) as part of the TCH set.

The pandas API isn't directly used at runtime so, for the code in just this sub-class, pandas is not needed. Hence the linter is asking for it to be placed in a special block that only the type hinter sees. But its parent need pandas, so it's a bit of a moot point.

We should probably:
i.) Use the pandas API directly (perhaps [pandas.Series.equals](https://pandas.pydata.org/docs/reference/api/pandas.Series.equals.html#pandas-series-equals)?) rather than `==`. Out of scope and a ticket in its own right, alongside a test showing where `==` is problematic
ii.) Keep pandas as an import (it's core to the package) and just enable `# noqa: TCH001` next to it.

tubular/dates.py
```diff
@@ -57,12 +58,12 @@ class DateDiffLeapYearTransformer(BaseTransformer):
...
         self,
-        column_lower,
-        column_upper,
-        new_column_name,
-        drop_cols,
-        missing_replacement=None,
-        **kwargs,
+        column_lower: str,
+        column_upper: str,
+        new_column_name: str,
```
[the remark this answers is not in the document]
adamsardar: Relate to #142 and subsequent PR activity at all?

tubular/capping.py
```diff
@@ -61,10 +63,10 @@ class CappingTransformer(BaseTransformer):
 
     def __init__(
         self,
-        capping_values=None,
-        quantiles=None,
-        weights_column=None,
-        **kwargs,
+        capping_values: dict[str, list[int | float | None]] | None = None,
+        quantiles: dict[str, list[int | float]] | None = None,
```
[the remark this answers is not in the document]
adamsardar: I think that I'd disagree here; `quantiles` can _only_ be a dict of key-value pairs between 0 and 1 (see error block on line 89). 

I'd query why `int` is permitted at all given that only 0L and 1L are permissible values. In fact, [this link](https://stackoverflow.com/a/50928627) insists that we should be using `float` *only*, as `int` is implied.

> Rather than requiring that users write import numbers and then use numbers.Float etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type float, an argument of type int is acceptable; similar, for an argument annotated as having type complex, arguments of type float or int are acceptable.

(from [PEP484 Type Hints](https://peps.python.org/pep-0484/#the-numeric-tower))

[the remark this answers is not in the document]
adamsardar: Reading through the conversation there, it seems as though people still seem to write `float | int` in the wild. I'd rather we left this as it is.

tubular/comparison.py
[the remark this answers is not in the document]
adamsardar: Elected to use option ii. until such time that we enable option i.

An arguably superior way to exempt classes from being moved into a `TYPE_CHECKING` block would be to use the [runtime-evaluated-base-classes](https://docs.astral.sh/ruff/settings/#flake8-type-checking-runtime-evaluated-base-classes) option in `ruff.toml`, but that would require a higher version of `ruff` than the project currently uses. Not for this PR.

tubular/capping.py
```diff
@@ -100,7 +102,11 @@ def __init__(
         self.weights_column = weights_column
         self._replacement_values = copy.deepcopy(self.capping_values)
 
-    def check_capping_values_dict(self, capping_values_dict, dict_name):
+    def check_capping_values_dict(
+        self,
+        capping_values_dict: dict[str, list[int | float]],
```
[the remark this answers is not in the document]
adamsardar: Agreed and added.

tubular/dates.py
```diff
@@ -285,10 +286,10 @@ class ToDatetimeTransformer(BaseTransformer):
 
     def __init__(
         self,
-        column,
-        new_column_name,
-        to_datetime_kwargs=None,
-        **kwargs,
+        column: str,
+        new_column_name: str,
+        to_datetime_kwargs: dict | None = None,
```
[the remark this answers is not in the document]
adamsardar: Changed

tubular/nominal.py
adamsardar: The object class name is capitalised - see [the docs](https://pandas.pydata.org/docs/reference/api/pandas.Series.html) - in keeping with [PEP8](https://peps.python.org/pep-0008/#class-names).

> Class names should normally use the CapWords convention.
> 
> The naming convention for functions may be used instead in cases where the interface is documented and used primarily as a callable.

No change needed?

tubular/dates.py
```diff
@@ -394,11 +395,11 @@ class SeriesDtMethodTransformer(BaseTransformer):
...
         self,
-        new_column_name,
-        pd_method_name,
-        column,
-        pd_method_kwargs=None,
-        **kwargs,
+        new_column_name: str,
+        pd_method_name: str,
+        column: str,
+        pd_method_kwargs: dict | None = None,
```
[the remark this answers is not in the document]
adamsardar: Changed to `dict[str, Any]`.

But riffing on your comment below, why not change this (and above datetime kwargs) to `dict[str, object]`.

From the [MyPy docs](https://mypy.readthedocs.io/en/latest/dynamic_typing.html#any-vs-object):

> If you’re not sure whether you need to use [object](https://docs.python.org/3/library/functions.html#object) or Any, use [object](https://docs.python.org/3/library/functions.html#object) – only switch to using Any if you get a type checker complaint.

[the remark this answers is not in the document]
adamsardar: Using `dict[str, object]` on the above advice. If it breaks then we can go back to using `Any`.

tubular/dates.py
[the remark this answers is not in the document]
adamsardar: Using `dict[str, object]` on the advice discussed below. If it breaks then we can go back to using `Any` - see below.

tubular/numeric.py
adamsardar: Changed.

tubular/numeric.py
adamsardar: Changed

tubular/numeric.py
adamsardar: Changed

tubular/strings.py
adamsardar: Changed

adamsardar:
- Comments above have been incorporated and resolved
- `main` has been merged and build restored to passing
- PR ready to merge now
- three responses to comments above outstanding

A really solid bit of work here - thanks for the hard work all!

tubular/capping.py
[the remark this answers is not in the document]
adamsardar: Looking at [OutOfRangeNullTransformer](https://github.com/lvgig/tubular/blob/fab26402081627f6f5f71bf26f3b84b4f1a2fc18/tubular/capping.py#L406):

```python
class OutOfRangeNullTransformer(CappingTransformer):
    ...
    def __init__(
        self,
        capping_values: dict[str, list[int | float | None]] | None = None,
        quantiles: dict[str, list[int | float | None]] | None = None,
        weights_column: str | None = None,
        **kwargs: dict[str, bool],
    ) -> None:
        super().__init__(
            capping_values=capping_values,
            quantiles=quantiles,
            weights_column=weights_column,
            **kwargs,
        )
```
You're right, there is a type hint for `None` to `quantiles` in there. But I think this a mistake as it gets passed to the parent class constructor, [CappingTransformer](https://github.com/lvgig/tubular/blob/fab26402081627f6f5f71bf26f3b84b4f1a2fc18/tubular/capping.py#L14):

```python
class CappingTransformer(BaseTransformer):
   ...
    def __init__(
        self,
        capping_values: dict[str, list[int | float | None]] | None = None,
        quantiles: dict[str, list[int | float]] | None = None,
        weights_column: str | None = None,
        **kwargs: dict[str, bool],
    ) -> None:

```

Which does not permit `None`. This is an error - will change.

tubular/dates.py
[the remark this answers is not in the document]
adamsardar: A good spot! I've changed the typehint and behaviour for `new_column_name` in `DateDiffLeapYearTransformer` to match and added a test.

As for the broader topic of combining `DateDiffLeapYearTransformer` and `DateDifferenceTransformer` classes; the above did leave me feeling as though I was running foul of Don't Repeat Yourself principles. It's not in scope for this PR, but it is a good point - what degree of API stability do we promise users? I'm not against combining objects which are very similar to decrease the size of the codebase (that is the point after all). The implementations checks make this tricky - see #155.

Let's capture this thought on #156 as it is only one of a variety of similar thoughts around general class architecture. As we move to arrow-backed types (and types more generally), it is timely to think about architecture.

Closing this thread.

tubular/dates.py
```diff
@@ -1089,10 +1104,10 @@ class DatetimeSinusoidCalculator(BaseDateTransformer):
...
     def __init__(
         self,
-        columns: Union[str, List[str]],
-        method: Union[str, List[str]],
-        units: Union[str, dict],
-        period: Union[int, float, dict, dict] = 2 * np.pi,
+        columns: str | list[str],
+        method: str | list[str],
+        units: str | dict,
+        period: int | float | dict | dict = 2 * np.pi,
```
adamsardar, line 1110: Typo - `dict | dict`. This is a union, so repeated elements are moot.

tubular/nominal.py
adamsardar: I've added the type to the docstring. BUT long term I really don't see much use in keeping type annotation in the code AND in the docs - let the Sphix documentation build extract the typehints and add them to the docs. See [this plugin](https://github.com/agronholm/sphinx-autodoc-typehints).

adamsardar (approved):
Looking good!

tubular/dates.py
adamsardar: Removed.

PR #132, opened 2023-07-26
Comply to ruff DTZ rules

ClaireF57:
Correct 114 errors linked to DTZ001  rule :The use of datetime.datetime() without tzinfo argument is not allowed


davidhopkinson26 (approved):
Changes look good.

Have updated CHANGELOG and fixed a small merge conflict before approving.

PR #135, opened 2023-09-11
134 remove returnkeydict class

davidhopkinson26:
ReturnKeyDict was used in the past to facilitate leaving unmapped values in the target dataframe unchanged.  Pandas also has a Series.replace() method which can achieve this functionality without the need for the ReturnKeyDict class.

This also means we can stop looping over the columns to map in BaseMappingTransformerMixin.transform() as the replace method natively supports reading a dictionary with the first level defining columns to map.

The behaviour of mapping required for MeanResponseTransformer was not preserved with this change.  However I have decoupled it from BaseMappingTransformerMixin and implemented MeanResponseTransformer.map_imputer_values() instead since the feature of leaving unmapped values alone is not required here and thus the coupling was unhelpful.

In going through this process I have come to the conclusion that MeanResponseTransformer could do with a refactor!


tubular/nominal.py
```diff
@@ -747,6 +747,12 @@ def fit(self, X, y):
 
         return self
 
+    def map_imputation_values(self, X):
```
TommyMatthews, line 750: Please can this have a docstring

tubular/nominal.py
```diff
@@ -747,6 +747,12 @@ def fit(self, X, y):
...
+        for c in self.columns:
+            X[c] = X[c].map(self.mappings[c])
```
TommyMatthews, line 752: I know it's a bit tangential but please could you rewrite the docstring for the self.mappings attribute, it's quite difficult to understand.

tubular/nominal.py
```diff
@@ -429,7 +429,7 @@ def transform(self, X):
         return X
 
 
-class MeanResponseTransformer(BaseNominalTransformer, BaseMappingTransformMixin):
+class MeanResponseTransformer(BaseNominalTransformer):
```
TommyMatthews, line 432: I know it's off topic but there's an important typo in line 444 - n>1 should read n =1

tubular/nominal.py
```diff
@@ -783,12 +789,14 @@ def transform(self, X):
             for c in self.columns:
                 # finding rows with values not in the keys of mappings dictionary
                 unseen_indices[c] = X[~X[c].isin(self.mappings[c].keys())].index
-            X = BaseMappingTransformMixin.transform(self, X)
+            X = super().transform(X)
```
TommyMatthews, line 792: Not one for now, but maybe one for the backlog? this section feels a bit clunky and hard to follow. Would be nice not to have the super.transform and self.map_imputation_values twice, and feels confusing that in once case we call the self.check_mappable_rows class method, and in the other we ignore this and call a slightly different but similarly named method.

Feel free to completely ignore this but I have stared at it for too long not to write something

tests/mapping/test_BaseMappingTransformMixin.py
```diff
@@ -77,34 +77,44 @@ def test_pd_series_map_call(self, mocker):
...
-        assert spy.call_count == 2, "unexpected number of calls to pd.Series.map"
+        assert spy.call_count == 2, "unexpected number of calls to pd.Series.replace"
 
         call_args = spy.call_args_list[0]
 
         call_pos_arg = call_args[0]
         call_kwargs = call_args[1]
 
-        assert call_kwargs == {}
+        assert call_kwargs == {"regex": False}
```
TommyMatthews, line 87: I can't see where this is coming from, please could you add a comment? (or tell me if it's obvious)

tubular/nominal.py
TommyMatthews: I've just noticed your comment that mean response transformer could do with a refactor, I agree

tubular/nominal.py
davidhopkinson26: resolve din 48aa1075cd59809aa0b918791a82d78078626182

tubular/nominal.py
davidhopkinson26: de265f2a3c073d4d7e4ed5ae70e5237ee2205cbb

tubular/nominal.py
davidhopkinson26: I have added a ticket here #149

tubular/nominal.py
davidhopkinson26: https://github.com/lvgig/tubular/pull/135/commits/48aa1075cd59809aa0b918791a82d78078626182

tests/mapping/test_BaseMappingTransformMixin.py
davidhopkinson26: I don't totally understand this either.  False is the default value for the regex arg in Series.replace so this will be giving us the behaviour we want, juts have no idea where it gets inserted as a kwarg. Have added a comment to highlight this https://github.com/lvgig/tubular/pull/135/commits/01c9f65da452797503b586c357a055dfb27034f4

davidhopkinson26: Will be something to do with the way df. replace is implemented vs map

TommyMatthews: Ok, cheers :)

TommyMatthews (approved):

PR #141, opened 2023-11-07
Adds capability to avoid passing unseen levels to rare

TommyMatthews:
Parameter 'encode_unseen_levels' can be set to False. (default True). 

If so, the levels present in the training data will be stored in during the fit method call. In transform, any new levels not present in the training data will be added to the mappings_ dictionary and thus not mapped to rare.

### DIFF
--- a/tests/nominal/test_GroupRareLevelsTransformer.py
+++ b/tests/nominal/test_GroupRareLevelsTransformer.py
@@ -142,9 +142,9 @@ class TestFit:
         ta.classes.test_object_attributes(
             obj=x,
             expected_attributes={
-                "mapping_": {"b": ["a", np.NaN], "c": ["a", "c", "e"]},
+                "non_rare_levels": {"b": ["a", np.NaN], "c": ["a", "c", "e"]},
             },
-            msg="mapping_ attribute",
+            msg="non_rare_levels attribute",
         )
 
     def test_learnt_values_weight(self):
@@ -157,8 +157,8 @@ class TestFit:
 
         ta.classes.test_object_attributes(
             obj=x,
-            expected_attributes={"mapping_": {"b": ["a", np.NaN]}},
-            msg="mapping_ attribute",
+            expected_attributes={"non_rare_levels": {"b": ["a", np.NaN]}},
+            msg="non_rare_levels attribute",
         )
 
     def test_learnt_values_weight_2(self):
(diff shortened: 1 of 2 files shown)


davidhopkinson26:
thanks @TommyMatthews! Please run the pre-commit hook to hopefully autofix the linting failures for Black and Ruff.  If you use github codespaces this will all be set up automatically for you.

tubular/nominal.py
```diff
@@ -403,6 +425,12 @@ def transform(self, X):
 
         self.check_is_fitted(["mapping_"])
 
+        if not self.encode_unseen_levels:
+            for c in self.columns:
+                unseen_vals = set(X[c]) - set(self.training_data_levels[c])
+                for unseen_val in unseen_vals:
+                    self.mapping_[c].append(unseen_val)
```
davidhopkinson26, line 432: Predates this PR but on seeing it here and getting stuck for a moment in confusion I'm wondering if `mapping_` is a bit of a misleading variable name in this transformer?  Would `non_rare` be clearer and a quick fix?

tubular/nominal.py
```diff
@@ -266,6 +276,7 @@ def __init__(
         weight=None,
         rare_level_name="rare",
         record_rare_levels=True,
+        encode_unseen_levels: bool = True,
```
davidhopkinson26, line 279: Nit -  `unseen_levels_to_rare` might be more explicit than `encode_unseen_levels`

tests/nominal/test_GroupRareLevelsTransformer.py
```diff
@@ -57,21 +57,30 @@ def test_weight_not_str_error(self):
...
 
+    def test_encode_unseen_levels_not_bool_error(self):
+        """Test that an exception is raised if encode_unseen_levels is not a bool."""
+        with pytest.raises(
+            ValueError,
+            match="GroupRareLevelsTransformer: encode_unseen_levels must be a bool",
+        ):
+            GroupRareLevelsTransformer(encode_unseen_levels=2)
+
     def test_values_passed_in_init_set_to_attribute(self):
```
davidhopkinson26, line 76: Thought we deleted these tests but must have missed some.  Please delete this and `test_arguments`, `test_class_methods` throughout this test module. We are ditching implementation tests where possible as they create friction for no real value.

davidhopkinson26:
Looks good @TommyMatthews.  Please see above for a couple of nitpicking comments on var names and some redundant tests to remove but I think the solution implemented here is quite elegant and neat :)

tubular/nominal.py
TommyMatthews: Yeah I agree that reads a lot more clearly. Done here:

https://github.com/lvgig/tubular/pull/141/commits/d1dd04bdff87b19a48bd449aa04eba6ff44e0a23

tubular/nominal.py
TommyMatthews: I agree - I went for non_rare_levels as I felt that was even clearer, but happy to go for non rare if you disagree. Done here:

https://github.com/lvgig/tubular/pull/141/commits/016a45b58fbaba7ea7aaab7a3b93305e80afe4f2

davidhopkinson26 (approved):

PR #158, opened 2024-01-09
v1.1.1 changes into main

davidhopkinson26:


tests/dates/test_DateDiffLeapYearTransformer.py
```diff
@@ -167,6 +167,25 @@ def test_inputs_set_to_attribute(self):
             msg="Attributes for DateDiffLeapYearTransformer set in init",
         )
 
+    def test_inputs_set_to_attribute_name_not_set(self):
```
TommyMatthews, line 170: This looks like an expected attribute test getting added back in?

Maybe an accidental result of another PR? Happy to let it through but thought I'd flag it first

davidhopkinson26: thanks for flagging.  I'll create a PR into develop to remove this test.

davidhopkinson26: On second reading this is not an implementation test and is one we should keep in (which we have flagged for removal).  It tests the functionality of setting new column_name correctly when a new name is not provided.

TommyMatthews (approved):

Issue #136, opened 2023-09-12
remove develop branch


davidhopkinson26:
The main/ develop split adds a frustrating roadblock to development as it effectively requires doubling up of review on changes.  Would be much more straightforward to just merge feature branches straight into main and tag releases.  Possibly also creating release branches.

I think this may have implications for the documentation on readthedocs so need to investigate that before removing the develop branch.

adamsardar:
I've come to learn that this style of development ('main+friends') is called [trunk-based development](https://trunkbaseddevelopment.com/).

I'm a fan of 'tag to release' as a model - [microsoft really like release branches](https://learn.microsoft.com/en-us/devops/develop/how-microsoft-develops-devops#differences-from-github-flow), but I believe that they also commit to patching critical bugs previous releases (hence causing the history to diverge), which we would never do for tubular - people should just update to a version with the bug patched. The release branches would effectively be save-game points on the journey through the commit history, which is exactly the same as tags.

davidhopkinson26:
Thanks for the link.  I agree about tagging vs. release branches, no current plans to bugfix old releases.  Trunk based development with tagged releases seems like the way to go.  Will look to make this change after the next release.

davidhopkinson26:
Changed pypi release process to trigger upon creating a release on github (which also creates a release tag) in #161.  Have now removed develop branch so we will use  [trunk-based development](https://trunkbaseddevelopment.com/) from main from now on.

PR #161, opened 2024-01-09
Feature/release pipeline

davidhopkinson26:
Pypi now requires 2FA and as such publishing new releases needs either an [api token](uhttps://pypi.org/help/#apitoken) or setting a ['trusted publisher' ](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) and a release workflow.  Uploading dist files using twine and entering username/password is no longer supported.

I have added a release pipeline to the .github folder. which is set up to use Trusted publishers.  this could be changed to use an API token instead by adding:

`      with:
        user: __token__
        password: ${{ secrets.PYPI_API_TOKEN }}`
        
Note however that the pypi documentation strongly recommends using trusted publishers over [api token](uhttps://pypi.org/help/#apitoken)

The pipeline is triggered on publication and uses the gh-action-pypi-publish git hub action.  The script is an amalgam of the [example scripts there](https://github.com/pypa/gh-action-pypi-publish) and the default publish to pypi action from the actions pane.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,6 @@
-on: workflow_dispatch
+on:
+  release:
+    types: [published]
     
 jobs:
   pypi-publish:
@@ -22,7 +24,3 @@ jobs:
 
     - name: Publish package distributions to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
-    - name: Github Release  
-      uses: softprops/action-gh-release@v1  
-      with:  
-        files: dist/*
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,7 @@
-
+on:
+  release:
+    types: [published]
+    
 jobs:
   pypi-publish:
     name: Upload release to PyPI
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v4
       with:
-        python-version: "3.8"
+        python-version: "3.x"
     - name: "Build Package"
       run: |
         python setup.py sdist bdist_wheel
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,15 +12,15 @@ jobs:
     permissions:
       id-token: write  # IMPORTANT: this permission is mandatory for trusted publishing
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v4
-        with:
-          python-version: "3.8"
-      - name: "Build Package"
-        run: |
-          python setup.py sdist bdist_wheel
-          pip install twine 
-          twine check dist/* 
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v4
+      with:
+        python-version: "3.8"
+    - name: "Build Package"
+      run: |
+        python setup.py sdist bdist_wheel
+        pip install twine 
+        twine check dist/* 
 
     - name: Publish package distributions to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,9 +18,9 @@ jobs:
           python-version: "3.8"
       - name: "Build Package"
         run: |
-        python setup.py sdist bdist_wheel
-        pip install twine 
-        twine check dist/* 
+          python setup.py sdist bdist_wheel
+          pip install twine 
+          twine check dist/* 
 
     - name: Publish package distributions to PyPI
       uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,23 @@
+
+jobs:
+  pypi-publish:
+    name: Upload release to PyPI
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/tubular
+    permissions:
+      id-token: write  # IMPORTANT: this permission is mandatory for trusted publishing
+    steps:
+    - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.8"
+      - name: "Build Package"
+        run: |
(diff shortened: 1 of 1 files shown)


davidhopkinson26:
There is currently a syntax error on line 14 of the release.yml pipeline.  I cant quite figure out the issue and would appreciate any insights others may have.

davidhopkinson26:
> There is currently a syntax error on line 14 of the release.yml pipeline. I cant quite figure out the issue and would appreciate any insights others may have.

I think this was an issue with indents in the steps block which I have now resolved.

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v4
+      with:
+        python-version: "3.8"
+    - name: "Build Package"
+      run: |
+        python setup.py sdist bdist_wheel
+        pip install twine 
+        twine check dist/* 
+
```
adamsardar, line 24: Thought: I think that we're missing a stage that uploads the binary and source packages to the releases page. From looking at the [softprops/action-gh-release](https://github.com/softprops/action-gh-release#%EF%B8%8F-uploading-release-assets) action, I think that it's as simple as adding:

```
- name: Github Release
        uses: softprops/action-gh-release@v1
        with:
          files: dist/*
```

adamsardar (changes requested):
It's pretty much there.

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v4
+      with:
+        python-version: "3.8"
+    - name: "Build Package"
+      run: |
+        python setup.py sdist bdist_wheel
+        pip install twine 
+        twine check dist/*
```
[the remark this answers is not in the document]
davidhopkinson26: Yes agree this is questionable.   The docs for the GH action say:

"It is recommended that you run twine check just after producing your files, but this also runs twine check before upload."

I don't see why it makes sense to run the same check twice though so happy to remove for simplicity.

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+    environment:
+      name: pypi
+      url: https://pypi.org/p/tubular
+    permissions:
+      id-token: write  # IMPORTANT: this permission is mandatory for trusted publishing
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/setup-python@v4
+      with:
+        python-version: "3.8"
```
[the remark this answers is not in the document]
davidhopkinson26: [changed here](https://github.com/lvgig/tubular/pull/161/commits/73a5a5ed915c9f46d5918d3cf21b0c8ac1ea3794)

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+    - name: "Build Package"
+      run: |
+        python setup.py sdist bdist_wheel
```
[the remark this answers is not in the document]
davidhopkinson26: Thanks, good catch.  Have added a step to install wheel before build.  Agree the package build method could do with an update but out of scope for this ticket.

.github/workflows/release.yml
davidhopkinson26: I ahve set this up to run on the event that we publish to GH. At the moment that process publishes the source code on GH and then this triggered pipeline would automatically build and upload binaries to pypi.

I think if we want to use the above action to publish a release to github too we would not be able to use publication to GH as the trigger.

From looking into it I think manually trigegring the GH action as described [here](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow) and publishing to both from this workflow is probably a better approach.

.github/workflows/release.yml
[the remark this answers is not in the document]
davidhopkinson26: [removed](https://github.com/lvgig/tubular/pull/161/commits/b5c443fddb29894d8969f67a1b5db2954879bd5a)

.github/workflows/release.yml
davidhopkinson26: changed [here](https://github.com/lvgig/tubular/pull/161/commits/37297c1083bbfa29b86bfb9c20d4862855257afe)

davidhopkinson26: looks like this action had the capability to upload assets to a published release but is now [ archived ](https://github.com/actions/upload-release-asset)

davidhopkinson26:
Note that Trusted publishing has now been set up on pypi for this org/repo with release.yml as the specified workflow and the environment specified as pypi.  I have also created the pypi environment.  Currently I and @adamsardar have permissions to review and approve deployments.  We can have up to 6 people with that permission.  Alternatively we can remove the environment specification and then anyone with commit permissions will be able to trigger a release.

.github/workflows/release.yml
adamsardar: I do like where this is going - it's the protected branch only (so main once we go with #136). And, whilst I can see the sense of always creating the artefacts upon manual runs on main, do we really want every manually triggered run to publish to PyPI? Might be worth a `if tag` level to the release? Or am I misunderstanding the intention?

davidhopkinson26: I think this may become a bit of a chicken and egg situation as the GH release creates the tagged branch.  This also raises the issue that the tag name needs to be set.  By default the [GH release action](https://github.com/softprops/action-gh-release#%EF%B8%8F-uploading-release-assets) uses the [github.ref variable](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) but I'm not sure yet how to set that as the version number.

I'm tempted to remove the GH release from this action and add that in later as a separate feature.  Either going back to triggering on a GH release or a manual run of the workflow like we have currently but with an if tag block for the release portion.

adamsardar: Yeah, I get your point. Shall we remove the `Github Release` stage and go back to the `on: release` trigger? This work item relates to publishing to PyPI, so let's get that working.

davidhopkinson26: Reverted to triggering on publish and removed publish to GH step [here](https://github.com/lvgig/tubular/pull/161/commits/1e54ea1278ee49acd7f26532fb1b4e6ee09ea0db)

adamsardar (approved):

PR #162, opened 2024-01-11
Feature/nominal dtype checks in transform

TommyMatthews:
Addresses issue https://github.com/lvgig/tubular/issues/145 in part. I was unable to replicate the error I found in the generic testing where the `fit` method did not raise an appropriate error for X  not being a `pandas.DataFrame object`. 

This bug definitely exists for the `transform` method so I have fixed that - all it required was adding or moving a call to `BaseTransformer.transform`. This was always getting called, but always after the `check_mappable_rows` method inherited from `BaseNominalTransformer`, which was making checks that assumed a dataframe. This is a good example of an implementation test (super transform call) not testing for the behaviour we want (error if X not pd dataframe). I have now added a test for this behaviour (copied from BaseTransformer test suite).

The issue still mentions the possibility of adding nominal type checks (maybe with a warning) for  nominal transformers. I think probably best to get the bug fix in, close the issue and open a new issue for this feature.

### DIFF
--- a/tests/nominal/test_MeanResponseTransformer.py
+++ b/tests/nominal/test_MeanResponseTransformer.py
@@ -1023,6 +1023,19 @@ class TestTransform:
         ):
             x.transform(df)
 
+    def test_not_dataframe_error_raised(self):
+
+        df = d.create_MeanResponseTransformer_test_df()
+
+        x = MeanResponseTransformer(columns="b")
+        x.fit(df, df["a"])
+
+        with pytest.raises(
+            TypeError,
+            match=f"{x.classname()}: X should be a pd.DataFrame",
+        ):
+            x.transform(X=[1, 2, 3, 4, 5, 6])
+
     def test_super_transform_called(self, mocker):
         """Test that BaseTransformer.transform called."""
         df = d.create_MeanResponseTransformer_test_df()
--- a/tests/nominal/test_NominalToIntegerTransformer.py
+++ b/tests/nominal/test_NominalToIntegerTransformer.py
@@ -178,6 +178,19 @@ class TestTransform:
         ):
             x.transform(df)
 
+    def test_not_dataframe_error_raised(self):
+
+        df = d.create_df_1()
+
+        x = NominalToIntegerTransformer(columns=["a", "b"])
+        x.fit(df, df["a"])
+
+        with pytest.raises(
+            TypeError,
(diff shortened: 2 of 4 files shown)


davidhopkinson26:
Thanks @TommyMatthews.  I agree this is all quite messy and could do with a further reaching refactor.  Going forward I would like to see less reliance on super init/fit/transform calls and move towards the error handling being in more specific methods to be called in child classes (like you have done with the columns_check method here)

davidhopkinson26:
I don't understand why moving the super_transform call to the start broke check_mappable_rows.  Can you elaborate on why it failed please?

I'm happy with the solution proposed (i.e. direct call to columns_check) but want to understand to help with future refactoring of BaseNominalTransformer as this seems like odd behaviour.

TommyMatthews:
@davidhopkinson26 I'm not 100% sure myself. If I run `X = super().transform(X)`, or even `X = super(BaseNominalTransformer, self).transform(X)` (to explicitly access the BaseTransformer) then the value error `BaseNominalTransformer.check_mappable_rows` is raised, e.g. `ValueError: OrdinalEncoderTransformer: nulls would be introduced into column b from levels not present in mapping`

image

So I guess super transform must be overwriting the self.mappings attribute? (I can't see it changing X.shape[0]?)

TommyMatthews:
Figured this one out: hadn't fully understood the inheritance set up of the transformer but super.transform is actually calling the transform method of `BaseMappingTransformerMixin`, which is actually doing the mapping and hence `check_mappable_rows` check fails. 

I think the best solution is still just calling the `BaseTransformer.columns_check()` as `BaseTransformer.transform()` is called in `BaseMappingTransformerMixin.transform()`, so calling this at the start of transform would duplicate a lot of checks

davidhopkinson26:
> Figured this one out: hadn't fully understood the inheritance set up of the transformer but super.transform is actually calling the transform method of `BaseMappingTransformerMixin`, which is actually doing the mapping and hence if you call `check_mappable_rows` afterwards it fails
> 
> I think the best solution is still just calling the `BaseTransformer.columns_check()` as `BaseTransformer.transform()` is called in `BaseMappingTransformerMixin.transform()`, so calling this at the start of transform would duplicate a lot of checks

That makes sense. I agree that this is the best solution.

davidhopkinson26 (approved):

PR #166, opened 2024-01-18
build pipeline run on schedule

davidhopkinson26:
Update python-package.yml to trigger build pipeline on Monday morning at 7:30.  Documentation [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule)

We could go more or less frequent than this but I feel weekly runs should be sufficient to catch issues with unpinned package versions etc in a timeframe where action can be taken quickly.  Daily seems excessive but I'm happy to be convinced otherwise.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -8,6 +8,9 @@ on:
     branches: [ main, develop ]
   pull_request:
     branches: [ main, develop ]
+  schedule:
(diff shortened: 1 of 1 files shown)


adamsardar:
Weekly sounds fine given the activity in this repo. Two PRs a week is high frequency for us.

Good documentation spot!

adamsardar (approved):

PR #169, opened 2024-01-25
Feature/update pre commit

TommyMatthews:
Updates pre commit to use the latest version of ruff and not black.

Had to add a few things to ignore in ruff.toml, generally I used the project template approach for guidance. The two I'm least sure about were the following (done in order to avoid having to manually edit loads of files):
- E501 line length - apparently 88 is the line length preferred by black but with this set it was suggesting editing a lot of lines in supposedly black approved code
- E721 - forces `isinstance()` over `is type()` etc

Fixed ruff's complaint about using `int | float`  instead of just `float`. Done in a specific revert so easy to revert if we don't like this.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Format with ruff
       if: always()
       run: |
-        ruff format --diff
+        ruff format . --diff
 
     - name: Test with pytest
       if: always()
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -27,13 +27,7 @@ jobs:
     - name: Lint with ruff
       uses: chartboost/ruff-action@v1
       with:
-        version: 0.0.275
-
-    - name: Lint with black
-      if: always()
-      run: |
-        black --version
-        black --check .
+        version: 0.1.6
 
     - name: Test with pytest
       if: always()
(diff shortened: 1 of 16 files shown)


davidhopkinson26:
on the E501 override - did you take a look at what changes the formatter would make here? Keeping line length down is one of the features of Black I would quite like to preserve here. So long as code functionality is not affected (which it shouldn't be!) I would be in favour of just making the automatic changes suggested by the formatter now.

TommyMatthews:
@davidhopkinson26 there are about 800 lines flagged by this error, but they are left unchanged by the formatting and the linting - I don't think that this functionality has made it into ruff formatting yet. Perhaps we leave this PR out until it's in?

davidhopkinson26:
> @davidhopkinson26 there are about 800 lines flagged by this error, but they are left unchanged by the formatting and the linting - I don't think that this functionality has made it into ruff formatting yet. Perhaps we leave this PR out until it's in?

@TommyMatthews I think another option might be to set the line length parameter in ruff.toml to a slightly higher number to accommodate the code as is.  [docs](https://docs.astral.sh/ruff/settings/#extend).  Would suggest trying 88 first as I'm guessing ruff might be set lower by default to be getting this issue.

davidhopkinson26:
closing this PR as it is very old now

PR #173, opened 2024-02-02
updated NullIndicator to convert to int8, fixed tests

limlam96:
Updating NullIndicator to coerce return to int8. Return cols will be all 0/1, so can't see a downside to this, and will give large memory savings to outputs.

PR #175, opened 2024-02-02
Feature/one hot encoder default int8

limlam96:
Edited OneHotEncodingTransformer to default as outputting int8 types. Output cols should be all 0/1, so do not see a downside to this. Will be significant memory savings compared to default float64.


CHANGELOG.rst
```diff
@@ -16,6 +16,12 @@ Subsections for each version can be one of the following;
 
 Each individual change should have a link to the pull request after the description of the change.
 
+1.1.2 (2024-02-7)
+------------------
+Added
+^^^^^
+- Update OneHotEncodingTransformer to default to returning int8 columns https://github.com/lvgig/tubular/pull/175
```
davidhopkinson26, line 23: please update link to same format as rest of changelog

tubular/imputers.py
```diff
@@ -469,6 +469,6 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         X = super().transform(X)
```
davidhopkinson26, line 470: Did you mean to update NullIndicator in this PR as well? Happy to have both in 1 PR if you just want to abandon  #173 and update this PR to reflect that both are here

davidhopkinson26: Or assuming changes are same in both might be simplest to just merge #173 in first since it looks fine.  Should be fine so long as we don't squash commit.

davidhopkinson26 (commented):
Looks good.  Comments on duplicate changes from  #173 and changelog update needing fixing.

CHANGELOG.rst
limlam96: updated here https://github.com/lvgig/tubular/pull/175/commits/72cd91f670a7547dd1384cb3b9efa06f7cced9fb

tubular/imputers.py
limlam96: oops sorry, not sure how this happened, must have been in a rush on Friday. Sounds good, can complete 173 then revisit this one to see if there are any issues

davidhopkinson26 (approved):
Looks good

PR #177, opened 2024-02-08
edited GroupRareLevelsTransformer to forget rare categories when work…

limlam96:
…ing with categorical data

-----

Fixed issue where this transformer would 'remember' removed categories. This causes unwanted behaviour downstream, as e.g. MeanResponseTransformer will try and encode not present categories, so learn null mappings.


tubular/nominal.py
```diff
@@ -445,19 +445,31 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
                     ),
                     index=X.index,
-                ).astype(dtype_before)
+                )
+
+                X[c] = pd.Categorical(
+                    X[c],
+                    categories=categories_before.tolist() + [self.rare_level_name],
+                )
+                rare_categories = [
```
TommyMatthews, line 467: could you remove the rare_categories from categories_before (making e.g. output_categories) before creating the categorical on line 463?

Feels like it makes more sense to create the column from the minimal list rather than create first and remove later

CHANGELOG.rst
```diff
@@ -16,13 +16,18 @@ Subsections for each version can be one of the following;
 
 Each individual change should have a link to the pull request after the description of the change.
 
+1.2.1 (2024-02-08)
+------------------
+Added
+^^^^^
+- Updated GroupRareLevelsTransformer so that when working with category dtypes it forgets categories encoded as rare `#177 https://github.com/lvgig/tubular/pull/177`_
```
TommyMatthews, line 23: "...it forgets categories which have been encoded as rare and so are no longer present in the data"

^without the last bit it wasn't clear to me on first read.

limlam96: thanks, updated here! https://github.com/lvgig/tubular/pull/177/commits/227519e8d55482856b16ef437a7077e3d452037f

tubular/nominal.py
limlam96: good suggestion, implemented here https://github.com/lvgig/tubular/pull/177/commits/66034aebe3167d2521b7988d5c5de08193987f12

TommyMatthews (approved):

Issue #184, opened 2024-02-26
[Bug]: Tests fail in Py > 3.8


adamsardar:
### What happened?

When running `pytest` in a python version greater than 3.8 (i.e. 3.9, 3.10, 3.11, 3.12 etc.) the following error is thrown:

`ERROR tests/test_transformers.py - TypeError: __init__() got an unexpected keyword argument 'sparse'`

The tests complete successfully on py38.

This issue was also noticed as part of a PR recently - see #179 

### Environment

See the [CI job](https://github.com/lvgig/tubular/actions/runs/8005968684/job/21866617504) associated with the branch [179-more-comprehensive-ci-process](https://github.com/lvgig/tubular/blob/179-more-comprehensive-ci-process/.github/workflows/python-package.yml). It tests against py38 through py312 and shows the failing test on ubuntu2204.

### Minimum reproducible code

```shell
python -m pip install uv
uv venv -p 3.8 py38 
uv venv -p 3.9 py39 
          

## Python 3.8
source py38/bin/activate
uv pip install -r requirements-dev.txt
pytest . # Success
deactivate

## Python 3.9+
source py39/bin/activate
uv pip install -r requirements-dev.txt
pytest . # Fails
# ERROR tests/test_transformers.py - TypeError: __init__() got an unexpected keyword argument 'sparse'
deactivate
```

### Relevant error output

```shell
=========================== short test summary info ============================
ERROR tests/test_transformers.py - TypeError: OneHotEncoder.__init__() got an unexpected keyword argument 'sparse'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
======================== 27 warnings, 1 error in 5.54s =========================
```
See CI job for more details.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

adamsardar:
Looking through the docs, the argument `sparse` was renamed to `sparse_output` [in v1.2](https://scikit-learn.org/stable/whats_new/v1.2.html#id22) and deprecated in v1.4. 1.3.2 is the [most recent wheel built for py38,](https://pypi.org/project/scikit-learn/1.3.2/#files) whereas [py39+ have wheels for 1.4.0](https://pypi.org/project/scikit-learn/1.4.0/#files).

[...]

PR #187, opened 2024-02-27
More comprehensive CI process 

### LINKED ISSUE #179: More comprehensive CI process
**What**

Review the CI template and tweak it to more up-to-date workflows.

**Why**

The current [CI workflow](https://github.com/lvgig/tubular/blob/main/.github/workflows/python-package.yml) has a few obvious potentials for improvement:
- Only a single python version is tested (py38 - which goes [EoL in October](https://devguide.python.org/versions/)), which means that we aren't testing against more recent python versions. Any issues aren't being proactively spotted, we'll just run into them when we try to upgrade.
  - Python 3.10 (the default on Ubuntu) jammy has a bunch of failing tests. I don't know why (I only spotted it as part of a PR - switching to 3.8 'fixed' things). But we should have a way of showing that these tests fail (or pass) in a central place.
- Removal of repeated stages (black and bandit are now redundant to ruff).
- The slowest section of the the CI process is package installation via pip. We could use the new [uv package](https://github.com/astral-sh/uv) to speed this up. Fast build pipelines allow you to test more cases (e.g. python environments) in the same amount of time.
- Currently we only test on develop (obsolete as of #136) and main branches - why not all branches? CI is there to catch problems before they are problems.

**How**

Critically think about each and every stage and ask what it adds. Also consider adding new ones - such as a [build matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) for different python versions.

Note that if we change black/bandit, we should also change the pre-commit hook so that developer experience remains smooth.

### LINKED ISSUE #186: [Bug]: OneHotEncodingTransformer no long compatible with sklearn
### What happened?

A bug happened!
'sparse' is renamed to 'sparse_output' in scikit-learn, believe tubular also need to make corresponding change

TypeError: __init__() got an unexpected keyword argument 'sparse' for OneHotEncodingTransformer 

### Environment

OS Name	Microsoft Windows Server 2019 Datacenter
Version	10.0.17763 Build 17763

python 3.9.18
tubular==1.2.1 (believe same issue still exists in 1.2.2)
scikit-learn==1.4.1post1

### Minimum reproducible code

_No response_

### Relevant error output

```shell
TypeError: __init__() got an unexpected keyword argument 'sparse' for OneHotEncodingTransformer
```

### Code of Conduct

adamsardar:
- :construction_worker: CI pipeline now tests against more versions of Python: 3.8 through 3.12
  - This unearthed a bug (#184) which has been fixed so that `main` can still be passing with the above new functionality. This isn't _ideal_, as typically the bug below would have been fixed on a different branch. Reality of changing CI I feel, but if the reviewer would prefer a separate PR we can do that. 
    - :ambulance: The same issue has also come up in #186, so this is somewhat of a hotfix.
- :bug: `sparse` argument for `sklearn.OneHotEncoder` is deprecated so we now use `sparse_output`. Closes #186 and #184
  - :green_heart: Fix CI build for Python 3.9+ (or rather, for `sklearn` 1.4+, which is the most recent wheel available for py39+)
  - :arrow_up: This has forced an increase in minimum `sklearn` version to `1.2.0`.
- :sparkles: CI now runs on all branches, not just `main`. Developers can receive feedback on their contributions before PR stage.
- :zap: Using `uv` to install dependencies. The 'Install Dependencies' stage has gone from [20 seconds](https://github.com/lvgig/tubular/actions/runs/7991574591/job/21823059072) to [4 seconds](https://github.com/lvgig/tubular/actions/runs/8071233753/job/22050419301?pr=187).
 - :heavy_minus_sign: Drops black and bandit in favour of a pure ruff setup.
  - :arrow_up: Updated `ruff` to more up-to-date version, unlocking the formatter options to replace `black`.
  - :rotating_light: ruff formatter has changed a handful of files alongside `ruff autofixes`.
  - I realised too late that some of this PR duplicates some efforts in #169 . I managed to get E721 under control using `lint.preview = true` (see [docs](https://docs.astral.sh/ruff/rules/type-comparison/)), but just added E501 to ignore, so that's still there for #169 to tackle.
  - :pushpin: Explicitly setting OS to test on to `ubuntu-22.04`.
 
Closes #179.


tubular/nominal.py
```diff
@@ -549,7 +549,7 @@ def __init__(
         weights_column: str | None = None,
         prior: int = 0,
         level: str | list | None = None,
-        unseen_level_handling: str | int | float | None = None,
+        unseen_level_handling: str | float | None = None,
```
davidhopkinson26, line 552: N - is this just because int is redundant or is this a fix for something?

.github/workflows/python-package.yml
```diff
@@ -3,65 +3,80 @@
 name: Python package build and test
```
davidhopkinson26, line 3: P - big fan of these changes.  Thanks!

.ruff.toml
```diff
@@ -1,38 +1,69 @@
...
+    "YTT",
+]
 
+# Enabled to allow direct type comparisons using is and is not under E721 (see https://docs.astral.sh/ruff/rules/type-comparison/)
+lint.preview = true
 
 # ignore E501 - linelength limit (covered by black except in docstrings) 
 # and PD901 - use of df variable name
-ignore = ["E501", "PD901", "ANN101"]
+lint.ignore = ["E501", "PD901", "ANN101"]
```
davidhopkinson26, line 48: Now we have removed black we should make sure to raise an issue to get line length checking back in soon since it is being ignored in ruff (E501).  However i think not having this is worth it for the other improvements/streamlining of the CI this change brings.  Potentially this could be addressed in #169

davidhopkinson26 (approved):
Approving as happy with changes.  Would still appreciate an answer to my above question re removal of int in some type hints as I'm not clear on why this has been done but can't see a problem with it.

tubular/nominal.py
adamsardar: It's redundant.

From [PEP 484](https://peps.python.org/pep-0484/#the-numeric-tower):

> Rather than requiring that users write import numbers and then use numbers.Float etc., this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type float, an argument of type int is acceptable; similar, for an argument annotated as having type complex, arguments of type float or int are acceptable. 

Good chat in [this thread](https://stackoverflow.com/questions/50928592/mypy-type-hint-unionfloat-int-is-there-a-number-type).

.ruff.toml
adamsardar: Added ticket #190. And yes agreed that we might look at it as part of #169

tubular/nominal.py
adamsardar: Whilst this PR has been approved - I'll just wait a day or so to check that something unexpected has happened here. We aren't in a rush.

adamsardar: I had a double check of the code path and am confident that we'll be ok. I'll merge this in.

PR #189, opened 2024-02-29
Makes an inheritable DataFrameMethodTransformerInitTests class

TommyMatthews:
I realised that DataFrameMethodTransformer contains a lot of input checks in the init method that are inherited and used by it's child classes, and have made the tests for these inheritable.

As an aside, the only child class of DataFrameMethodTransformer is TwoColumnOperatorTransformer.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,7 @@ Added
 
 Changed
 ^^^^^^^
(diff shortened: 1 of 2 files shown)


Decima2014 (approved):
We picked this up separately but thanks for adding to the changelog. Changes look good to me so happy to approve, thanks for picking this up!

PR #192, opened 2024-03-04
Feature/refactor CrossColumnMappingTransformer tests

davidhopkinson26:
Applying the new testing framework to CrossColumnMappingTransformer.

Broadly very simple, although I had to modify some tests in BaseMappingTransformerInitTests to be more generally applicable by calling the minimal_attribute_dict fixture.

After struggling a little with this at first I added some documentation to explain how to use those fixtures and the interaction with  with setup_class methods.  Initially tried adding docstring explaining when it is called to every setup_class method but this was very messy, seemed excessive to repeat so much, and did not seem as helpful as having the framework explained in one place.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -222,12 +222,12 @@ def minimal_attribute_dict():
 
 
 @pytest.fixture()
-def instantiated_transformers(minimal_attribute_dict):
+def initialized_transformers(minimal_attribute_dict):
     """dictionary of {transformer name : initiated transformer} pairs for all transformers"""
     return {x[0]: x[1](**minimal_attribute_dict[x[0]]) for x in get_all_classes()}
 
 
 @pytest.fixture()
-def uninstantiated_transformers():
+def uninitialized_transformers():
     """dictionary of {transformer name : uninitiated transformer} pairs for all transformers"""
     return {x[0]: x[1] for x in get_all_classes()}
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,24 @@ import pytest
 
 import tubular.base as base
 
+"""
+How To Use This Testing Framework
+-----------------------------------
+
+Fixtures in this file are used in tests for shared behaviour which can be run on multiple transformers through inheritance.
+setup_class is defined in each test class to define cls.transformer_name as the name of the transformer being tested.
+
+Test methods for shared behaviour are defined in base_tests.py or child test classes which are themselves inherited.  The
+fixtures in this file can be used in these tests to access initiated/ uninitiated transformers and minimum attributes for
+intiatialization.
+
+Test classes which can import tests for shared behaviour inherit from an appropriate parent and define cls.transformer_name
+in setup_class so that these fixtures will retrieve the appropriate transformer instance. setup_class is called by pytest
+at a class level before its test methods are run https://docs.pytest.org/en/8.0.x/how-to/xunit_setup.html.
+
+New transformers will need to be added to minimal_attribute_dict.
+ """
+
 
 def get_all_classes():
     root = str(Path(__file__).parent.parent)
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -40,6 +58,8 @@ def get_all_classes():
 
 @pytest.fixture()
 def minimal_attribute_dict():
+    """defines minmal attributes (values) needed to initiate each transformer named (key).
+    New transformers need to be added here"""
```
TommyMatthews, line 62: S: Could also mention here that these values may need updating when a transformer is added to the new testing framework? (if there's some arg refactoring in the process)

tests/conftest.py
```diff
@@ -7,6 +7,24 @@
...
 
+"""
+How To Use This Testing Framework
+-----------------------------------
+
+Fixtures in this file are used in tests for shared behaviour which can be run on multiple transformers through inheritance.
+setup_class is defined in each test class to define cls.transformer_name as the name of the transformer being tested.
+
+Test methods for shared behaviour are defined in base_tests.py or child test classes which are themselves inherited.  The
+fixtures in this file can be used in these tests to access initiated/ uninitiated transformers and minimum attributes for
```
TommyMatthews, line 18: N/Q: I've called my fixtures (un)instantiated transformers, you have referenced them as them (un)initiated transformers. I'm not 100% sure what it should be, but I think we should be consistent. After looking around online I think we might be both wrong and want "(un)initialised".

As far as I understand it:
- Instantiation -> refers to creating a copy (instance) of an object, this is the work done by any constructor methods
- Initialisation -> passes arguments to the instance to set it up with an appropriate initial state (init)
 
I think the difference is more relevant in other OOP languages. Perhaps worth a look at whether a find and replace cleans this up?

https://realpython.com/python-class-constructor/

tests/conftest.py
```diff
@@ -7,6 +7,24 @@
 
 import tubular.base as base
 
+"""
+How To Use This Testing Framework
```
TommyMatthews, line 11: P: think this description is very good and should hopefully make the whole process clearer :)

tests/mapping/test_CrossColumnMappingTransformer.py
```diff
@@ -214,3 +164,15 @@ def test_mappings_unchanged(self):
             actual=x.mappings,
             msg="CrossColumnMappingTransformer.transform has changed self.mappings unexpectedly",
         )
```
TommyMatthews, line 166: This test is already inherited from BaseMappingTransformerTransformTests

tests/conftest.py
davidhopkinson26: Thanks for the suggestion, I think explaining how it works is enough to highlight that this might be an area to debug?  I think this suggestion might be a bit excessive.

tests/conftest.py
davidhopkinson26: simple with find and replace for instantiated to initialized: [ee8c4fa](https://github.com/lvgig/tubular/pull/192/commits/ee8c4fad11ddeedbcbbd42ef1488c06affcada92)

tests/conftest.py
davidhopkinson26: thanks, hope so!

tests/mapping/test_CrossColumnMappingTransformer.py
davidhopkinson26: removed  [00a7535](https://github.com/lvgig/tubular/pull/192/commits/00a753509eac6bc1c14f7dc672c78f10d5a99523)

tests/conftest.py
TommyMatthews: Yeah that makes sense

tests/conftest.py
TommyMatthews: cheers! :)

tests/mapping/test_CrossColumnMappingTransformer.py
TommyMatthews: thanks

TommyMatthews (approved):

TommyMatthews (approved):

Issue #191, opened 2024-03-04
jupyter notebook linting via CI


adamsardar:
**What**

Enable jupyter linting support in our CI pipelines (via ruff)

**Why**

Our examples are still part of our codebase, so much be under linting control. We used to have a stage in the pre-commit template that linted jupyter notebooks, but running it through CI is superior (single source of truth that can't be turned off).

**How**

Follow the instructions in the[ ruff docs ](https://docs.astral.sh/ruff/faq/#does-ruff-support-jupyter-notebooks) to enable jupyter linting and fix any failures.

Issue #185, opened 2024-02-26
Consider switching to the Dataframe API rather than hard-tethering to pandas


adamsardar:
**What**

[dataframe-api-compat](https://data-apis.org/dataframe-api-compat/) is an implementation of the [Python dataframe API standard](https://data-apis.org/dataframe-api/draft/index.html#index--page-root). This would permit tubular functions to operate on all dataframe libraries that implement the standard - most importantly [polars](https://pola.rs/).

**Why**

Currently the implementation of tubular is hard-tethered to pandas, and numpy-backed pandas at that for some cases (see explicit use of `np.int8` in [some functions](https://github.com/lvgig/tubular/blob/7b21a491de736a3fd9ee8069140e2809c349c478/tubular/imputers.py#L471)). This prevents users from adopting other libraries for analysis that might be better suited to their needs. In particular, the current implementation of tubular is quite memory hungry (see lvgig/tubular#182) and Polars permits so-called 'lazy-evaluation' and out-of-core computation.

**How**

Dataframe specific sections of tubular would need to be rewritten using the dataframe-agnostic API in [dataframe-api-compat](https://data-apis.org/dataframe-api-compat/basics/dataframe/). This is currently hard-blocked by the tests that inspect the way that functions are called (see lvgig/tubular#155), but with robust tests on behaviour we could look to extend support to polars. 

This item could be broken down into stages - transformers could be adapted one-by-one.

_References:_

- [Using Polars in a Pandas world](https://pythonspeed.com/articles/polars-pandas-interopability/)
- [Dataframe API Compat package documentation](https://data-apis.org/dataframe-api-compat/basics/dataframe/)
- [scikit-learn has some support for polars](https://scikit-learn.org/dev/modules/generated/sklearn.base.TransformerMixin.html#sklearn.base.TransformerMixin.set_output) as part of transform outputs.
  - See this scikit-learn/scikit-learn#25896 for some discussion on how sklearn are thinking about the polars/pandas support options.
- Also  look at data-apis/dataframe-api#346 for some insightful comments on the complexities of cross-DF support. It should make us question whether we really want to undertake such support ourselves.

davidhopkinson26:
A nice idea.  Some thoughts on the transition period if transformers adapted one by one:

Do we want to accept a period where e.g. polars compatibility is only true for a subset of transformers (thus a pipeline must be constructed only using these ones, with polars -> pandas conversion happening outside that pipeline)

or 

[...]

Issue #182, opened 2024-02-23
Profile tubular for slow operations


adamsardar:
**What**

Find out where execution time (and memory allocations) are being spent in tubular processes.

**Why**

> Premature optimization is the root of all evil (Donald Knuth)

Several suggestions have been made to improve performance in tubular (e.g. #181), but without any kind of metric to measure where time is being spent or thinking about _how_ to measure performance gains these efforts are doomed to fail.

**How**

Find some examples from the docs/examples and then time how long execution is taking (and how much memory is being allocated/reallocated). Consider using [timeit](https://docs.python.org/3/library/timeit.html) for timing and [memray](https://github.com/bloomberg/memray) for memory (linux only).

_Definition of done_ would look like some flamegraphs across realistic examples.

**References**
- [Premature Optimization](https://wiki.c2.com/?PrematureOptimization)
- [Fast csv writing for R](https://h2o.ai/blog/2016/fast-csv-writing-for-r/) a great case-study in how to optimise a library whilst profiling
- [THE FALLACY OF PREMATURE OPTIMIZATION](https://ubiquity.acm.org/article.cfm?id=1513451#:~:text=Trying%20to%20do%20the%20optimization,best%20practice%20among%20software%20engineers.)

limlam96:
Will summarise some results of this below, and set up issues separately. Just dug into the worst performing transformers initially:
1.  LogTransformer
- The numeric check here is quite slow, suspect it could be made more efficient (if it's actually needed at all?)
- Dropping the original column is also quite slow, should look into alternatives

2. OneHotEncoder

- Sklearn implementation is quite slow, and apparently pd.get_dummies is preferable:
https://stackoverflow.com/questions/53907202/why-is-sklearn-preprocessing-onehotencoder-so-slow
- The concat operation is quite slow, should look into alternatives

3. ArbitraryImputer

- Digging into this, fillna is called both in the base and child transformer, so that one is a bug and an easy fix
- fillna is generally quite slow, suggested alt here https://stackoverflow.com/questions/57954550/any-quicker-way-than-pandas-fillna

4. CappingTransformer

- Copy shows up among slowest parts, believe Lorna is working on removing these.
- Again the type check is quite slow, improve/remove

Also looked at fubular profiling, here the slowest parts are by far the copy creations (60-90% transformer runtime!), so worth looking to remove/improve. Worth highlighting, the improvements mentioned here are not near the scale of the improvement removing copies would have on fubular, so these are not going to bring us in range of fubular.

PR #197, opened 2024-03-12
181 migrate pandas usage to copy on write and remove copy arg from transforms

lsumption:
Set `pd.options.mode.copy_on_write = True`. This will become default in pandas 3.0.0 so it is recommended to turn it on now to get the benefits & ensure code is compliant with the upcoming changes (https://pandas.pydata.org/docs/dev/user_guide/copy_on_write.html). CoW stops making copies of the entire df unecessarily, instead just copying columns that change. It also prevents the original df changing when a view of the object is altered (see link for more info) & should limit the incorrect setwithcopy warnings. However it will still alter the original `df` in the following snippet, as it is not creating a view of the original `df` `df_temp = df`. Therefore, to prevent this I have enforced views to make use of CoW with the following: `df = df[df.columns]`

As the copy argument is no longer needed, I have added a warning to the copy argument of the base transformer, informing that it will soon be deprecated and updated the tests to drop this argument. I have not updated the example nbs as the notebooks were already generating errors and needed updating with the latest changes - I will open up a separate issue for this.

CoW was only added as optional in pandas 1.5.0 so I have also updated the requirements for tubular enforcing this new lower limit.

Profiling of updated changes (using same approach as in pyarrow tubular benchmarking repo)
Current results:
image
Proposed results:
image
There is a very marginal increase in the overall time to transform the df, however making these changes will allow us to implement faster options going forward e.g. polars


adamsardar:
There are some merge conflicts with `main` now - and importantly CI with ruff rather than black/bandit. `ruff format .` or `ruff . --fix` should sort most things out.

I had a look through the repo - where is `profile_tubular.py`? I think it interesting that the profiling script went up - my naive play with a couple of transformers and setting the `copy` flag to `false` cut compute time in roughly half and decreased memory allocations by 70% or so. It might be worth kicking more thorough profiling to #182?

adamsardar:
> I have not updated the example nbs as the notebooks were already generating errors and needed updating with the latest changes - I will open up a separate issue for this.

Consider hooking into #191?

tubular/base.py
```diff
@@ -64,6 +66,14 @@ def __init__(
             msg = f"{self.classname()}: verbose must be a bool"
             raise TypeError(msg)
 
+        # Keyword argument `kwarg2` is going to be dropped completely.
+        if type(copy) is int:
```
adamsardar, line 70: Q: Why test the type of copy? Wouldn't `if copy is not None` be right?

lsumption: Yep you're right, updated it in this commit [here](https://github.com/lvgig/tubular/pull/197/commits/67401d7a1a19d9e4a36902975bfa3bae83deb415)

lsumption:
I've pulled in the latest changes and fixed the merge conflicts with main. I've also included the profiling scripts. The above screenshots are from me pulling down the code and running profiling as it was easier to compare to the current main branch. But I've included it here in order to see what is being tested. I don't know if we want to include these files going forward or not

tubular/base.py
```diff
@@ -167,7 +172,7 @@ def _combine_X_y(self, X: pd.DataFrame, y: pd.Series) -> pd.DataFrame:
                 stacklevel=2,
             )
 
-        X_y = X.copy()
+        X_y = X[X.columns]
```
adamsardar, line 175: S: Would `pd.assign` work here? A quick plan in codespaces suggests that ti's much more efficient:

```python
import timeit
import pandas as pd

pd.options.mode.copy_on_write = True

df = pd.DataFrame({"student_id": [1, 2, 3], "grade": ["A", "C", "D"]})
y = pd.Series(['a', 'b', 'c'])

time_assign_col = timeit.timeit("X_y = df[df.columns]; X_y['_temporary_response'] = y.to_numpy()", globals=globals(), number=1000)
print(f'{time_assign_col*1000:.3f} microseconds/run')
# 355.369 microseconds/run

time_assign_col = timeit.timeit("X_y2 = df.assign(_temporary_response = y)", globals=globals(), number=1000)
print(f'{time_assign_col*1000:.3f} microseconds/run')
# 173.598 microseconds/run
# Twice as fast - I suspect that assign doesn't copy

X_y = df[df.columns]; X_y['_temporary_response'] = y.to_numpy()
print(X_y)
#   student_id grade _temporary_response
#0           1     A                   a
#1           2     C                   b
#2           3     D                   c

X_y2 = df.assign(_temporary_response = y)
print(X_y2)
#   student_id grade _temporary_response
#0           1     A                   a
#1           2     C                   b
#2           3     D                   c

print(df)
#   student_id grade
#0           1     A
#1           2     C
#2           3     D
```

No updates by reference - references are behaving themselves.

tubular/base.py
```diff
@@ -64,6 +66,13 @@ def __init__(
             msg = f"{self.classname()}: verbose must be a bool"
             raise TypeError(msg)
 
+        if copy is not None:
+            warnings.warn(
+                "copy argument no longer used and will be deprecated in a future release",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
```
adamsardar, line 75: S: Would a check on init for `pd.options.mode.copy_on_write` being set to `True` be of use? I doubt many users want the old behaviour (which will actually become unsupported behaviour come Pandas 3.0 - 'default and only mode' is the line).

tubular/dates.py
```diff
@@ -49,7 +49,7 @@ def cast_columns_to_datetime(self, X: pd.DataFrame) -> pd.DataFrame:
 
         self.check_columns_are_date_or_datetime(X)
 
-        temp = X[self.columns].copy()
+        temp = X[self.columns]
```
adamsardar, line 52: Q: Looking at these sorts of lines:

> Therefore, to prevent this I have enforced views to make use of CoW with the following: `df = df[df.columns]`

Are you saying that without these lines (which are copies) then frames will update by reference, even though `pd.options.mode.copy_on_write` is set to `True`? Because if so, copy on write isn't doing what it says. `X[self.columns]` is making a copy I think (would need to profile), but that defeats the whole purpose of the refactor!

tests/comparison/test_EqualityChecker.py
```diff
@@ -19,7 +19,7 @@ def test_super_init_call(self, mocker):
         expected_call_args = {
             0: {
                 "args": (),
-                "kwargs": {"columns": ["a", "b"], "verbose": False, "copy": False},
+                "kwargs": {"columns": ["a", "b"], "verbose": False},
```
adamsardar, line 22: P: Huge praise for going through and pulling out all of the `copy` args - it must have been a real chore. This keeps us in line with the pandas API, which will actively warn when people use `copy` from pandas >=3.0.

tubular/dates.py
adamsardar: I had a play and just changing:

```python
X_y = X[X.columns]

X_y["_temporary_response"] = y.to_numpy()
```

to

```python
X_y = X.assign(_temporary_response = y)
```

in `_combine_X_y` alongside removing `X_temp = X[X.columns]` in `BaseTransformer` decreased runtime by 25%:

```bash
# Unchanged
Tubular mean CPU time and SD
0.024680734599999973
0.005784166446244637
***Use cprofile to inspect transform calls
Overall Runtime:  0.040817475857000005

# Above changed
***Use time function to inspect transform calls
Tubular mean CPU time and SD
0.018341116700000028
0.006091312706994393
***Use cprofile to inspect transform calls
Overall Runtime:  0.029613771222
```

which I make to be (0.024 - 0.018)/0.024 = 25%, and that's before profiling memory (I will add these results). All of the relevant tests pass and, crucially, additional columns aren't added to the input dataframes. There are some tests that fail, but these are tests that inspect whether copies were made, and the whole point is that we are placing all of this onto pandas copy-on-write methodology.

Most important of all though is that if the plan is migrate to the `DataFrameAPI` (see #185) then we need to have as clear a layout of code as possible, with just the operations performed - no directives to make defensive copies. In the above case, we could just reparameterise the function to use [`DataFrame.assign`](https://data-apis.org/dataframe-api/draft/API_specification/dataframe_object.html#dataframe_api.DataFrame.assign) and the code should (🤞) just work with polars.

adamsardar: OK, some progress.

In the pandas docs I came across the sections [Returning a view versus a copy](https://pandas.pydata.org/pandas-docs/version/0.13/indexing.html#returning-a-view-versus-a-copy), which states:

> The rules about when a view on the data is returned are entirely dependent on NumPy. Whenever an array of labels or a boolean vector are involved in the indexing operation, the result will be a copy. With single label / scalar indexing and slicing, e.g. `df.ix[3:6]` or `df.ix[:, 'A']`, a view will be returned.

(docs seem to be a little out of date as `.ix` [has been replaced by](https://pandas.pydata.org/pandas-docs/version/0.23/generated/pandas.DataFrame.ix.html) `.iloc`).

So the fact that `X[self.columns]` returns a copy is expected behaviour given the rules of numpy. For this PR, this means that we should prefer:

```python
def transform(self, X: pd.DataFrame) -> pd.DataFrame:
       
        ...

        # Create a view of X, permitting copy-on-write semantics.  Via scalar indexing a view will be returned.
        X_view = X.iloc[:]

        return X_view
```

Over

```python
def transform(self, X: pd.DataFrame) -> pd.DataFrame:
       
        ...

        # Create a view of X, permitting copy-on-write semantics.  Via scalar indexing a view will be returned.
        X_copy = X[self.columns]

        return X_copy
```

A brief check on the `capper` example above shows that the original dataframe is not modified, but running the `profile_tubular.py` script demonstrates the ~25% time saving (and that's before we look at memory).

For this ticket then, shall we just add a test to check that the original frame is not updated upon applying a transform, then employ the `X.iloc` style to prevent triggering the copy?

However, this indexing resulting in copies raises more questions than it closes, because if an array-backed indexing operation results in a copy, then all of the lines below are copying the dataframe (!). I will guarantee that this is why the timing profiling is dominated by indexing operations (`__getitem__`, `_take_with_is_copy`, `take`). Tackling this is beyond the scope of this ticket, but it is worth thinking about decreasing this runtime burden. In order to do so with confidence, we need some checks that input frames aren't modified.

```bash
(py38) @adamsardar ➜ /workspaces/tubular/tubular (181-migrate-pandas-usage-to-copy-on-write-and-remove-copy-arg-from-transforms) $ grep -F '[self.columns' ./*
./base.py:        X[self.new_column_names] = getattr(X[self.columns], self.pd_method_name)(
./capping.py:        numeric_column_types = X[self.columns].apply(
./comparison.py:        X[self.new_col_name] = X[self.columns[0]] == X[self.columns[1]]
./dates.py:        temp = X[self.columns]
./dates.py:        temp = X[self.columns]
./dates.py:        if (pd.isna(row[self.columns[0]])) or (pd.isna(row[self.columns[1]])):
./dates.py:        age = row[self.columns[1]].year - row[self.columns[0]].year
./dates.py:            if (row[self.columns[1]].month, row[self.columns[1]].day) < (
./dates.py:                row[self.columns[0]].month,
./dates.py:                row[self.columns[0]].day,
./dates.py:        elif age < 0 and (row[self.columns[1]].month, row[self.columns[1]].day) > (
./dates.py:            row[self.columns[0]].month,
./dates.py:            row[self.columns[0]].day,
./dates.py:            temp[self.columns[1]] - temp[self.columns[0]]
./dates.py:            X[self.columns[0]],
./dates.py:                temp[self.columns[0]].dt,
./dates.py:                temp[self.columns[0]].dt,
./dates.py:        if not (X[self.columns[0]] <= X[self.columns[2]]).all():
./dates.py:            lower_comparison = X[self.columns[0]] <= X[self.columns[1]]
./dates.py:            lower_comparison = X[self.columns[0]] < X[self.columns[1]]
./dates.py:            upper_comparison = X[self.columns[1]] <= X[self.columns[2]]
./dates.py:            upper_comparison = X[self.columns[1]] < X[self.columns[2]]
./misc.py:        X[self.columns] = self.value
./misc.py:        X[self.columns] = X[self.columns].astype(self.dtype)
./nominal.py:        OneHotEncoder.fit(self, X=X[self.columns], y=y)
./nominal.py:        X_transformed = OneHotEncoder.transform(self, X[self.columns])
./numeric.py:        numeric_column_types = X[self.columns].apply(
./numeric.py:            if (X[self.columns] <= -1).sum().sum() > 0:
./numeric.py:                X[new_column_names] = np.log1p(X[self.columns])
./numeric.py:                X[new_column_names] = np.log1p(X[self.columns]) / np.log(self.base)
./numeric.py:            if (X[self.columns] <= 0).sum().sum() > 0:
./numeric.py:                X[new_column_names] = np.log(X[self.columns])
./numeric.py:                X[new_column_names] = np.log(X[self.columns]) / np.log(self.base)
./numeric.py:        if not pd.api.types.is_numeric_dtype(X[self.columns[0]]):
./numeric.py:            msg = f"{self.classname()}: {self.columns[0]} should be a numeric dtype but got {X[self.columns[0]].dtype}"
./numeric.py:            X[self.columns[0]].to_numpy(),
./numeric.py:        is_numeric = X[self.columns].apply(pd.api.types.is_numeric_dtype, axis=0)
./numeric.py:        numeric_column_types = X[self.columns].apply(
./numeric.py:        self.scaler.fit(X[self.columns])
./numeric.py:        X[self.columns] = self.scaler.transform(X[self.columns])
./numeric.py:            [self.columns[col_idx] for col_idx in interaction_combination]
./numeric.py:        numeric_column_types = X[self.columns].apply(
./numeric.py:            if 0 < self.n_components <= min(X[self.columns].shape):
./numeric.py:                msg = f"{self.classname()}: n_components {self.n_components} must be between 1 and min(n_samples {X[self.columns].shape[0]}, n_features {X[self.columns].shape[1]}) is {min(X[self.columns].shape)} with svd_solver '{self.svd_solver}'"
./numeric.py:        self.pca.fit(X[self.columns])
./numeric.py:        X[self.feature_names_out] = self.pca.transform(X[self.columns])
./strings.py:        X[self.new_column_name] = getattr(X[self.columns[0]].str, self.pd_method_name)(
./strings.py:            X[self.columns].astype(str).apply(self.separator.join, axis=1)
```

tubular/base.py
lsumption: Updated to use df.assign for temporary response field [here](https://github.com/lvgig/tubular/pull/197/commits/e15f94976aeb13ef5547b537de31b14e5f27bf28)

tubular/base.py
lsumption: Test for copy on write added [here](https://github.com/lvgig/tubular/pull/197/commits/4657f8cbeda7b61ced5f38bb6dc6b7c1077d4d90)

tubular/dates.py
lsumption: Updated to use view [here](https://github.com/lvgig/tubular/pull/197/commits/33f347f373acfcd71f99291b9d1b46e7e5959a55)
Add test to check original df not changed [here](https://github.com/lvgig/tubular/pull/197/commits/32d917b6e181136262b1607c3702e16a27d51a9b)

tubular/base.py
```diff
@@ -89,10 +98,6 @@ def __init__(
             msg = f"{self.classname()}: columns must be a string or list with the columns to be pre-processed (if specified)"
             raise TypeError(msg)
 
-        if not isinstance(copy, bool):
-            msg = f"{self.classname()}: copy must be a bool"
-            raise TypeError(msg)
-
         self.copy = copy
```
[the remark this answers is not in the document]
lsumption: I've kept it in as otherwise tests fails as transformers expect there to be a copy attribute as it's an argument in transformer init methods. Until we fully deprecate this method I don't think we're able to drop this line

tubular/base.py
lsumption: Dropped test [here](https://github.com/lvgig/tubular/pull/197/commits/fc56bc00cc194420a96fa7acc6f85f5789f4eed1) decided not to add warning to base for this behaviour as very unlikely to set argument to False by accident.

adamsardar: Agreed - if a user knows to change the copy-on-write setting, they know enough to burn their own fingers. Plus, it's a setting that is pandas specific, so wouldn't be active in a abstracted version for #185

adamsardar:
Very nice - any puts us in a good place to tackle #185

adamsardar (approved):
Approved. Discussion in thread above.

PR #198, opened 2024-03-14
Remove fillna call in ArbitraryImputer (Issue 195)

Ritunjai-Sharma:
ArbitraryImputer was calling fillna both in it's own transform method, and in the BaseImputer method.

To increase tubular's efficiency and maintainability, have removed fillna call in ArbitraryImputer and moved the logic for "casting imputer value as same dtype as original column" below the BaseImputer.transform call.

Note that the above change caused the test_impute_values_set in test_ArbitraryImputer.py to fail since the initial mock setup for BaseImputer.transform was returning a simple integer:1234. However, this causes 'TypeError: 'int' object is not superscriptable' for the following code line in transform method of ArbitraryImputer: X_transformed[c] = X_transformed[c].astype(dtype), since an integer cannot be treated as a collection. Therefore have updated test_impute_values_set to mock BaseImputer.transform to return a dataframe rather than an int

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,7 @@ Changed
 - Moved BaseTransformer, DataFrameMethodTransformer, BaseMappingTransformer, BaseMappingTransformerMixin, CrossColumnMappingTransformer and Mapping Transformer over to the new testing framework.
 - Refactored MappingTransformer by removing redundant init method.
 - Updated tests for 
+- Refactored ArbitraryImputer by removing redundant fillna call in transform method. This should increase tubular's efficiency and maintainability.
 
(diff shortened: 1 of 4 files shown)


limlam96:
This looks good to me RJ, looks like there are just a couple of ruff things failing in CI - could you fix those and then will approve? :)

limlam96 (approved):
Great, thanks RJ - approved!

limlam96 (changes requested):
actually just realised, can you please update the changelog :)

Ritunjai-Sharma:
Done! Have a look at the changelog and let me know if it's fine :)

Issue #193, opened 2024-03-06
Replace verbose=true print statements with logging


adamsardar:
**What**

Rather than rely on a `verbose` flag to dictate whether we should print output to stdout, it would be better to rely on standard [logging tools](https://docs.python.org/3/library/logging.html) and severity warning levels to control what is directed where. It would also offer greater granularity over alerts: INFO, DEBUG, WARNING, ERROR and CRITICAL. When troubleshooting a user could enable INFO level,but in prod just WARNING (and above).

**Why**

From the core [Python docs](https://docs.python.org/3/howto/logging.html#when-to-use-logging), the advice is:

If you want to:

> Display console output for ordinary usage of a command line script or program

Use `print(msg)`.

But if you want to

> Report events that occur during normal operation of a program (e.g. for status monitoring or fault investigation)

Use `logging.info()` or `logging.debug()`. There's an analogous statement for warnings and `logging.warning()`.

**How**

Here is an [example](https://github.com/lvgig/tubular/blob/main/tubular/base.py#L113):

```python
if self.verbose:
    print("BaseTransformer.fit() called")
```

Which I would suggest replacing with:

```python
import logging
logging.info('BaseTransformer.fit() called')
```

At runtime, a user can opt-in to seeing these log traces:

```python
logging.basicConfig(level=logging.DEBUG)
logging.debug('this will be displayed')
logging.info('this will not because INFO is below DEBUG')
```

The task of this ticker is therefore to find all places in the code that uses print to report application state to the user/process and replace it with logging. This would leave the `verbose` redundant and allow for its removal, simplifying the API - delete their use. This may break some implementation tests (see #155), in which case consider their removal if they are testing function call arguments (but not if they are testing behaviour of the code).

PR #201, opened 2024-03-25
edited mode imputed to work with all null columns when weights are pr…

limlam96:
…ovided

Previously this was causing an error. slightly niche scenario that was exposed by running small synthetic data samples through, but worth fixing


adamsardar (approved):
Good catch. And a small PR that I'm happy to approve without revisions.

It opens a question of whether there is a number or fraction of entries that might be considered too small to estimate a replacement value and should therefore issue a warning too. Or should that be at least logged? One for #193?

limlam96:
Thanks! :)

Yeah good point, suppose ideal behaviour in that scenario is as you say to raise a warning and leave it to user discretion whether to drop/fix/continue

Issue #165, opened 2024-01-18
Add checks to CI that Changelog has been updated and version incremented


davidhopkinson26:
When making a PR to add a feature to the package we should be updating both CHANGELOG.rst and the version number in __version.py to reflect the changes.

If this is forgotten the we need to make an additional PR to add these changes in before publishing a release, which slows down the release process or if forgotten means we create a release with incorrect documentation (Changelog) or cause the deployment to fail (version update).

Adding a check to the build pipeline that these two files have been updated should prevent this mistake.  Having this check in place could also potentially facilitate a move to continuous deployment where a successful PR into main could automatically trigger a release workflow, should we wish to do so.

The following github action may be useful [verify-changed-files](https://github.com/marketplace/actions/verify-changed-files)

PR #202, opened 2024-03-25
updated median computer to fit successfully on all null columns

limlam96:
Fixing bug where median imputer would error on all null columns

### DIFF
--- a/tubular/imputers.py
+++ b/tubular/imputers.py
@@ -191,7 +191,10 @@ class MedianImputer(BaseImputer):
                 filtered = X[X[c].notna()]
 
                 # below algorithm only works for >1 non null values
-                if len(filtered) > 0:
+                if len(filtered) <= 0:
+                    median = np.nan
+
+                else:
                     # first sort df by column to be imputed (order of weight column shouldn't matter for median)
                     filtered = filtered.sort_values(c)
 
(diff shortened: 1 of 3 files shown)


adamsardar (dismissed):
Small PR - straightforward bugfix and test.

I considered prompting for a test of arrow-backed DF functionality, but I think it not worthwhile given the massive overhaul coming with #185

Consider adding an entry to the changelog? Highlights a need for #165

tubular/imputers.py
```diff
@@ -190,17 +190,22 @@ def fit(self, X: pd.DataFrame, y: pd.Series | None = None) -> pd.DataFrame:
                 # filter out null rows so their weight doesn't influence calc
                 filtered = X[X[c].notna()]
 
-                # first sort df by column to be imputed (order of weight column shouldn't matter for median)
-                filtered = filtered.sort_values(c)
+                # below algorithm only works for >1 non null values
+                if len(filtered) > 0:
```
davidhopkinson26, line 194: Think this might read more clearly if the logic is flipped i.e. 
`if len(filtered) <=0:`
`   median = np.nan`
`else:`
`...`

Then the handling of this issue is in one place rather than sandwiched around the rest of the logic.

davidhopkinson26 (changes requested):
Looks good. See below for a suggested minor change

tubular/imputers.py
limlam96: Sure, updated here https://github.com/lvgig/tubular/pull/202/commits

davidhopkinson26 (approved):
Happy with changes

PR #236, opened 2024-04-23
moved CrossColumnMappingTransformer tests to new approach

limlam96:
Issues: [Base Class](https://github.com/lvgig/tubular/issues/214), [main class](https://github.com/lvgig/tubular/issues/215)

Created BaseCrossColumnMappingTransformer class with functionality shared throughout classes in file, integrated this into new testing approach

### DIFF
--- a/tests/mapping/test_BaseMappingTransformer.py
+++ b/tests/mapping/test_BaseMappingTransformer.py
@@ -1,4 +1,5 @@
 import re
+from collections import OrderedDict
 
 import pytest
 import test_aide as ta
@@ -101,7 +102,11 @@ class BaseMappingTransformerTransformTests(GenericTransformTests):
     Note this deliberately avoids starting with "Tests" so that the tests are not run on import.
     """
 
-    def test_mappings_unchanged(self, uninitialized_transformers):
+    def test_mappings_unchanged(
+        self,
+        minimal_attribute_dict,
+        uninitialized_transformers,
+    ):
         """Test that mappings is unchanged in transform."""
         df = d.create_df_1()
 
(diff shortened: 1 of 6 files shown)


tests/mapping/test_BaseMappingTransformer.py
```diff
@@ -110,7 +115,13 @@ def test_mappings_unchanged(self, uninitialized_transformers):
             "b": {"a": 1, "b": 2, "c": 3, "d": 4, "e": 5, "f": 6},
         }
 
-        x = uninitialized_transformers[self.transformer_name](mappings=mapping)
+        if self.transformer_name == "CrossColumnMappingTransformer":
```
davidhopkinson26, line 118: Not sure I like having conditional splits in logic for different transformers within a test like this.  Might be clearer and less coupled if specific behaviour for children is specified and tested in its own test module. Can we just overwrite the test for CrossColumnMappingTransformer? Appreciate that means duplicating code but would like to keep the logical flow of these tests consentient.

tubular/mapping.py
```diff
@@ -217,7 +217,70 @@ def transform(
         return X
 
 
-class CrossColumnMappingTransformer(BaseMappingTransformer):
+class BaseCrossColumnMappingTransformer(BaseMappingTransformer):
```
davidhopkinson26, line 220: P - good shout on pulling this out into a base class.  A case could be made for creating a new sub module for cross column transformers but for now I think this is fine.  (and these test updates hopefully make such a refactor very straightforward in future)

davidhopkinson26 (changes requested):
Looks good, changes requested to `test_mappings_unchanged`.  Let me know what you think.

davidhopkinson26:
please also update the CHANGELOG

tubular/mapping.py
limlam96: Thanks David :) agree could be worth splitting in future

limlam96:
changelog updated [here](https://github.com/lvgig/tubular/pull/236/commits/030f0b32cbdb5076b731374ba364426c1de3dc92)

tests/mapping/test_BaseMappingTransformer.py
limlam96: ah overwriting/overloading makes sense thanks, updated [here](https://github.com/lvgig/tubular/pull/236/commits/bbffa2b5b196e12d92565f0500c3f81c79479b0d)

davidhopkinson26 (approved):
Looks good

PR #240, opened 2024-04-25
Created generic two column init tests and tidied test_EqualityChecker

Chip2916:
Looking to address two outstanding issues

[204, generic init class for two column transformers](https://github.com/lvgig/tubular/issues/204
)
[205, bring EqualityChecker test in line with new format](https://github.com/lvgig/tubular/issues/205)

Hopefully this new generic class can be used for many of the date transformers, which use two columns as an input. At this stage these transformers take two separate args which are combined into a list, rather than the EqualityChecker which takes a list with two objects. After speaking with @limlam96 agreed that updating the date transformers to take as single arg sits outside this ticket. 

**'test_columns_non_list_error' seems to fail if anything other than string passed to pytest.mark.parametrize due to EqualityChecker reference to BaseTransformer which only accepts str or list, might be missing something here.**


limlam96:
Looks good, could you just add an entry to the changelog?

Chip2916:
> Looks good, could you just add an entry to the changelog?

Suree have done so now :)

limlam96 (approved):

PR #245, opened 2024-05-02
Feature/equality transformer

Chip2916:
Picked up second aspect of this ticket - https://github.com/lvgig/tubular/issues/204

Added an additional base class for two columns transformers and an associated test script which inherits from TwoColum Init in base tests. Might have missed the mark a little on this one, so have set up as a draft.

### DIFF
--- a/tubular/comparison.py
+++ b/tubular/comparison.py
@@ -33,18 +33,6 @@ class EqualityChecker(BaseTwoColumnTransformer):
     ) -> None:
         super().__init__(columns=columns, new_col_name=new_col_name, **kwargs)
 
-        if not (isinstance(columns, list)):
-            msg = f"{self.classname()}: columns should be list"
-            raise TypeError(msg)
-
-        if len(columns) != 2:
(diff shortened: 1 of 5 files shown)


limlam96:
Looks good! could you also:

- update changelog
- For any relevant transformers that have already been through the test refactor, could you update to inherit from this new base class (leave those that haven't been through the refactor for now as it will likely break lots of tests and cause a headache)

Chip2916:
Hey @limlam96 - forgot changelog again!!! Updated now, and added inheritance from new base class to equality checker transformer (don't think date transformers have been refactored to use two cols yet)

tubular/comparison.py
```diff
@@ -31,7 +31,7 @@ def __init__(
         drop_original: bool = False,
         **kwargs: dict[str, bool],
     ) -> None:
-        super().__init__(columns=columns, **kwargs)
+        super().__init__(columns=columns, new_col_name=new_col_name, **kwargs)
 
         if not (isinstance(columns, list)):
```
limlam96, line 36: can now delete lines 36:46 :)

PR #246, opened 2024-05-07
refactored dates file to prepare for testing. Edited base classes to …

limlam96:
…better follow existing base class format (validate at init/fit/transform). Made transformers more opinionated on types to reduce complexity, and standardised common arguments across transformers.

Fixed tests for new setting, but just removed e.g. assert call tests that we will no longer be interested in post test refactor.

[Issue](https://github.com/lvgig/tubular/issues/213)

### DIFF
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import datetime
-import itertools
 import warnings
 
 import numpy as np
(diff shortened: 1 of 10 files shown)


tubular/dates.py
```diff
@@ -14,132 +13,196 @@
...
+        super().__init__(columns=columns, **kwargs)
 
-        self._is_datetime64_dict = {}
+        if not (isinstance(new_column_name, str)):
+            msg = f"{self.classname()}: new_column_name should be str"
+            raise TypeError(msg)
 
-        for col in self.columns:
-            self._is_datetime64_dict[col] = pd.api.types.is_datetime64_any_dtype(X[col])
+        if not (isinstance(drop_original, bool)):
```
davidhopkinson26, line 58: S - imminent incoming change in #248  may allow inheriting this bit of logic.  Worth small change if #248 in before this one

tubular/dates.py
```diff
@@ -14,132 +13,196 @@
...
+        present_types = set(type_dict.values())
 
-        return temp
+        only_datetime_present = present_types == {datetime_type}
+        only_date_present = present_types == {date_type}
+        date_allowed = ~datetime_only
 
-    def _cast_non_matching_columns(
+        if not only_datetime_present and not (date_allowed and only_date_present):
+            msg = f"{self.classname()}: Columns fed to datetime transformers should be {allowed_types} and have consistent types, but found {present_types}. Please use ToDatetimeTransformer to standardise."
```
davidhopkinson26, line 110: P - good call on dropping the recasting and pointing to ToDateTimeTransformer.  Much simpler.

davidhopkinson26 (approved):
This looks like a good starting point for the test refactor and I agree with the changes you have made to columns arguments and on making the transformers more opinionated.

Was about to leave a comment asking where all the tests were for the BaseDateTransformers but can see you have opted to separate out the refactor of the code form the refactor of the testing which I agree is a sensible approach here.  Happy to make an exception to usual expectation of approved code being fully tested on understanding this will be addressed quickly in a follow up.

davidhopkinson26 (dismissed):

davidhopkinson26 (approved):

PR #248, opened 2024-05-09
Create BaseDropOriginalMixin Class

Chip2916:
As per ticket [here](https://github.com/lvgig/tubular/issues/243)

- Added mixin class for transformers that take 'drop_original' argument. 
- Added this class to mixin script as per weight class mixin script created during test refactor.
- Added associated tests to the base tests file rather than creating a new script specifically for this (as per weight class mixin also)
- Updated transformers to inherit from new mixin class where required
- Updated test files to inherit from new mixin test class where required and where refactor has taken place

### DIFF
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -120,6 +120,30 @@ class ColumnStrListInitTests(GenericInitTests):
             uninitialized_transformers[self.transformer_name](**args)
 
 
+class DropOriginalInitTests(GenericInitTests):
+    """
+    Tests for BaseTransformer.init() behaviour specific to when a transformer accepts a "drop_original" column.
+    Note this deliberately avoids starting with "Tests" so that the tests are not run on import.
+    """
+
+    @pytest.mark.parametrize("drop_orginal_column", (0, "a", ["a"], {"a": 10}, None))
+    def test_drop_column_arg_errors(
+        self,
+        uninitialized_transformers,
+        minimal_attribute_dict,
+        drop_orginal_column,
+    ):
+        """Test that appropriate errors are throwm for non boolean arg."""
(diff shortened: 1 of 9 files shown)


Chip2916:
Linting errors caused by resolution of merge conflicts, will set up another

tubular/nominal.py
```diff
@@ -1246,9 +1247,12 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
                     )
 
         # Drop original columns
```
davidhopkinson26, line 1249: I think the comment here could do with extending with ' if self.drop_original is true' as the functionality is not obvious without reading the called function.

tests/base_tests.py
```diff
@@ -529,6 +553,48 @@ def test_original_df_not_updated(self, initialized_transformers):
...
+
+        x.columns = ["a", "b"]
+        x.drop_original = False
+
+        df_transformed = x.transform(df)
+
+        assert ("a" in df_transformed.columns.to_numpy()) and (
+            "b" in df_transformed.columns.to_numpy()
+        ), "original columns not kept"
+
```
davidhopkinson26, line 596: S - ideally would also have a test that other columns in X  are unaffected when dropping columns addressed by the transformer.

davidhopkinson26 (changes requested):
Looks good, couple of suggestions I would like to see addressed but both minor.

tests/base_tests.py
Chip2916: Thanks for reviewing David, have pushed recommend changes :)

davidhopkinson26 (approved):
Looks good! Thanks 🙏

PR #250, opened 2024-05-13
Updated tested approach for meanimputer

Chip2916:
As per issue [here](https://github.com/lvgig/tubular/issues/228)

Not much to do on this one, most of the refactor for imputers done as part of adding WeightMixin class & refactoring the median/mode imputers in earlier tickets.

- Removed inherited tests from mean imputer test script.
- Updated changelog.


davidhopkinson26 (approved):
Brilliant, thanks for this! A great example of how much more streamlined the new testing format is!

davidhopkinson26 (approved):

PR #251, opened 2024-05-15
refactored DatetimeInfoExtractor to improve readability/condense

limlam96:
Refactored DatetimeInfoExtractor for readability reasons

### DIFF
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -871,6 +871,56 @@ class DatetimeInfoExtractor(BaseDateTransformer):
 
     """
 
+    TIME_OF_DAY = "timeofday"
+    TIME_OF_MONTH = "timeofmonth"
+    TIME_OF_YEAR = "timeofyear"
+    DAY_OF_WEEK = "dayofweek"
+
+    DEFAULT_MAPPINGS = {
+        TIME_OF_DAY: {
+            "night": range(6),  # Midnight - 6am
+            "morning": range(6, 12),  # 6am - Noon
+            "afternoon": range(12, 18),  # Noon - 6pm
+            "evening": range(18, 24),  # 6pm - Midnight
+        },
+        TIME_OF_MONTH: {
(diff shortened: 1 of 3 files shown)


tubular/dates.py
```diff
@@ -1033,37 +1023,17 @@ def _map_values(self, value: float, interval: str) -> str:
...
+            if value.is_integer():
+                value = int(value)
 
-        if np.isnan(value):
-            return np.nan
+        if isinstance(value, int) and value in self.RANGE_TO_MAP[include_option]:
+            return self.inverted_datetime_mappings[include_option][value]
 
-        return mappings[interval][value]
+        msg = f"{self.classname()}: value for {include_option} mapping in self._map_values should be an integer value in {min(self.RANGE_TO_MAP[include_option])}-{max(self.RANGE_TO_MAP[include_option])}"
```
davidhopkinson26, line 1035: S/N - feel like this might read better as an `else` even if not required for the logic to work.

davidhopkinson26 (approved):
Great improvement thanks!  Reads more clearly, as advertised.  One tiny suggestion but happy for you to ignore and merge as is if you prefer.

tubular/dates.py
limlam96: personally I don't disagree, but this has been beaten into me by ruff rules that we've added to HomeX haha https://docs.astral.sh/ruff/rules/superfluous-else-raise/

If we think we might turn on more of these in future maybe would leave?

davidhopkinson26:
> Spotted I hadn't updated changelog so pushed again :)

thanks

tubular/dates.py
davidhopkinson26: I bow to the authority of ruff in that case.  Prefer to leave stylistic choices to a linter where possible.

davidhopkinson26 (approved):

PR #252, opened 2024-05-21
started reworking mixin tests

limlam96:
Edited Mixin tests to remove inheritances - this was creating issues with method resolutions order for inheriting from multiple test classes


davidhopkinson26 (approved):
Yep, a sensible change.

Issue #257, opened 2024-06-17
[Bug]: DatetimeInfoExtractor prevents pipeline from being saved using joblib or pickle


Decima2014:
### What happened?

The DatetimeInfoExtractor stores the mappings_provided attribute as a .keys() type. If you try and save a pipeline object which has that pipeline step using joblib or pickle you get an error stating that:

"TypeError: cannot pickle 'dict_keys' object". 

This attribute should be stored as a list instead. We should maybe also add a test that our transformers can be saved as a pickle file

### Environment

scikit-learn == 1.1.3
Python == 3.9.13
joblib == 1.3.2
tubular == 1.1.0

### Relevant error output

```shell
TypeError: cannot pickle 'dict_keys' object
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

davidhopkinson26:
Thanks for raising this issue.  Should be a simple fix to switch this to a list and I agree that adding a test we can pickle the objects should be added to generic tests.  the test class base_tests.OtherBaseBehaviourTests was created for this purpose and should be inherited by all transformer test modules.

PR #258, opened 2024-06-27
:bug: set mappings_provided to list type so object is picklable

davidhopkinson26:
fix for issue #257 

Changes mappings_provided attribute to a list type rather than dict.keys() as the latter causes an error when pickling the transformer.

I have added tests that object is serialisable both to base_tests.py and the to test_DateTimeInfoExtractor.py  (as this module has not yet been updated to the new testing framework)

These currently use a slightly hacky clean up using os.remove and a try except block to avoid persisting pickle files.  I tried using the [pytest tmpdir fixture](https://docs.pytest.org/en/6.2.x/tmpdir.html) instead but it didn't play well with joblib and kept giving me unicode errors when calling joblib.dump:

` UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte`

Another approach to this test would be to directly check the dtypes of each of the transformers attributes to check they are serialisable objects i.e. one of number/ list/ string/ dict which i think should catch the issue too.  Please mention in review if you think this would be better.

At on point i had the test loading the pickle and comparing to original object, however i removed this as [comparing transformer objects with __eq__ doesn't work ](https://stackoverflow.com/questions/64964765/assert-two-sklearn-object-are-equal)and also because it was straying into unit testing joblib!  If we really wanted to do this we would need to implement something to iterate over attributes and compare one by one.


tests/base_tests.py
```diff
@@ -816,3 +818,12 @@ class OtherBaseBehaviourTests(
     Class to collect and hold tests for BaseTransformerBehaviour outside the three standard methods.
     Note this deliberately avoids starting with "Tests" so that the tests are not run on import.
     """
+
+    def test_is_serialisable(self, initialized_transformers):
+        path = "transformer.pkl"
```
Decima2014, line 823: I do think the solution of using the pytest tmp path is neater if we can get it working. I looked at the past commits you made and it looks like you used tmpdir instead of tmp_path. Wondering if you had tried both?

SO thread here discussing the differences - https://stackoverflow.com/questions/36070031/creating-a-temporary-directory-in-pytest

CHANGELOG.rst
```diff
@@ -15,7 +15,12 @@ Subsections for each version can be one of the following;
 - ``Security`` in case of vulnerabilities.
 
 Each individual change should have a link to the pull request after the description of the change.
+1.3.1 (unreleased)
+------------------
 
+Changed
+^^^^^^^
+- DatetimeInfoExtractor.mappings_provided changed from a dict.keys() object to list so transformer is serialisable. `#258 <https://github.com/lvgig/tubular/pull/258>`_
```
Decima2014, line 23: Worth mentioning that we are now testing that all transformers are serializable here?

CHANGELOG.rst
```diff
@@ -15,7 +15,12 @@ Subsections for each version can be one of the following;
 - ``Security`` in case of vulnerabilities.
 
 Each individual change should have a link to the pull request after the description of the change.
+1.3.1 (unreleased)
```
Decima2014, line 18: If we have added this version to the changelog feels like we want to update the version file to match this 1.3.1

tests/dates/test_DateTimeInfoExtractor.py
```diff
@@ -391,3 +394,14 @@ def test_correct_df_returned_datetime_input(self):
             expected,
             "incorrect dataframe returned",
         )
+
+    def test_is_serialisable(self):
```
Decima2014, line 398: I have been less on top of the latest testing changes so more of a question for my own understanding. If we are adding the generic test for serialization to the base transformer, do we need this test here too?

Decima2014 (changes requested):
Agree with the general approach of not testing the dtype of every single attribute here and comparing transformers between joblib dump and load. Would like to play around a bit more with using tmp path but if that doesnt work either then happy to move on

Thank you for picking up these changes

tests/dates/test_DateTimeInfoExtractor.py
Decima2014: Missed the note in the PR explaining why this was the case, will close this comment

tests/base_tests.py
davidhopkinson26: good call, tmp_path worked well - [commit](https://github.com/lvgig/tubular/pull/258/commits/cf55df03b931dd05e5307d5fdfee870ad0da6d05)

CHANGELOG.rst
davidhopkinson26: [fix here](https://github.com/lvgig/tubular/pull/258/commits/aad2c62f9675e2cc00b25efbd1ff046595a40c7a)

CHANGELOG.rst
davidhopkinson26: [oops](https://github.com/lvgig/tubular/pull/258/commits/c9daccfc3e2feb0bea9dac3e9f58d2c38ec58c7a)

tests/base_tests.py
Decima2014: Nice, thanks :)

Decima2014 (approved):

Decima2014:
Happy with the changes, have approved, thanks again for picking this up :)

PR #259, opened 2024-06-27
Bring GroupRareLevelsTransformer in line with new testing setup

davidhopkinson26:
addressing #227


tests/nominal/test_GroupRareLevelsTransformer.py
```diff
@@ -2,28 +2,26 @@
...
+    GenericTransformTests,
+    OtherBaseBehaviourTests,
+    WeightColumnFitMixinTests,
+    WeightColumnInitMixinTests,
+)
 from tubular.nominal import GroupRareLevelsTransformer
 
 
-class TestInit:
+class TestInit(ColumnStrListInitTests, WeightColumnInitMixinTests):
```
limlam96, line 19: Q: do these test classes not need to inherit from the BaseNominalTransformer classes?

limlam96:
oh also remember changelog!

tests/nominal/test_BaseNominalTransformer.py
```diff
@@ -29,7 +29,7 @@ def test_not_fitted_error_raised(self, initialized_transformers):
             with pytest.raises(NotFittedError):
```
limlam96, line 29: Q: will be a separate issue, but should we create a FitTestsMixin class to cover this one?

tests/nominal/test_BaseNominalTransformer.py
limlam96, line 1: a bit outside the scope of this review, but would the generic transform tests 'test_original_df_not_updated' and 'test_no_rows_error' not sit better in GenericTransformTests?

tests/nominal/test_GroupRareLevelsTransformer.py
davidhopkinson26: test_BaseNominalTransformer only defines specific tests for the transform method.  I guess there might be a case for creating placeholder test classes for BaseNominalTransformerInitTests and BaseNominalTransformeFitTests in case some were added in future and child classes would be set to inherit by default?

tests/nominal/test_BaseNominalTransformer.py
davidhopkinson26: Arguably.  Would only be inherited here an in test_BaseImputer so might actually add more complexity right now.  honestly I feel like this might be best left for now but happy to raise an issue or make a quick change if you think it would be better.

tests/nominal/test_BaseNominalTransformer.py
limlam96: thanks :)

tests/nominal/test_BaseNominalTransformer.py
limlam96: set up an issue so can look into fully at later date! https://github.com/lvgig/tubular/issues/264

tests/nominal/test_GroupRareLevelsTransformer.py
limlam96: ah sorry I was looking at the e.g. 'GenericTransformTests' format classes in that file - but that raises a point, shouldn't these classes be inheriting from test_BaseNominalTransformer.GenericTransformTests/GenericInitTests/GenericFitTests rather than the base_tests versions? then as you say would future proof it and tests added in future would be auto-inherited

davidhopkinson26: Those classes aren't modified in the test_BaseNominalTransformer module.  We would need to inherit a class here which is created in that module and inherits from the base classes, hence the suggestion we create placeholder classes above.

davidhopkinson26: I have renamed GenericBaseNominalTransformerTests to BaseNominalTransformTests class for clarity [here](https://github.com/lvgig/tubular/pull/259/commits/70892b5d7d1b76be0971a4bf98b96b796e899ef1)

davidhopkinson26: Have also made GenericNominalTransformTests inherit from GenericTransformTests and removed multi inheritance in the child.  https://github.com/lvgig/tubular/pull/259/commits/913254eaee200c082f4f33a82c1b71cfd35a6ea1

limlam96: ahh sorry think I was muddled with classes, am with you now, looks good thanks!

limlam96 (approved):

limlam96 (approved):

PR #261, opened 2024-06-27
bring NominalToIntegerTransformer in line with new testing setup

davidhopkinson26:
addressing #226 

This seems to be the only transformer with an inverse_transform method.  In this case I believe the coverage is adequate from existing tests and inherited tests from GenericTransformTests and GenericBaseNominalTransformerTests.


tests/nominal/test_NominalToIntegerTransformer.py
```diff
@@ -118,55 +72,6 @@ def expected_df_1():
...
-        with ta.functions.assert_function_call(
-            mocker,
-            tubular.base.BaseTransformer,
-            "transform",
-            expected_call_args,
-            return_value=d.create_df_1(),
-        ):
-            x.transform(df)
-
     def test_learnt_values_not_modified(self):
```
limlam96, line 75: this feels like a test that could be moved into a base class if we added a new fixture storing 'learnt_attributes' by transformer, thoughts on that? might create a ticket

tests/nominal/test_BaseNominalTransformer.py
limlam96, line 1: Q: is the transform test 'test_non_mappable_rows_raises_error' not inherited from the base class?

tests/nominal/test_NominalToIntegerTransformer.py
davidhopkinson26: That's a good idea, please do.

tests/nominal/test_BaseNominalTransformer.py
davidhopkinson26: So it is. Removed[ here](https://github.com/lvgig/tubular/pull/261/commits/2a54029691d1b3caaa4f25f120b6febb5fde6d0c)

tests/nominal/test_NominalToIntegerTransformer.py
limlam96: https://github.com/lvgig/tubular/issues/265

tests/nominal/test_NominalToIntegerTransformer.py
```diff
@@ -1,61 +1,37 @@
 import pandas as pd
 import pytest
 import test_aide as ta
+from test_BaseNominalTransformer import GenericBaseNominalTransformerTests
 
 import tests.test_data as d
-import tubular
+from tests.base_tests import (
+    ColumnStrListInitTests,
+    GenericFitTests,
```
limlam96, line 9: Same Q as on the other PR, should these generic classes be imported from test_BaseNominalTransformer?

davidhopkinson26: Have made same changes here as I made in  #259  https://github.com/lvgig/tubular/pull/261/commits/c896b06f09f8804fd8c8c3e69d7b1716964bf8be

limlam96 (approved):

limlam96 (approved):

PR #262, opened 2024-06-27
Bring MeanResponseTransformer in line with new test set up

davidhopkinson26:
Bringing MeanResponseTransformer in line with the new testing set up.

Have noticed during this update that WeightColumnFitMixinTests and WeightColumnInitMixinTests might not be being used in other transformers we have already refactored so might be worth a quick audit.


davidhopkinson26:
In the process of making this PR i created issue #269 as I found the `minimal_dataframe_lookup` method a bit clunky to interact with.

tests/nominal/test_MeanResponseTransformer.py
```diff
@@ -1141,77 +1049,6 @@ def expected_df_10():
...
-        with ta.functions.assert_function_call(
-            mocker,
-            tubular.nominal.BaseNominalTransformer,
-            "transform",
-            expected_call_args,
-            return_value=create_MeanResponseTransformer_test_df(),
-        ):
-            x.transform(df)
-
-    def test_learnt_values_not_modified(self):
```
limlam96, line 1195: Q: how come we're dropping this one?

davidhopkinson26: that was a mistake.  added back in https://github.com/lvgig/tubular/pull/262/commits/2bcaab00452ea73a41d15efc5e75c299eea62ad1

limlam96:
fyi CI is failing, but happy to approve otherwise!

davidhopkinson26:
> fyi CI is failing, but happy to approve otherwise!

Bit strange, an error seems to have appeared with NearestMeanResponseImputer.  Will investigate.

davidhopkinson26:
Had to reconfigure a few things in Imputer tests to make some shared generic tests compatible with this transformer.

- convention now is that the target is column 'a' in the default dataframes.  
- generic tests specify a y argument for fit rather than calling transformer.fit(df)

I'm wondering if it might be best to separate out GenericFitTests for the limited subset of cases where y is a compulsory argument (this and NearestMeanResponse imputer). Quite a few of the generic tests before this PR did not specify a y argument.  I think it is better that they do even though most can take y=None (and in fact default to it) as they will mostly be called in a pipeline where y will be specified so we need to assert they work in those circumstances.  Arguably a test checking those which can accept None work in those circumstances should be created.

Chip2916:
Hi David - picking up this review as Liam is on AL. On your comment regarding transformers where y is not a compulsory argument and a test to confirm they work when None is passed should be created, is there a ticket for this? If not can we get one down :)

Chip2916:
Reviewed scripts, no further comments on Liam's additional review. Agree with the changes made to standardise things across the imputer test scripts

davidhopkinson26:
> Reviewed scripts, no further comments on Liam's additional review. Agree with the changes made to standardise things across the imputer test scripts

https://github.com/lvgig/tubular/issues/287

Chip2916 (approved):
Happy to approve based on mine and Liam's reviews.

PR #267, opened 2024-07-03
Feature/ohe refactor

Chip2916:
PR to resolve following issues:

Bring OHE transformer testing setup in line [here](https://github.com/lvgig/tubular/issues/230)
Alphabetise minimum attribute dict [here](https://github.com/lvgig/tubular/issues/232)

Nothing unusual needed to be done for the OHE testing steup, drop_original mixin already exists, so really just removing redundant implementation tests and removing tests which are covered by the generic test suite. Added a single type test for the separator argument as this didn't exist.


tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -142,97 +59,13 @@ def test_fields_with_over_100_levels_error(self):
...
-        df["c_c"] = [0.0, 0.0, 1.0, 0.0, 0.0]
-
-        return df
-
-    def test_columns_check_call(self, mocker):
-        """Test the first call to BaseTransformer columns_check."""
-        df = d.create_df_1()
 
-        x = OneHotEncodingTransformer(columns="b")
+class TestTransform(DropOriginalTransformMixinTests, GenericTransformTests):
```
davidhopkinson26, line 63: Think this wants to inherit from test_BaseNominalTransformer.GenericBaseNominalTransformerTests (which in other parallel PRS #259 #261 I have renamed to GenericNominalTransformTests because the name is currently a bit confusing and not following the pattern set elsewhere)

davidhopkinson26:
reminder to update changelog :)

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -1,121 +1,38 @@
...
-    def test_one_hot_encoder_fit_called(self, mocker):
-        """Test that fit calls OneHotEncoder.fit."""
-        expected_keyword_args = {"X": d.create_df_1()[["b"]], "y": None}
-
-        df = d.create_df_1()
-
-        x = OneHotEncodingTransformer(columns="b")
-
-        mocker.patch("sklearn.preprocessing.OneHotEncoder.fit")
+    def test_separator_column_not_str_error(self):
```
davidhopkinson26, line 24: thanks for adding!

tests/test_data.py
```diff
@@ -319,6 +330,60 @@ def create_MeanResponseTransformer_test_df_unseen_levels():
     return df
 
 
+def create_OneHotEncoderTransformer_test_df_1():
```
davidhopkinson26, line 333: S: If these are specific to the OneHotEncoderTransformer tests I think it might be better to create them in that test module.

davidhopkinson26 (changes requested):

tests/test_data.py
Chip2916: Hey David - there are many functions in here which seem to be specific to only one test module e.g. 'create_datediff_test_df', 'create_date_test_df', 'create_NearestMeanResponseImputer_test_df'. Would you prefer that I just move the OHE specific functions here and then create a fast follow ticket to move the others? Or I could do them all as part of this PR?

tests/nominal/test_OneHotEncodingTransformer.py
Chip2916: No probs!

tests/nominal/test_OneHotEncodingTransformer.py
Chip2916: Hey David - might need to discuss this further. On the 'test_not_fitted_error_raised' which is inherited from 'GenericBaseNominalTransformerTests' the test fails for OHE as trying to call transform without fitting throws the following error for OHE 'AttributeError: 'OneHotEncodingTransformer' object has no attribute 'categories_''' - I came across this with some other tests but added in a fit method to resolve. 

Also the 'test_exception_raised' passes due to the behaviour of OHE, which I don't think is unexpected (although might not have thought it through properly). I know your off today so we can catch up next week!

davidhopkinson26: Options here are to either fix the code to generate the same error as the test is expecting  or to overwrite the test (i.e. create a test in the child test class with the same name that just contains` pass`, which will then supersede the test from the parent).

For test_not_fitted_error_raised I think we should add in a check that the transformer is fitted which should be caught before the call to sklearn.

For the other one, I suggest we overwrite the test.  Note this test has been given a more descriptive name in #259 `test_non_mappable_rows_exception_raised`

tests/test_data.py
davidhopkinson26: Yes please do that.  I think that fixtures should probably be co-located with tests in the scope they serve.

tests/nominal/test_OneHotEncodingTransformer.py
Chip2916: Thanks David - have addressed these comments [here](https://github.com/lvgig/tubular/pull/267/commits/d2ffa719e974345a86a86ab7b9f3893f5bc1bb26).

In terms of the 'test_exception_raised' (which I know has been renamed in another PR so will need to be resolved when merging) I have overwritten the test as suggested.

After our conversation in relation to 'test_not_fitted_error_raised ' I have used the check_is_fitted wrapper we have created for the sklearn.utils check_is_fitted function to check the attribute '_categories' exists before the .transform method is called. '_categories' is created by the Sklearn base encoder so will only be there if our call to the OHE transformer is successfully fit.

As per our conversation, this **has introduced a dependency on the sklearn OHE implementation.** Think we are both agreed that this is something we would rather catch so that if OHE is changed, we get notification rather than a change of behaviour not highlighted. Added this as a note to the changelog also.

tests/test_data.py
Chip2916: Dealt with that in [this](https://github.com/lvgig/tubular/pull/267/commits/b6488c504bf8a80e55900478ac285545b9c0d7bd) commit. If a fixture is used in just a single test have moved it to that test.

tests/nominal/test_OneHotEncodingTransformer.py
Chip2916: Renamed 'test_exception_raised' [here](https://github.com/lvgig/tubular/pull/267/commits/1e1e6e41223321c59e0ac5f1d21e239e7db91bc9) - didn't realise new naming convention was already in main!

davidhopkinson26: Thanks Sam.

davidhopkinson26 (approved):

PR #271, opened 2024-07-16
256 all configs within pyprojecttoml for a streamlined build experience

davidhopkinson26:
Draft of moving all project configs to pyproject.toml

moved into pyrpoject.toml:

- build spec (setup.py)
- ruff.toml
- . coveragerc

deleted:
- top level conftest.py
- bandit.yml

Haven't <del> yet </del> got rid of requirements.txt or requirements-dev.txt.  

- requirements-dev.txt because still using pip in the devcontainer set up and I'm not sure how to get pip to install deps from pyproject.toml dependency spec.  _Edit: now understand this could simply be done with `pip install .[dev`] in this case. Keeping pinned requirments-dev.txt as a fixed developer environment is still handy to have._

- requirements.txt because it seems like build still wants it to exist separately from pyproject.toml which I find strange and frankly I'm quite sure I am missing something.

<del>Might just move over to poetry for env managment and build to resolve the above.</del>

Moved to using pip-tools workflow, defining requirments.txt and requirements.dev by running pip compile on requirements specified in pyproject.toml which resolves the above nicely.

Doing so and fixing the versions of all dependencies has revealed an incompatibility between specified packages and python 3.12:  `scipy==1.10.1 depends on Python>=3.8,<3.12`.   Previously scipy was not specified in requirements so presumably a later version of the package was picked up when installing in python 3.12. 

Now changed to install dependiencies from pyproject.toml in build pipeline so it creates a compatible build for each python version and will flag on scheduled run if this is not possible.

It is possible to specify different package sepec depending on python version in pyproject.toml [example](https://github.com/pypa/build/blob/main/pyproject.toml) but [pip-tools pip-compile will not run for multiple versions of python or even a different version to the one it is run in](https://github.com/jazzband/pip-tools/issues/639).


pyproject.toml
```diff
@@ -0,0 +1,135 @@
...
+Repository = "https://github.com/lvgig/tubular"
+Issues = "https://github.com/lvgig/tubular/issues"
+Changelog = "https://github.com/lvgig/tubular/CHANGELOG.md"
+
+[tool.ruff]
+# Same as Black.
+line-length = 88
+
+# Assume Python 3.8.
+target-version = "py38"
```
adamsardar, line 54: S: Worth us targeting py39 explicitly now that 3.8 goes EoL in October of this year? Or should we kick that into another ticket?

pyproject.toml
```diff
@@ -0,0 +1,135 @@
...
+    "setuptools-scm[toml]>=6.0"
+   ]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "tubular"
+version = "1.3.1"
+dependencies = [
+    "pandas>=1.5.0",
+    "scikit-learn>=1.2.0"
```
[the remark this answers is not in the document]
adamsardar: Ah - answered my own question by seeing that you install directly from the pyproject file.

pyproject.toml
adamsardar: Right, I've orientated myself with the PR a bit and now understand that:

- We run CI against pythons 3.8 through 3.12 and would like to continue to support all of these versions as we have the potential for users on all pys.
- [scipy 1.10.1](https://pypi.org/project/scipy/1.10.1/) (released in Feb'23) supports 3.8<=python<3.12 and [scipy 1.11.1](https://pypi.org/project/scipy/1.11.1/) (released Jun'23) supports 3.9<=python<3.13. There is therefore no way to cover all of the python versions with a single pinned `requirements.txt`.
  - People do have debates about using multiple `requirements.txt` files (see jazzband/pip-tools#1326), but I don't really think that we should get into that.
- We still want to maintain a pinned set of requirements for users to develop against.

My proposal would be that we accept this PR's suggestion and install against `pyproject.toml` for now. We should advertise that our next release will be the last to support python 3.8 ahead of its EoL status inbound this October. At that stage, we shall drop py38 support, likely add a 3.13 testing stage and then switch to using pinned requirements again (we'll probably need to use [scipy-1.11.4](https://pypi.org/project/scipy/1.11.4/), which requires py39+).

Let's write as a spin-out ticket and then progress with this PR.

adamsardar: Right, I've orientated myself with the PR a bit and now understand that:

- We run CI against pythons 3.8 through 3.12 and would like to continue to support all of these versions as we have the potential for users on all pys.
- [scipy 1.10.1](https://pypi.org/project/scipy/1.10.1/) (released in Feb'23) supports 3.8<=python<3.12 and [scipy 1.11.1](https://pypi.org/project/scipy/1.11.1/) (released Jun'23) supports 3.9<=python<3.13. There is therefore no way to cover all of the python versions with a single pinned `requirements.txt`.
  - People do have debates about using multiple `requirements.txt` files (see jazzband/pip-tools#1326), but I don't really think that we should get into that.
- We still want to maintain a pinned set of requirements for users to develop against.

My proposal would be that we accept this PR's suggestion and install against `pyproject.toml` for now. We should advertise that our next release will be the last to support python 3.8 ahead of its EoL status inbound this October. At that stage, we shall drop py38 support, likely add a 3.13 testing stage and then switch to using pinned requirements again (we'll probably need to use [scipy-1.11.4](https://pypi.org/project/scipy/1.11.4/), which requires py39+).

Let's write as a spin-out ticket and then progress with this PR.

adamsardar: Right, I've orientated myself with the PR a bit and now understand that:

- We run CI against pythons 3.8 through 3.12 and would like to continue to support all of these versions as we have the potential for users on all pys.
- [scipy 1.10.1](https://pypi.org/project/scipy/1.10.1/) (released in Feb'23) supports 3.8<=python<3.12 and [scipy 1.11.1](https://pypi.org/project/scipy/1.11.1/) (released Jun'23) supports 3.9<=python<3.13. There is therefore no way to cover all of the python versions with a single pinned `requirements.txt`.
  - People do have debates about using multiple `requirements.txt` files (see jazzband/pip-tools#1326), but I don't really think that we should get into that.
- We still want to maintain a pinned set of requirements for users to develop against.

My proposal would be that we accept this PR's suggestion and install against `pyproject.toml` for now. We should advertise that our next release will be the last to support python 3.8 ahead of its EoL status inbound this October. At that stage, we shall drop py38 support, likely add a 3.13 testing stage and then switch to using pinned requirements again (we'll probably need to use [scipy-1.11.4](https://pypi.org/project/scipy/1.11.4/), which requires py39+).

Let's write as a spin-out ticket and then progress with this PR.

adamsardar:
OK, I think that this PR is close.

- [setup.py](https://github.com/lvgig/tubular/blob/256-all-configs-within-pyprojecttoml-for-a-streamlined-build-experience/setup.py) is still in the repo, and it can be fully replaced by settings in `pyproject.toml`
  -  playing on my local linux machine, all I needed to add to pyproject.toml was the section (build would think that 'profiling' was a package otherwise):
  ```toml
  [tool.setuptools]
  packages = ["tubular"]

  [tool.setuptools_scm]
  # can be empty if no extra settings are needed, presence enables setuptools_scm
  ```
- version number is still statically set in the config, even though `setuptools-scm` is installed as part of the build system. See their docs for details, but the TL;DR is to add `dynamic = ["version"]` under `project and then replace `tubular/__version__.py` with a variable set in `__init__.py`:
```python
try:
    __version__ = version("tubular")
except PackageNotFoundError:
    # package is not installed
    pass
```

(you need to remove the import of `__version__` as you set it dynamically). 

With these parameters in place, we can draw the tag version from git and use that to define package version:

```bash
$ git tag v2.1.0 #Not a tag in our repo - for illustration
$ python -m build
* Creating isolated environment: venv+pip...
* Installing packages in isolated environment:
  - setuptools >= 61.0
  - setuptools-scm[toml]>=6.0
  - wheel
* Getting build dependencies for sdist...
[ ... ]
Successfully built tubular-2.1.1.dev0+g3942b60.d20240719.tar.gz and tubular-2.1.1.dev0+g3942b60.d20240719-py3-none-any.whl
```
(the extra stuff `dev0+g3942b60.d20240719` is because I had unstaged code in my repo - as per [the versioning schema](https://setuptools-scm.readthedocs.io/en/stable/usage/#default-versioning-scheme)).



Otherwise, it looks good.

pyproject.toml
adamsardar: Captured as #282. Nothing needed for this PR here (unless there's a point of disagreement with above).

requirements.txt
```diff
@@ -1,2 +1,29 @@
-pandas>=1.5.0
-scikit-learn>=1.2.0
+#
```
adamsardar, line 1: Q: do we need `requirements.txt` in the repo at all now? If you delete `setup.py` and move all those details to `pyproject.toml`, then I don't think that requirements has a purpose any more (other than perhaps listing indirect dependencies and where they come from).

davidhopkinson26: listing the indirect dependencies was the intent but ultimately this will be different based on the environment tubular is being installed into so probably not that useful.  Will delete.

davidhopkinson26: https://github.com/lvgig/tubular/pull/271/commits/fad971fc3c94fa2c20e4c60c3e3757dda7f37ed9

davidhopkinson26:
> OK, I think that this PR is close.
> 
> * [setup.py](https://github.com/lvgig/tubular/blob/256-all-configs-within-pyprojecttoml-for-a-streamlined-build-experience/setup.py) is still in the repo, and it can be fully replaced by settings in `pyproject.toml`
>   
>   * playing on my local linux machine, all I needed to add to pyproject.toml was the section (build would think that the 'profiling' directory was a package otherwise):
>   
>   ```toml
>   [tool.setuptools]
>   packages = ["tubular"]
>   
>   [tool.setuptools_scm]
>   # can be empty if no extra settings are needed, presence enables setuptools_scm
>   ```
> * version number is still statically set in the config, even though `setuptools-scm` is installed as part of the build system. See their docs for details, but the TL;DR is to add `dynamic = ["version"]` under `project and then replace `tubular/**version**.py`with a variable set in`**init**.py`:
> 
> ```python
> try:
>     __version__ = version("tubular")
> except PackageNotFoundError:
>     # package is not installed
>     pass
> ```
> 
> (you need to remove the import of `__version__` as you set it dynamically).
> 
> * Delete `requirements.txt` - it's not needed anywhere now. `requirements-dev.txt` is used by the devcontainer setup.
> 
> With these parameters in place, we can draw the tag version from git and use that to define package version:
> 
> ```shell
> $ git tag v2.1.0 #Not a tag in our repo - for illustration
> $ python -m build # From within the tubular root dir
> * Creating isolated environment: venv+pip...
> * Installing packages in isolated environment:
>   - setuptools >= 61.0
>   - setuptools-scm[toml]>=6.0
>   - wheel
> * Getting build dependencies for sdist...
> [ ... ]
> Successfully built tubular-2.1.1.dev0+g3942b60.d20240719.tar.gz and tubular-2.1.1.dev0+g3942b60.d20240719-py3-none-any.whl
> $ ls dist/*
> dist/tubular-2.1.1.dev0+g3942b60.d20240719-py3-none-any.whl  dist/tubular-2.1.1.dev0+g3942b60.d20240719.tar.gz
> ```
> 
> (the extra stuff `dev0+g3942b60.d20240719` is because I had unstaged code in my repo - as per [the versioning schema](https://setuptools-scm.readthedocs.io/en/stable/usage/#default-versioning-scheme)).
> 
> (I tried the above build without `requirements.txt`/`requirements-dev.txt` and it all worked fine - I think that is was `setup.py` pulling them in before).
> 
> Otherwise, it looks good.

Thanks Adam.  Must have rushed this a bit because I thought I had removed setup.py already! Done here: https://github.com/lvgig/tubular/pull/271/commits/1e0bbf10867ea7cc9e388dee7ab41533df782222  also intended to remove setuptools-scm form this PR and add in a follow up but I have got it working in this commit: https://github.com/lvgig/tubular/pull/271/commits/1e0bbf10867ea7cc9e388dee7ab41533df782222

As a knock on of this I have removed logic in base.py that set a self.version attribute.  Can do this with importlib but this attribute doesnt get used anywhere as far as i can tell so I removed to declutter a bit.

pyproject.toml
davidhopkinson26: I agree. Best place to advertise?  tried adding a warning to __init__.py but doesn't seem to raise anything when doing an editable install in python 3.8. Tried both:

`if sys.version_info < (3,9):
    raise RuntimeError ("Support for python versions < 3.9 will be deprecated soon")`

`if sys.version_info < (3,9):
    warn ("Support for python versions < 3.9 will be deprecated soon")`
    
Maybe enough to just put it in the changelog/release notes?

tubular/__init__.py
```diff
@@ -9,4 +12,6 @@
     numeric,
     strings,
 )
-from tubular._version import __version__
+
+with suppress(ModuleNotFoundError):
```
adamsardar, line 16: P: took me a second to grok what happening here. But I like it more than my clunky suggestion that I made in the PR.

adamsardar (dismissed):
Looks good. A good simplification of the package and good foundations of the next change.

I tried creating a package via the tag mechanism and it worked. We do have to rather hope that the CD process holds, but such is the way with these things often.

Approved.

adamsardar (dismissed):
Reapproved after Merge Conflict fixes

adamsardar (approved):
third time is a charm - _version crept back in there.

PR #280, opened 2024-07-18
273 Test module for BaseDateTransformer in line with new test set up

davidhopkinson26:
- Adds Test modules for BaseDateTransformer and BaseDateTwoColumnTransformer in line with the new test set up
- created new mixin classes to handle new_column_name arguments and for cases with two columns (this required some changes to comparison.py)
- renamed new_col_name to new_column_name to be consistent across th apckage and allow use of a shared mixin to handle validation.
- changed the name of BaseDropOriginaltransformer to DropOriginalTransformer.  This affected quite a few additional files.

I had some trouble in this caused by the high degree of coupling introduced by conftest.  If there is a bug initialising one transformer it can cause virtually all tests in the package to fail.  I have created issue #285 to explain in more detail some proposed changes to explore.  This is not urgent but the current set up can make some problems hard to debug (always scary when a change seems to break hundreds of unrelated tests!)

On the plus size I went round the houses a bit on class structures and what to put in mixins vs base classes in developing this PR and the new test set up meant that I mostly didn't need to make changes to tests when doing so as tests were grouped by function rather than mimicking the class structure or testing implementation.  This was a pleasant experience and I think bodes well for future refactoring efforts going smoothly!


tests/base_tests.py
```diff
@@ -214,29 +244,6 @@ def test_list_length_error(
         ):
             uninitialized_transformers[self.transformer_name](**args)
 
-    @pytest.mark.parametrize(
```
Chip2916, line 217: Q: Just wondering, does the TestInit class in 'test_BaseTwoColumnTransformer.py' need updating to inherit 'NewColumnNameInitMixintests' as per 'test_EqualityChecker.py' now this has been removed from here? Asking as TestInit in 'test_BaseTwoColumnTransformer' inherits from 'TwoColumnListInitTests' as per - 

https://github.com/lvgig/tubular/blob/e92f8dc95cb6016a4d730fd1dc6a9977a6b46e69/tests/base/test_BaseTwoColumnTransformer.py#L9

tests/conftest.py
```diff
@@ -137,10 +137,6 @@ def minimal_attribute_dict():
         "BaseTransformer": {
             "columns": ["a"],
         },
-        "BaseTwoColumnTransformer": {
```
Chip2916, line 140: Q: Why is this removed, are we not using this anymore?

tests/dates/test_BaseDateTransformer.py
```diff
@@ -0,0 +1,153 @@
...
+        for col in ["date_col_1", "date_col_2"]:
+            df[col] = pd.to_datetime(df[col]).dt.date
+
+        for col in ["datetime_col_1", "datetime_col_2"]:
+            df[col] = pd.to_datetime(df[col])
+
+        present_types = (
+            {"datetime64", "date"} if datetime_col == 0 else {"date", "datetime64"}
+        )
+        msg = rf"Columns fed to datetime transformers should be \['datetime64', 'date'\] and have consistent types, but found {present_types}. Please use ToDatetimeTransformer to standardise"
```
Chip2916, line 108: N: Is this message a bit misleading? Just to check my understanding we want both columns to have a consistent type, but the 'should be \['datetime64', 'date'\]' suggests that we want one dt64 and one dt column?

tubular/base.py
```diff
@@ -241,50 +241,7 @@ def columns_check(self, X: pd.DataFrame) -> None:
                 raise ValueError(f"{self.classname()}: variable " + c + " is not in X")
 
 
-class BaseTwoColumnTransformer(BaseTransformer):
```
Chip2916, line 244: Q: Is this still required here - https://github.com/lvgig/tubular/blob/e92f8dc95cb6016a4d730fd1dc6a9977a6b46e69/tubular/comparison.py#L9

Chip2916: Actually, that's the next file, and can see you've changed it :)

tubular/dates.py
```diff
@@ -9,11 +9,12 @@
 import pandas as pd
 
 from tubular.base import BaseTransformer
+from tubular.mixins import DropOriginalMixin, NewColumnNameMixin, TwoColumnMixin
 
 
-class BaseDateTransformer(BaseTransformer):
+class BaseDateTransformer(DropOriginalMixin, BaseTransformer):
```
Chip2916, line 15: Q: Should NewColumnNameMixin be specified here?

tubular/mixins.py
```diff
@@ -60,6 +60,30 @@ def drop_original_column(
...
+            msg = f"{self.classname()}: new_column_name should be str"
+            raise TypeError(msg)
+
+        self.new_column_name = new_column_name
+
+
+class TwoColumnMixin:
+    """helper to validate columns when exactly two columns are required"""
+
+    def check_two_columns(self, columns: list[str] | str) -> None:
```
Chip2916, line 77: Q: Can two column input be just a single str? Would that not fail the if not (isinstance(columns, list)) test

tests/conftest.py
davidhopkinson26: yes have refactored the logic form this transformer into two mixin classes NewColumnNameMixin and TwoColumnMixin so no longer have this base class.

tests/dates/test_BaseDateTransformer.py
davidhopkinson26: This is set up to read the allowed types argument passed to the method so this error message might say datetime64 or date in it and this expression will accept either.

tests/base_tests.py
davidhopkinson26: yes it does as I moved the NewColumnNameMixin tests into the mixn tests out of TwoColumnListInitTests.  Looks like it does already inherit them to me? Might be missing something here.

Chip2916: Ahh okay, this was my first comment but can see from your comments below we are no longer using BaseTwoColumnTransformer

tests/dates/test_BaseDateTransformer.py
Chip2916: Ahh okay thanks :)

tubular/mixins.py
Chip2916: Thanks David

tubular/dates.py
Chip2916: Cool thanks :)

Chip2916 (approved):
Thanks David, happy to approve

PR #281, opened 2024-07-19
Feature/nearestmeanresponse update

Chip2916:
Linked to [this](https://github.com/lvgig/tubular/issues/211) issue.

Brought nearest mean response tests in line with new framework. Had to make very small change to minimal_dataframe_lookup function in conftest as NMI transformer requires a dataframe where columns have some nans in order to pass generic tests. 

Also added in minor error handling step in the imputer script itself, as the default value for the y parameter is None which falls over immediately as one of the first steps in fit is 'y.isna()' and None has no 'isna' attribute. 

Also overwrote 'test_fit_not_changing_data' in generic tests in the NMI specific tests scripts as base test for 'test_fit_not_changing_data' is set up without a y value being called in fit, which works fine for other transformers but not for NMI.


tests/imputers/test_NearestMeanResponseImputer.py
```diff
@@ -1,14 +1,23 @@
...
+)
+from tests.imputers.test_BaseImputer import (
+    GenericImputerTransformTests,
+)
 from tubular.imputers import NearestMeanResponseImputer
 
 
-# Dataframe used exclusively in this testing script
+# Dataframes used exclusively in this testing script
 def create_NearestMeanResponseImputer_test_df():
```
limlam96, line 21: Q: does this fill a different purpose to create_test_numeric_df_2?

tests/imputers/test_NearestMeanResponseImputer.py
```diff
@@ -24,56 +33,78 @@ def create_NearestMeanResponseImputer_test_df():
...
+        x = initialized_transformers[self.transformer_name]
+
+        with pytest.raises(
+            TypeError,
+            match="unexpected type for y, should be a pd.Series",
         ):
-            x.fit(df, df["c"])
+            x.fit(df)(columns=["c"], separator=333)
+
+    def test_fit_not_changing_data(
```
limlam96, line 83: Q: is this one not already contained in a generic class?

tests/imputers/test_NearestMeanResponseImputer.py
```diff
@@ -160,161 +158,12 @@ def test_learnt_values2(self):
...
-            x.transform(df)
-
-    @pytest.mark.parametrize(
-        ("df", "expected"),
-        ta.pandas.adjusted_dataframe_params(
-            create_NearestMeanResponseImputer_test_df(),
-            expected_df_1(),
-        ),
-    )
-    def test_nulls_imputed_correctly(self, df, expected):
```
limlam96, line 243: Q: do we not need this one? (same Q for the 2 version too)

limlam96: and 3

tests/imputers/test_NearestMeanResponseImputer.py
```diff
@@ -160,161 +158,12 @@ def test_learnt_values2(self):
...
-
-        df_transformed = x.transform(df)
-
-        ta.equality.assert_equal_dispatch(
-            expected=expected,
-            actual=df_transformed,
-            msg="Check nulls filled correctly in transform",
-        )
-
-    def test_learnt_values_not_modified(self):
```
limlam96, line 302: Q: do we not need this one?

tubular/imputers.py
```diff
@@ -413,6 +413,10 @@ def fit(self, X: pd.DataFrame, y: pd.Series | None = None) -> pd.DataFrame:
         """
         super().fit(X, y)
 
+        if not isinstance(y, pd.Series):
```
limlam96, line 416: Maybe we'll pull this out into a mixin at some point, but good for now!

tests/imputers/test_NearestMeanResponseImputer.py
```diff
@@ -24,56 +33,78 @@ def create_NearestMeanResponseImputer_test_df():
...
+
+        x.fit(df, df["c"])
+
+        ta.equality.assert_equal_dispatch(
+            expected=original_df,
+            actual=df,
+            msg="Check X not changing during fit",
+        )
 
     def test_null_values_in_response_error(self):
```
limlam96, line 103: This is also one that could be pulled out into a mixin eventually, but good for now

tests/imputers/test_NearestMeanResponseImputer.py
Chip2916: Hey Liam - is this not covered by the 'test_impute_value_unchanged' test in the 'test_BaseImputer' script?

tests/imputers/test_NearestMeanResponseImputer.py
Chip2916: Mighta been a bit hasty to get rid of these (especially 2/3) as thought they were covered by the generic tests. Can put those back in, but isn't the 'test_nulls_imputed_correctly' test in the test_NearestMeanResponseImputer script not covered by the 'test_expected_output_1' test in the 'test_BaseImputer' script?

tests/imputers/test_NearestMeanResponseImputer.py
Chip2916: This one is in the generic class but as per the PR description I moved here the NearestMeanResponseImputer requires a 'y' arg to be passed to it, if you pass it None it'll fall over. The generic class does not pass a 'y' argument  I didn't know whether it was best to change the generic test or to add one in here to overwrite?

tests/imputers/test_NearestMeanResponseImputer.py
Chip2916: Mmm the answer is no! I removed that and also got rid of 'test_learnt_values2' as it was testing the same behaviour just with two different expected results 

[here](https://github.com/lvgig/tubular/pull/281/commits/1441ed28f1b074aa0314149b4306088e2787fca4)

tests/imputers/test_NearestMeanResponseImputer.py
limlam96: ahh OK, no this seems fine - again something we might cover in a mixin but in future

tests/imputers/test_NearestMeanResponseImputer.py
limlam96: ahh sorry think you were right, they look to be adding anything above the inherited tests. happy to leave removed

limlam96 (approved):

PR #268, opened 2024-07-05
reworked log transformer tests

limlam96:
moved tests for LogTransformer to new setup

PR #284, opened 2024-07-23
Refactor ScalingTransformer to Inherit from BaseNumericTransformer and Update Tests

favouribude1:
Summary of Refactoring

    Class Inheritance: Changed inheritance from BaseTransformer to BaseNumericTransformer in the ScalingTransformer class

    Scaler Initialization: Used a dictionary mapping for scaler types to avoid repetitive if-else statements a

    Added checks for scaler_kwargs to ensure they are correctly formatted.

    fit() and transform() methods now relying on BaseNumericTransformer for numeric checks.                                                 

Modified the tests based on refactored code


tubular/numeric.py
```diff
@@ -427,7 +429,7 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         return X
 
 
-class ScalingTransformer(BaseTransformer):
+class ScalingTransformer(BaseNumericTransformer):
```
davidhopkinson26, line 432: P: thanks for the clean up here it is much nicer now!

tests/numeric/test_ScalingTransformer.py
```diff
@@ -123,7 +123,7 @@ def test_super_init_called(self, mocker):
 
 
 class TestCheckNumericColumns:
```
davidhopkinson26, line 125: B: we now set up our test classes to inherit a set of generic tests from other test modules. See the guidance in the issue #275 #268  for an example in this numeric module.  You need to set up the test classes for init, fit and transform to inherit from a parent test class and add 

   `@classmethod
    def setup_class(cls):
        cls.transformer_name = "ScalingTransformer"`
        
 to each so that the inherited tests can function (they look up initialised transformers and minimal dataframes from conftest.py using this)
 
 After doing this check_numeric will be covered by tests in BaseNumericTransformerFitTests and BaseNumericTransformerTransformTests so these tests can be deleted form here once those tests have been inherited.
 
 You can also delete any implementation tests so we are just testing functionality.  There should be far fewer tests in this module when you are done!

tests/numeric/test_ScalingTransformer.py
```diff
@@ -1,398 +1,119 @@
...
-    """Tests for the check_numeric_columns method."""
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "ScalingTransformer"
+        cls.scaler_type = "standard"
 
-    def test_exception_raised(self):
-        """Test an exception is raised if non numeric columns are passed in X."""
-        df = d.create_df_2()
+    def test_non_numeric_exception_raised(self):
```
davidhopkinson26, line 37: Q: This test is already inherited from BaseNumericTransformerFitTests so I don't think it is needed here.  Am I missing osmething that is not covered by the test in the base class? https://github.com/lvgig/tubular/blob/feature/refactor_ScalingTransformer_tests/tests/numeric/test_BaseNumericTransformer.py

davidhopkinson26 (changes requested):
Thanks for the contribution @favouribude1!  All looks good apart from one possibly redundant test I have flagged in comments. 

There are also a couple of merge conflicts to resolve and you need to update the changelog.

tests/numeric/test_ScalingTransformer.py
favouribude1: I checked the base class tests and saw that it doesn’t cover the specific error messages and checks needed for ScalingTransformer. The base test is more general and doesn’t handle the exact way ScalingTransformer raises errors with non-numeric data.

So, I kept the specific test in ScalingTransformer to make sure it’s working correctly.

davidhopkinson26: how so? as far as I can tell the handling should be the same.  fit calls super.fit which calls the same _check_numeric method so seems like something is going wrong if the error is returning something different here?

davidhopkinson26: had a bit more of a dig and it looks like we forgot to implement any tests for fit for LogTransformer so the inheritance of this test has not been tested before this.

davidhopkinson26: in this case the test doesn't work because the  `x = initialized_transformers[self.transformer_name]` takes the default initialised transformer for ScalingTransformer and that transformer has only columns = ['a'].  Since the df we want to check only has non numeric data in column b we are not throwing an error.  adding b to columns in conftest.minimal_attribute_dict['ScalingTransformer] seems to fix the issue without breaking other tests.

favouribude1: Thank you for this. I have now resolved it and it's working fine now

CHANGELOG.rst
```diff
@@ -46,6 +46,8 @@ Changed
 - DatetimeInfoExtractor.mappings_provided changed from a dict.keys() object to list so transformer is serialisable. `#258 <https://github.com/lvgig/tubular/pull/258>`_
 - Created BaseNumericTransformer class to support test refactor of numeric file `#266 <https://github.com/lvgig/tubular/pull/266>`_
 - Updated testing approach for LogTransformer `#268 <https://github.com/lvgig/tubular/pull/268>`_
+- Updated `ScalingTransformer` default columns to `["a", "b"]` in `minimal_attribute_dict` to fix test issues.
```
davidhopkinson26, line 49: I don't think this needs a line in the changelog, its just part of the below point.

davidhopkinson26: otherwise happy with changes so happy to approve once we delete this line

favouribude1: I have removed the unnecessary changelog entry

davidhopkinson26:
Thanks favour, happy with this now

davidhopkinson26 (approved):

PR #286, opened 2024-08-07
Feature/string transformer refactor

Chip2916:
Changes to update StringConcatenator and SeriesStrMethodTransformer tests to bring in line with new testing structure.

Issues: [278](https://github.com/lvgig/tubular/issues/278) and [279](https://github.com/lvgig/tubular/issues/279)

- As part of the StringConcatenator refactor, added a separator mixin, as the arg is also used by the OHE transformer. 
- For the SeriesStrMethodTransformer I had to overwrite one of the generic init tests as the 'columns' arg for the SeriesStrMethodTransformer can only accept a list of length 1, unique to this transformer. Could be we want to refactor the actual transformer more than I have done, which I can pick up if the reviewer thinks this is in scope.


davidhopkinson26 (dismissed):
I think a case could be made for creating a test class for single column input transformers but since this is the only one it is pretty thin.  Overwriting like this for one off exceptions is the intended use of the test framework.

Happy with changes, thanks!

Chip2916 (commented):
My changes approved by David, who has then resolved merge conflicts, happy with this change.

davidhopkinson26 (approved):

PR #289, opened 2024-08-08
switch to setuptools-git-versioning over setuptools-scm

davidhopkinson26:
setup tools scm automatic versioning with every commit seems useful in the context of continuous deployment but i would propose we switch to the slightly neater setuptools-git-versioning (i.e. version bump only on tags) until such a time as we aree ready to move to continuous deployment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,15 +2,15 @@
 requires = [
     "setuptools >= 61.0",
     "wheel",
-    "setuptools-scm[toml]>=6.0"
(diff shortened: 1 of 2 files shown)


davidhopkinson26:
Abandoning as current release pipeline would only publish tagged versions anyway so the patch versions created by setuptools-scm will not be published and we may as well leave ourselves ready for automatic changelogs/ continuous deployment in future.

PR #290, opened 2024-08-08
:bug: fix documentation

davidhopkinson26:
Noticed that documentation is not updating on readthedocs.

Addedd in make.bat copied from model_interpreter.  Also updated conf.py with different logic for accessing version.


Decima2014:
I think readthedocs would use this requirements file here to work which does have that sphinx_rtd_theme installed - https://github.com/lvgig/tubular/blob/main/docs/requirements.txt

davidhopkinson26:
> I think readthedocs would use this requirements file here to work which does have that sphinx_rtd_theme installed - https://github.com/lvgig/tubular/blob/main/docs/requirements.txt

Yes had just realised this and edited the description to remove confusion.  Thanks!

davidhopkinson26:
The docs still aren't building and have numerous errors so there is clearly more work to do here.

Decima2014 (changes requested):
I tried running it locally and got a similar error as to when building the docs for the model_interpreterpackage. The sphinx version is too old:

> Running Sphinx v4.0.2

Sphinx version error:
The sphinxcontrib.applehelp extension used by this project needs at least Sphinx v5.0; it therefore cannot be built with this version.

I would leave it as unpinned in requirements.txt file

davidhopkinson26:
Closing for now and deleting broken docs files in #348 with a view to fixing and reintroducing later #347

PR #293, opened 2024-09-09
Feature/pca transformer

Chip2916:
Issue [here](https://github.com/lvgig/tubular/issues/277) 

PR to refactor PCA transformer to fall in line with new testing structure

Have also added a mixin for check for numeric columns, which is shared by scaling transformer (not yet refactored)


davidhopkinson26 (approved):
Thanks @Chip2916, no issues with this.

PR #305, opened 2024-09-16
Feature/basedate tests

limlam96:
Hit some issues testing the datetime only transformers, so did a slight rework of the base class/base test setup to make generic date vs just datetime transformers less coupled

### DIFF
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -12,9 +12,13 @@ from tubular.base import BaseTransformer
 from tubular.mixins import DropOriginalMixin, NewColumnNameMixin, TwoColumnMixin
 
 
-class BaseDateTransformer(NewColumnNameMixin, DropOriginalMixin, BaseTransformer):
+class BaseGenericDateTransformer(
+    NewColumnNameMixin,
+    DropOriginalMixin,
+    BaseTransformer,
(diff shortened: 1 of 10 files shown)


tubular/dates.py
```diff
@@ -69,23 +73,25 @@ def check_columns_are_date_or_datetime(
...
             else:
-                msg = f"{self.classname()}: {col} type should be in {allowed_types} but got {X[col].dtype}"
+                col_dtype = date_type if is_date else X[col].dtype
+
+                msg = f"{self.classname()}: {col} type should be in {allowed_types} but got {col_dtype}"
                 raise TypeError(msg)
 
         present_types = set(type_dict.values())
 
         only_datetime_present = present_types == {datetime_type}
```
davidhopkinson26, line 92: Q: I think this bit of logic might be cleaned up a bit by comparing present_types to set(allowed_types)?

limlam96: good shout, updated here https://github.com/lvgig/tubular/pull/305/commits/ae94c5520a09174cdd3ff8406e2c31782ae0b5fa

davidhopkinson26 (approved):
Happy with these changes

Issue #308, opened 2024-09-24
[Bug]: OneHotEncoderTransformer broken with sklearn>1.2.1


davidhopkinson26:
### What happened?

When trying to transform using this transformer we get an error 

`AttributeError: 'OneHotEncodingTransformer' object has no attribute '_drop_idx_after_grouping'`

This attribute was added in sklearn> 1.2.1 and is set in init for OneHotEncoder.  This should be fine but in OneHotEncoderTransformer it looks like we do not actually persist the initialised object, instead calling 

in init
```
# Set attributes for scikit-learn's OneHotEncoder
        OneHotEncoder.__init__(
            self,
            sparse_output=False,
            handle_unknown="ignore",
            dtype=dtype,
            **kwargs,
        )
```
in fit

` OneHotEncoder.fit(self, X=X[self.columns], y=y)`
 
 and in transform:
 
` X_transformed = OneHotEncoder.transform(self, X[self.columns])`

So the intialisation in this classes init method does not actually persist to he object called in transform. Up to now this has run as just calling the transform method without any initialisation worked ok (however it may not have been functioning as expected!).  However with the newer versions of sklearn the method looks for an attribute which does not exist (a) because it is called within the wrong class so self directs it to OneHotEncoderTransformer and (b) because the OneHotEncoder has not been initialised.

To fix this we should assign an instance of OneHotEncoder to an attribute of OneHotEncoderTransformer, say '_encoder',  and persist this through init, fit and transform.

### Environment

sklearn 1.5.2
tubular 1.3.1

### Minimum reproducible code

```shell
x = OneHotCoderTransformer()

x.transform(X)
```

### Relevant error output

```shell
`AttributeError: 'OneHotEncodingTransformer' object has no attribute '_drop_idx_after_grouping'`
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

davidhopkinson26:
The way it is now relies on the calls to init and fit methods in OneHotEncoder setting attributes in the OneHotEncodeingTransformer.  This is a very unsafe way to do things (as we have just found out!) but has probably been working sort of ok up to now.  Not quite sure why the call to init is not setting the  '_drop_idx_after_grouping' attribute but still expect that swithcing to initialising and persisting a OneHotEncoder object will be a safer way for this to work.

davidhopkinson26:
closed with #309

PR #309, opened 2024-09-24
308 ohe bugfix

alphanumericmale:
PR for bugfix for OHE transformer class. This update aims to allow for the acceptance of sklearn _drop_idx_after_grouping attribute. This will in turn allow OHE to be used with versions of sklearn > 1.2.1

Issue here: https://github.com/lvgig/tubular/issues/308

### DIFF
--- a/tests/nominal/test_OneHotEncodingTransformer.py
+++ b/tests/nominal/test_OneHotEncodingTransformer.py
@@ -2,7 +2,6 @@ import numpy as np
 import pandas as pd
 import pytest
 import test_aide as ta
-from sklearn.exceptions import NotFittedError
 from test_BaseNominalTransformer import GenericNominalTransformTests
 
 import tests.test_data as d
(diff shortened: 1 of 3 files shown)


davidhopkinson26:
Have just noted that running tests on main with sklearn 1.3.2 does not seem to fail which suggest the issue encountered in #308 is not as reproducible as first thought.

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -264,3 +265,76 @@ def test_unseen_categories_encoded_as_all_zeroes(self, df_test, expected):
...
+    @pytest.fixture
+    def sample_data(self):
+        return pd.DataFrame(
+            {
+                "color": ["red", "blue", "green", "blue", "red"],
+                "size": ["S", "M", "L", "S", "M"],
+            },
+        )
+
+    def test_not_fitted_error_transform(self, sample_data):
```
davidhopkinson26, line 282: duplicate of GenericNominalTransformTests.test_not_fitted_error_raised

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -264,3 +265,76 @@ def test_unseen_categories_encoded_as_all_zeroes(self, df_test, expected):
...
+
+    def test_initialization_and_fitting(self, sample_data):
+        transformer = OneHotEncodingTransformer(columns=["color", "size"])
+        transformer.fit(sample_data)
+        assert hasattr(
+            transformer._encoder,
+            "categories_",
+        ), "The encoder should have 'categories_' attribute after fitting."
+
+    def test_transformation(self, sample_data):
```
davidhopkinson26, line 295: We already test outputs of transform in TestTransform.  Don't think this is adding anything new?  if it is then it should be in that test class with amore descriptive name.

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -264,3 +265,76 @@ def test_unseen_categories_encoded_as_all_zeroes(self, df_test, expected):
...
+            "color_green",
+            "size_S",
+            "size_M",
+            "size_L",
+        ]
+        assert all(
+            col in transformed.columns for col in expected_columns
+        ), "All expected columns should be in the transformed dataframe."
+
+    def test_unseen_categories(self, sample_data):
```
davidhopkinson26, line 311: again, duplication of tests in TestTransform

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -264,3 +265,76 @@ def test_unseen_categories_encoded_as_all_zeroes(self, df_test, expected):
...
+            "color_green",
+            "size_S",
+            "size_M",
+            "size_L",
+        ]
+        assert all(
+            col in transformed.columns for col in expected_columns
+        ), "All expected columns should be in the transformed dataframe."
+
+    def test_bug_fix(self, sample_data):
```
davidhopkinson26, line 334: I think this is a duplicate as well

davidhopkinson26 (changes requested):
Despite not being sure about the necessity of this change I think the set up is more sensible than what we had previously (direct calls to OneHotEncoder methods).  Arguably a still better approach might be to inherit from OnehotEncoder but that does risk some complex interactions form multiinheritance.

I do not think the additional tests you have added are needed though.

tests/nominal/test_OneHotEncodingTransformer.py
alphanumericmale: removed

tests/nominal/test_OneHotEncodingTransformer.py
alphanumericmale: remove

tests/nominal/test_OneHotEncodingTransformer.py
alphanumericmale: removed

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -264,3 +265,76 @@ def test_unseen_categories_encoded_as_all_zeroes(self, df_test, expected):
...
+                "size": ["S", "M", "L", "S", "M"],
+            },
+        )
+
+    def test_not_fitted_error_transform(self, sample_data):
+        transformer = OneHotEncodingTransformer(columns=["color", "size"])
+        with pytest.raises(NotFittedError):
+            transformer.transform(sample_data)
+
+    def test_initialization_and_fitting(self, sample_data):
```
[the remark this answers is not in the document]
alphanumericmale: removed

tests/nominal/test_OneHotEncodingTransformer.py
alphanumericmale: removed

davidhopkinson26 (approved):
Thanks @alphanumericmale!

davidhopkinson26 (changes requested):
One last thing - could you update the changelog please?

davidhopkinson26 (approved):
thanks!

PR #326, opened 2024-10-09
added narwhals/polars to env

limlam96:
preliminary for narwhals milestones


.github/workflows/release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v4
       with:
-        python-version: "3.x"
+        python-version: ">=3.9"
```
davidhopkinson26, line 18: Q: curious why this needed to change?  Just conscious that the release pipeline changes are untested here.

.devcontainer/devcontainer.json
```diff
@@ -3,7 +3,7 @@
 {
 	"name": "Python 3",
 	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-	"image": "mcr.microsoft.com/devcontainers/python:0-3.8",
+	"image": "mcr.microsoft.com/devcontainers/python:0-3.10",
```
davidhopkinson26, line 6: Q: wondering if we should develop in the lowest supported version of python.  Running things like this we could introduce breaking changes that don't get picked up until the CI pipeline is run on 3.9, which could be a painful workflow.  This might not be too much of a big deal in practice but the need to ignore 2 ruff rules due to differences between 3.9/3.10 feels like a bit of a sign.  Thoughts?

davidhopkinson26 (changes requested):
some questions on changes to release pipeline and assumed python verison.

.github/workflows/release.yml
limlam96: This was because the env setup failed on 3.8, looking into it I don't think newest polars supports 3.8 https://pypi.org/project/polars/

.devcontainer/devcontainer.json
limlam96: yeah had thought about that so happy to change https://github.com/lvgig/tubular/pull/326/commits/2843ed4263ed83a650a7e329f132ba45b3bab5b4

Those ruff rules are a bit odd, they still get recommended for 3.9 despite erroring

davidhopkinson26 (approved):
Happy with changes

PR #328, opened 2024-10-10
Feautre/narwhals base transformer

limlam96:
narwhalify-d basetransformer https://github.com/lvgig/tubular/issues/313

Note, polars is not interested in default indexes in the same way that pandas is, so have dropped tests around this.

should complete after env PR is resolved https://github.com/lvgig/tubular/pull/326

### DIFF
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -333,7 +333,7 @@ class GenericFitTests:
 
         with pytest.raises(
             TypeError,
-            match=f"{self.transformer_name}: X should be a pd.DataFrame",
+            match=f"{self.transformer_name}: X should be a polars or pandas DataFrame/LazyFrame",
         ):
             x.fit(non_df, df["a"])
 
@@ -352,7 +352,7 @@ class GenericFitTests:
 
         with pytest.raises(
             TypeError,
-            match=f"{self.transformer_name}: unexpected type for y, should be a pd.Series",
+            match=f"{self.transformer_name}: unexpected type for y, should be a polars or pandas Series",
         ):
             x.fit(X=df, y=non_series)
 
@@ -621,7 +621,7 @@ class GenericTransformTests:
 
         with pytest.raises(
             TypeError,
-            match=f"{self.transformer_name}: X should be a pd.DataFrame",
+            match=f"{self.transformer_name}: X should be a polars or pandas DataFrame/LazyFrame",
         ):
             x_fitted.transform(X=non_df)
 
@@ -748,7 +748,7 @@ class ColumnsCheckTests:
 
         with pytest.raises(
             TypeError,
-            match=f"{self.transformer_name}: X should be a pd.DataFrame",
+            match=f"{self.transformer_name}: X should be a polars or pandas DataFrame/LazyFrame",
         ):
             x.columns_check(X=[1, 2, 3, 4, 5, 6])
 
@@ -806,7 +806,7 @@ class CombineXYTests:
 
         with pytest.raises(
             TypeError,
-            match=f"{self.transformer_name}: X should be a pd.DataFrame",
+            match=f"{self.transformer_name}: X should be a polars or pandas DataFrame/LazyFrame",
         ):
             x._combine_X_y(X=non_df, y=pd.Series([1, 2]))
 
@@ -822,7 +822,7 @@ class CombineXYTests:
 
         with pytest.raises(
             TypeError,
-            match=f"{self.transformer_name}: y should be a pd.Series",
+            match=f"{self.transformer_name}: y should be a polars or pandas Series",
         ):
             x._combine_X_y(X=pd.DataFrame({"a": [1, 2]}), y=non_series)
 
@@ -841,23 +841,6 @@ class CombineXYTests:
             ),
         ):
             x._combine_X_y(X=pd.DataFrame({"a": [1, 2]}), y=pd.Series([2]))
-
-    def test_X_and_y_different_indexes_warning(
-        self,
-        initialized_transformers,
-    ):
-        """Test a warning is raised if X and y have different indexes, but the output is still X and y."""
-
-        x = initialized_transformers[self.transformer_name]
-
-        with pytest.warns(
-            UserWarning,
-            match=f"{self.transformer_name}: X and y do not have equal indexes",
-        ):
-            x._combine_X_y(
-                X=pd.DataFrame({"a": [1, 2]}, index=[1, 2]),
-                y=pd.Series([2, 4]),
-            )
 
     def test_output_same_indexes(
         self,
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -281,6 +281,11 @@ class GenericFitTests:
     Note this deliberately avoids starting with "Tests" so that the tests are not run on import.
     """
 
+    @pytest.mark.parametrize(
+        "minimal_dataframe_lookup",
+        ["pandas", "polars"],
+        indirect=True,
+    )
     def test_fit_returns_self(
         self,
         initialized_transformers,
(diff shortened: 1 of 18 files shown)


davidhopkinson26:
I think we might need to modify tests more extensively? tests in base_tests at the moment using the minimal dataframe lookup which gives a pandas df.  Currently we are not testing functionality with a polars dataframes passed?  I made a separate ticket to figure this out https://github.com/lvgig/tubular/issues/318

Ideally want to find a solution here which minimises the overhead.

limlam96:
added a rough POC for how we might approach this, haven't added docstrings etc yet as still a point of discussion

These cases were all simple-ish, where tests do more manipulation of dfs we would have to try converting to narwhals frames and working just with methods within the narwhals package

tests/conftest.py
```diff
@@ -334,9 +336,43 @@ def minimal_dataframe_lookup() -> dict[str, pd.DataFrame]:
...
 
+    if library == "polars":
+        for key in min_df_dict:
+            min_df_dict[key] = pl.from_pandas(min_df_dict[key])
+
     return min_df_dict
 
 
+@pytest.fixture
+def get_assert_frame_equal(request):
```
davidhopkinson26, line 347: if we pass the frame to this function it can lookup the type to get the appropriate library so we don't have to parameterise it?

tests/base_tests.py
```diff
@@ -281,6 +281,11 @@ class GenericFitTests:
     Note this deliberately avoids starting with "Tests" so that the tests are not run on import.
     """
 
+    @pytest.mark.parametrize(
```
davidhopkinson26, line 284: P: quite like this!

tests/base_tests.py
```diff
@@ -372,10 +392,16 @@ def test_X_no_rows_error(
         ):
             x.fit(df, df["a"])
 
+    @pytest.mark.parametrize(
+        ("minimal_dataframe_lookup", "get_series_init"),
```
davidhopkinson26, line 396: since this is just for this test it might make more sense to parameterise the series init methods here rather than parameterize the call to the fixture?

limlam96: yeah fair enough, implemented as an if/else inside the test https://github.com/lvgig/tubular/pull/328/commits/aef866eecd25b4db84d12765cfb8684d09996ae1#diff-09e907e492bc399633765530bd5f9a6d3a4d867e068002e04b79d26bd6f17f68

limlam96:
added a 'narwhalified_transformers_dict' fixture that we can use as a stopgap to control whether polars tests are run on a given transformer:
https://github.com/lvgig/tubular/pull/328/commits/aef866eecd25b4db84d12765cfb8684d09996ae1#diff-09e907e492bc399633765530bd5f9a6d3a4d867e068002e04b79d26bd6f17f68

tests/conftest.py
limlam96: updated this https://github.com/lvgig/tubular/pull/328/commits/fd33732623e0650fc284db5935468a547ff753ca

limlam96:
> I wonder if there might be value in a function that asserts equality of actual to expected dataframe after converting the expected to the format of the actual? A prize to whoever can come up with a good name for that one! This would allow us to just parameterise the inputs in a lot of cases.

hmm, wonder if this would make things easier at the cost of testing not quite the right thing though - because in practice if we input polars we wouldn't generally convert back to pandas at the end right?

tests/conftest.py
davidhopkinson26: think this is better but the call to get_assert_frame_equal(df)(df, original_df) is a bit clunky.  Might be better to refactor such that the function runs the check rather than just retuning a function so the call is just get_assert_frame_equal(original_df, df).  Could potentially just wrap the current get_assert_frame_equal_fucntion to achieve this.

If doing so it would probably make sense to name the function to just assert_frame_equal but for the conflict with the pd/pl method names. Maybe that's not really an issue to worry about though?

If it is then maybe: assert_frame_equal_dispatch? assert_pl_or_pd_frames_equal? assert_frameT_equal?

tests/conftest.py
```diff
@@ -91,6 +91,61 @@ def get_all_classes(
     return set(all_classes)
 
 
+@pytest.fixture()
+def narwhalified_transformers_dict():
```
davidhopkinson26, line 95: Just raising this as we are setting precedents for a big piece of work so want to evaluate options.

Another way to do this would be to set cls.narwhalified as an attribute in the test class for a method.  I think this dict is better as it has the added bonus of showing us a list all in one place of which ones have been converted.

advantage of cls attribute would be that we wouldn't need to keep importing the fixture.  I guess it would also mean each method could be handled separately so could potentially refactor transform but leave fit for later (not sure if this would actually work).  Cant see this being terribly useful even if it did work though.

limlam96: to be fair, I could be open to the cls attribute as then info on whether/not narwhalified is less hidden from users... but yes would lose that nice collated view of which had been updated/have more places to keep track of updating. Happy for you to have final call, what do you think?

davidhopkinson26:
> > I wonder if there might be value in a function that asserts equality of actual to expected dataframe after converting the expected to the format of the actual? A prize to whoever can come up with a good name for that one! This would allow us to just parameterise the inputs in a lot of cases.
> 
> hmm, wonder if this would make things easier at the cost of testing not quite the right thing though - because in practice if we input polars we wouldn't generally convert back to pandas at the end right?

So idea is that we are inputting pandas, and thus getting a pandas actual, then the expected might be defined in the test as a polars df but in the function comparing the two converts the actual polars df to a pandas df before comparing.  
 
There is a risk associated with mutating the expected as it is then not clearly defined (so maybe an int16 changes to an int 32 during conversion for example).  I guess it really depends on how safe that conversion is.

I started typing this out to try and justify the approach but find myself turned around! Still don't like the idea of duplicating so many expected outputs as it feels quite bloated but not defining them explicitly is unlikely to be a good idea.

tests/conftest.py
limlam96: cool think that is better, actioned here https://github.com/lvgig/tubular/pull/328/commits/e3114d99e18b8615a28d30edb3e107e20de385b2

tests/conftest.py
limlam96: updated here, went with assert_frame_equal_dispatch https://github.com/lvgig/tubular/pull/328/commits/fd39b7dced8f724dd5b35d410a595de21635f01f

tests/conftest.py
limlam96: updated the attr name here https://github.com/lvgig/tubular/pull/328/commits

davidhopkinson26 (approved):
happy with changes.  Thanks @limlam96!

tubular/base.py
```diff
@@ -120,10 +131,8 @@ def fit(self, X: pd.DataFrame, y: pd.Series | None = None) -> BaseTransformer:
             raise ValueError(msg)
 
         if y is not None:
-            if not isinstance(y, pd.Series):
-                msg = (
-                    f"{self.classname()}: unexpected type for y, should be a pd.Series"
-                )
+            if not isinstance(y, nw.Series):
+                msg = f"{self.classname()}: unexpected type for y, should be a polars or pandas Series"
```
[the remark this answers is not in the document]
limlam96: Thanks Francesco! these error messages are legacy really, so happy for a good opportunity to delete - will pick this up :)

tubular/base.py
```diff
@@ -5,13 +5,18 @@
 from __future__ import annotations
 
 import warnings
+from typing import TYPE_CHECKING
 
+import narwhals as nw
```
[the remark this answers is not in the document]
limlam96: this is useful, thank you!

PR #333, opened 2024-10-16
imputer narwhalified v1

alphanumericmale:
Narwhalification of the BaseImputer and supporting test suite

### DIFF
--- a/tubular/imputers.py
+++ b/tubular/imputers.py
@@ -3,12 +3,17 @@
 from __future__ import annotations
 
 import warnings
+from typing import TYPE_CHECKING
 
+import narwhals as nw
 import numpy as np
 import pandas as pd
 
 from tubular.base import BaseTransformer
 from tubular.mixins import WeightColumnMixin
+
+if TYPE_CHECKING:
+    from narwhals.typing import FrameT
 
 
 class BaseImputer(BaseTransformer):
@@ -29,17 +34,18 @@ class BaseImputer(BaseTransformer):
 
     FITS = False
 
-    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+    @nw.narwhalify
+    def transform(self, X: FrameT) -> FrameT:
         """Impute missing values with median values calculated from fit method.
 
         Parameters
         ----------
-        X : pd.DataFrame
+        X : FrameT
             Data to impute.
 
         Returns
         -------
-        X : pd.DataFrame
+        X : FrameT
             Transformed input X with nulls imputed with the median value for the specified columns.
 
         """
@@ -48,7 +54,7 @@ class BaseImputer(BaseTransformer):
         X = super().transform(X)
 
         for c in self.columns:
-            X[c] = X[c].fillna(self.impute_values_[c])
+            X = X.with_column(X[c].fill_null(self.impute_values_[c]))
 
         return X
 


tubular/imputers.py
```diff
@@ -48,7 +54,7 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
         X = super().transform(X)
 
         for c in self.columns:
-            X[c] = X[c].fillna(self.impute_values_[c])
+            X = X.with_column(X[c].fill_null(self.impute_values_[c]))
```
limlam96, line 57: S: might need to add .alias(c) on the end? Not sure, maybe it does work without though.. 
`X = X.with_columns(X[c].fill_null(self.impute_values_[c]).alias(c))`

tests/imputers/test_BaseImputer.py
```diff
@@ -2,202 +2,268 @@
...
+    library = getattr(request, "param", "pandas")
+    df = pd.DataFrame(
+        {
+            "a": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0],
+            "b": ["a", "b", "c", "d", "e", "f", np.nan],
+            "c": ["a", "b", "c", "d", "e", "f", np.nan],
+        },
+    )
+    if library == "polars":
+        return pl.from_pandas(df)
```
davidhopkinson26, line 34: Q: if this is our approach to producing expected polars df (i.e converting form pandas) then maybe we can wrap the conversion in comparison logic? Had a bit of a discussion on prods/cons of this approach here https://github.com/lvgig/tubular/pull/328#issuecomment-2414138420 

Tentative conclusion there was we should probably explicitly define each data frame.

Have since then considered trying a df comparison function that internally uses nw.from_native on both dfs and compares using pl.assert_frame_equal?  I really like the idea of getting to a place where we only need define one set of expected values, but I'm honestly very torn between the safety of explicitly defining both pandas and pl dfs and something like this to keep bloat down.  Keen to have others thoughts on this.  @limlam96 @adamsardar?

limlam96: just off a call with David, I think what we've settled on is where we have data fixtures/functions like: 
```
@pytest.fixture()
def test_fit_df(request):
    library = getattr(request, "param", "pandas")
    df = pd.DataFrame(...)
    if library == "polars":
        return pl.from_pandas(df)
    return df
```

We change to like:
```
@pytest.fixture()
def test_fit_df(request):
    library = getattr(request, "param", "pandas")
    dataframe_init=tests.utils.get_dataframe_init(library)
    df = dataframe_init(...)
    return df
```
and get_dataframe_init is a new function that returns pl.DataFrame or pd.DataFrame depending on the input library argument

hope that makes sense @alphanumericmale ? apologies for the change in approach

tests/imputers/test_BaseImputer.py
```diff
@@ -2,202 +2,268 @@
...
 
-class GenericImputerTransformTestsWeight:
-    def expected_df_weights():
-        """Expected output for test_nulls_imputed_correctly_weights."""
-        df = d.create_df_9()
 
-        for col in ["b"]:
-            df.loc[df[col].isna(), col] = 4
+@pytest.fixture
+def expected_df_weights():
```
limlam96, line 206: need to add the library logic to this one

tests/imputers/test_BaseImputer.py
```diff
@@ -2,202 +2,268 @@
...
+from tests import utils as u
 from tests.base_tests import (
     ColumnStrListInitTests,
     GenericFitTests,
     OtherBaseBehaviourTests,
 )
 
 
+@pytest.fixture
+def library(request):
```
limlam96, line 19: don't think this is used

tests/imputers/test_BaseImputer.py
```diff
@@ -2,202 +2,268 @@
...
+    return request.param
+
+
+@pytest.fixture()
```
limlam96, line 23: Ah we're getting some errors with e.g. MeanImputer tests because the tests can't see the fixtures - since we already have some data creation functions in this script, I would just convert these fixtures into functions for now which should get round this

tests/imputers/test_BaseImputer.py
```diff
@@ -2,202 +2,268 @@
...
-        )
+        # Initialize the transformer
+        transformer = initialized_transformers[self.transformer_name]
+        transformer.impute_values_ = {"a": 7}
+        transformer.columns = ["a"]
+
+        # Transform the DataFrame
+        df_transformed = transformer.transform(create_df_2)
+
+        # Check outcomes for single rows
```
limlam96, line 125: S: on tests like this where you look at single rows, could you also do and assert_frame_equal on the fulls dfs at the end? just to cover everything

limlam96:
remember changelog :)

pyproject.toml
```diff
@@ -18,7 +18,7 @@ dynamic = ["version"]
 dependencies = [
     "pandas>=1.5.0",
     "scikit-learn>=1.2.0",
-    "narwhals >= 1.9.2",
+    "narwhals >= 1.9.4",
```
adamsardar, line 21: N: In future, could you include the rationale for version bumps in the git commit message? 'env update pyarrow' Doesn't help me understand why the point version had to be increase.

adamsardar: Do we need this version and up only? It's [very recent](https://pypi.org/project/narwhals/#history).

I don't think that we should change this in our PR here, but a separate piece of work is to define a strategy around API versions. narwhals has a [Perfect backwards compatibility policy,](https://narwhals-dev.github.io/narwhals/backcompat/) which feels like something to make use of.

limlam96: fair point - re narwhals, we do actually need this one upwards as it's where they actioned this bug for us https://github.com/narwhals-dev/narwhals/issues/1200

tubular/imputers.py
```diff
@@ -25,30 +30,31 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
+        X : FrameT
             Transformed input X with nulls imputed with the median value for the specified columns.
 
         """
         self.check_is_fitted(["impute_values_"])
 
-        X = super().transform(X)
+        X = nw.from_native(super().transform(X))
 
         for c in self.columns:
```
adamsardar, line 56: B: this isn't using the narwhal's API particularly as you index the frame in the same way as pandas. As they say in [the polars docs](https://docs.pola.rs/user-guide/migration/pandas/#key-syntax-differences):

> If your Polars code looks like it could be pandas code, it might run, but it likely runs slower than it should.

Use the `nw.col(name)` syntax instead:

```python
import pandas as pd
import polars as pl
import narwhals as nw

df_pd = pd.DataFrame({"a": [2, 4, None, 3, 5], "b": [2.0, 4.0, float("nan"), 3.0, 5.0]})
df_pl = pl.DataFrame({"a": [2, 4, None, 3, 5], "b": [2.0, 4.0, float("nan"), 3.0, 5.0]})
impute_values = {'a' : 1, 'b' : 2}

>>> cols = ['a', 'b']

@nw.narwhalify
def func(df):
    return df.with_columns(nw.col(c).fill_null(impute_values[c]) for c in cols)

>>> func(df_pl)
shape: (5, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ f64 │
╞═════╪═════╡
│ 2   ┆ 2.0 │
│ 4   ┆ 4.0 │
│ 1   ┆ NaN │
│ 3   ┆ 3.0 │
│ 5   ┆ 5.0 │
└─────┴─────┘


>>> func(df_pd)
     a    b
0  2.0  2.0
1  4.0  4.0
2  1.0  2.0
3  3.0  3.0
4  5.0  5.0

```

adamsardar: (apologies, this should be a few lines down)

limlam96: actioned here https://github.com/lvgig/tubular/pull/333/commits/b995e2785a70d7082eef1367dcd745dc5ea1700d

tests/imputers/test_BaseImputer.py
adamsardar:
```
library = getattr(request, "param", "pandas") 
dataframe_init=tests.utils.get_dataframe_init(library)
```

This feels like deep magick on the Python side - worth wrapping in some sugar methods so that it's a little more obvious for new contributors. Happy to kick into another ticket as this is scope creep.

limlam96: https://github.com/lvgig/tubular/issues/338

tubular/imputers.py
adamsardar: Sorry to be a stickler on this one, but this:

```python
for c in self.columns:
            X = X.with_columns(nw.col(c).fill_null(self.impute_values_[c]).alias(c))
```

Isn't the same as:

```python

X = X.with_columns(nw.col(c).fill_null(self.impute_values_[c]) for c in self.columns)
```

a.) You don't need the `alias(c)` bit - you only need that if you want [to rename something](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.alias.html).
b.) repeated `with_columns` calls have an overhead as polars can't parallelise operations on an eager frame (on a lazy frame it matters less as the query plan isn't constructed until the end).

If you feel that the above line is too long, you can construct the exprs separately:

```python

fill_null_from_fitted_exprs = [
    nw.col(c).fill_null(self.impute_values_[c]) for c in self.columns
]

X = X.with_columns(fill_null_from_fitted_exprs)
```

limlam96: cool, actioned here https://github.com/lvgig/tubular/pull/333/commits/ac84dcc38a2d8b5bf5e3d85fafeb82392c94a6a3

adamsardar:
I believe that this PR is ready other than changelog.

limlam96:
> I believe that this PR is ready other than changelog.

updated here https://github.com/lvgig/tubular/pull/333/commits/fa6e1d7e2093f72a99325982c3bc45684cea0e18

also added placeholder lines to cut down on annoying conflicts

adamsardar (approved):

PR #335, opened 2024-10-17
converted basenominaltransformer to polars

limlam96:
edited transformer and tests to use narwhals framework

https://github.com/lvgig/tubular/issues/334

### DIFF
--- a/tests/nominal/test_BaseNominalTransformer.py
+++ b/tests/nominal/test_BaseNominalTransformer.py
@@ -1,4 +1,4 @@
-import pandas as pd
+import polars as pl
 import pytest
 from sklearn.exceptions import NotFittedError
 
@@ -9,6 +9,7 @@ from tests.base_tests import (
     GenericTransformTests,
     OtherBaseBehaviourTests,
 )
+from tests.utils import assert_frame_equal_dispatch
 
 
 # The first part of this file builds out the tests for BaseNominalTransformer so that they can be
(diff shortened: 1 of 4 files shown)


davidhopkinson26 (changes requested):
think we also need to edit inverse_transform.  It has a replace operation in there.

tests/nominal/test_BaseNominalTransformer.py
```diff
@@ -20,22 +21,38 @@ class GenericNominalTransformTests(GenericTransformTests):
...
+        self,
+        initialized_transformers,
+        library,
+    ):
         """Test an exception is raised if non-mappable rows are present in X."""
-        df = d.create_df_1()
+        df = d.create_df_1(library=library)
 
-        x = initialized_transformers[self.transformer_name]
+        transformer = initialized_transformers[self.transformer_name]
```
davidhopkinson26, line 47: P: thanks for changing from x -> transformer!

limlam96:
> think we also need to edit inverse_transform. It has a replace operation in there.

inverse_transform is in 'NominalToIntegerTransformer' so was leaving for that issue? unless i'm mixed up!

davidhopkinson26:
> > think we also need to edit inverse_transform. It has a replace operation in there.
> 
> inverse_transform is in 'NominalToIntegerTransformer' so was leaving for that issue? unless i'm mixed up!

My bad sorry!

davidhopkinson26 (approved):

PR #339, opened 2024-10-29
refactored CheckNumericMixin

limlam96:
refactored for narwhals

spotted that this wasn't being used consistently across numeric file, so also addressed this

https://github.com/lvgig/tubular/issues/336


limlam96:
TODO - make new test run in polars/pandas once baseimputer PR completes

tests/mixins/test_CheckNumericMixin.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class TestCheckNumericMixin:
+    "tests for CheckNumericMixin class"
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_check_numeric_columns(self, library):
+        "test check_numeric_columns method raises appropriate error"
+
+        df = create_df_2(library=library)
```
davidhopkinson26, line 14: I think we should probably add in testing for more dtypes here.  Currently create_df_2 only has integers and strings but I think we should also cover floats, all None and np.nan at least?

Potential for pandas / polars to have different handling for which of these count as numeric so I think the narwhalification makes this more important than perhaps it was before.

davidhopkinson26 (changes requested):

tubular/mixins.py
```diff
@@ -5,31 +5,46 @@
...
+
+    Attributes
+    ----------
+
+    polars_compatible : bool
+        class attribute, indicates whether transformer has been converted to polars/pandas agnostic narwhals framework
+
+    """
+
+    polars_compatible = False
```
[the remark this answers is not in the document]
limlam96: yeah you're right - but also was unused so removed here https://github.com/lvgig/tubular/pull/339/commits/3b89bdca63915726224e887dba769a3c50c8a676

tests/mixins/test_CheckNumericMixin.py
limlam96: yeah fair point, added more cases here https://github.com/lvgig/tubular/pull/339/commits/f913f98bc365c93ca8a4951b22ddbf88ce316190

(this did spillover into a couple of other files)

davidhopkinson26 (approved):
Thanks @limlam96!

PR #341, opened 2024-10-30
added separate tests for WeightColumnMixin

limlam96:
linked to https://github.com/lvgig/tubular/issues/322

added separate tests for this mixin, and made code more polars-y

### DIFF
--- a/tests/mixins/test_WeightColumnMixin.py
+++ b/tests/mixins/test_WeightColumnMixin.py
@@ -1,0 +1,150 @@
+import narwhals as nw
+import numpy as np
+import pytest
+
+from tests.test_data import create_df_2
+from tubular.mixins import WeightColumnMixin
+
+
+class TestCheckAndSetWeight:
+    @pytest.mark.parametrize("weights_column", (0, ["a"], {"a": 10}))
+    def test_weight_arg_errors(
+        self,
+        weights_column,
+    ):
+        """Test that appropriate errors are throw for bad weight arg."""
+
+        obj = WeightColumnMixin()
+
+        with pytest.raises(
+            TypeError,
+            match="weights_column should be str or None",
+        ):
+            obj.check_and_set_weight(weights_column)
+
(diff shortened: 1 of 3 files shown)


tubular/mixins.py
```diff
@@ -138,7 +138,11 @@ class WeightColumnMixin:
         class attribute, indicates whether transformer has been converted to polars/pandas agnostic narwhals framework
     """
 
-    polars_compatible = False
+    polars_compatible = True
+
+    def classname(self) -> str:
```
davidhopkinson26, line 143: Q: I'm probably missing something but why do we need this here? any class which uses this Mixin will inherit this method from BaseTransformer?

tests/mixins/test_WeightColumnMixin.py
```diff
@@ -0,0 +1,150 @@
+import narwhals as nw
```
davidhopkinson26, line 1: P: thanks for adding this test module!

davidhopkinson26 (changes requested):

tubular/mixins.py
limlam96: ah yeah you're right it wasn't doing anything, actioned here https://github.com/lvgig/tubular/pull/341/commits/65af6b10e16b820f477d114432c19d3b9b35f206

davidhopkinson26: Sorry, looks like it is still there? did you push all changes?

limlam96: sorry not sure what happened there, must have been working on a random branch, removed now https://github.com/lvgig/tubular/pull/341/commits/0c6996372f1852bfb310102a9348769bcd63743e

davidhopkinson26 (dismissed):
thanks Liam!

limlam96:
FYI had a conflict so just pushed again :)

davidhopkinson26 (approved):

PR #340, opened 2024-10-29
refactored missed basetransformer test

limlam96:
missed one test as part of this one https://github.com/lvgig/tubular/issues/313

once BaseImputer PR completes, will also refactor test dfs in same format

Update - now BaseImputer PR is complete, went ahead and updated all test_dfs to pave way for future work

PR #342, opened 2024-11-05
Narwhalify nearest mean response imputer

Chip2916:
As per issue #320 

Some overlap with Liam's outstanding PR (#340) to move all test datasets to new format, should be no merge conflict.

### DIFF
--- a/tubular/imputers.py
+++ b/tubular/imputers.py
@@ -475,7 +475,7 @@ class NearestMeanResponseImputer(BaseImputer):
         response_column = "_temporary_response"
 
         for c in self.columns:
-            c_nulls = X[c].is_null()
+            c_nulls = X.select(nw.col(c).is_null())[c]
 
             if c_nulls.sum() == 0:
                 msg = f"{self.classname()}: Column {c} has no missing values, cannot use this transformer."
@@ -488,16 +488,16 @@ class NearestMeanResponseImputer(BaseImputer):
             mean_response_nulls = X_y.filter(c_nulls)[response_column].mean()
 
             mean_response_by_levels = mean_response_by_levels.with_columns(
-                abs_diff_response=np.abs(
-                    mean_response_by_levels[response_column] - mean_response_nulls,
-                ),
+                (nw.col(response_column) - mean_response_nulls)
+                .abs()
+                .alias("abs_diff_response"),
             )
 
             # take first value having the minimum difference in terms of average response
             self.impute_values_[c] = mean_response_by_levels.filter(
                 mean_response_by_levels["abs_diff_response"]
                 == mean_response_by_levels["abs_diff_response"].min(),
-            )[c].to_numpy()[0]
+            )[c].item(index=0)
 
         return self
 
--- a/tubular/imputers.py
+++ b/tubular/imputers.py
@@ -13,6 +13,7 @@ from tubular.base import BaseTransformer
 from tubular.mixins import WeightColumnMixin
 
 if TYPE_CHECKING:
+    import pandas as pd
     from narwhals.typing import FrameT
 
 
@@ -433,7 +434,7 @@ class NearestMeanResponseImputer(BaseImputer):
 
     """
 
-    polars_compatible = False
+    polars_compatible = True
 
     FITS = True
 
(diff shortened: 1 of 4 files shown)


tubular/imputers.py
```diff
@@ -461,40 +463,41 @@ def fit(self, X: pd.DataFrame, y: pd.Series) -> pd.DataFrame:
...
 
         self.impute_values_ = {}
 
-        X_y = self._combine_X_y(X, y)
+        X_y = nw.from_native(self._combine_X_y(X, y))
         response_column = "_temporary_response"
 
         for c in self.columns:
-            c_nulls = X[c].isna()
+            c_nulls = X[c].is_null()
```
limlam96, line 478: S: following adam's lead to push for polars-y syntax, think this may be faster with:
`c_nulls=X.select(nw.col(c).is_null())`

tubular/imputers.py
```diff
@@ -461,40 +463,41 @@ def fit(self, X: pd.DataFrame, y: pd.Series) -> pd.DataFrame:
...
+                X_y.filter(~c_nulls).group_by(c).agg(nw.col(response_column).mean())
+            )
 
-            mean_response_nulls = X_y.loc[c_nulls, response_column].mean()
+            mean_response_nulls = X_y.filter(c_nulls)[response_column].mean()
 
-            mean_response_by_levels["abs_diff_response"] = np.abs(
-                mean_response_by_levels[response_column] - mean_response_nulls,
+            mean_response_by_levels = mean_response_by_levels.with_columns(
+                abs_diff_response=np.abs(
```
limlam96, line 491: S: would this work like:
```
mean_response_by_levels=mean_response_by_levels.with_columns(
(nw.col(response_column)-mean_response_nulls).abs().alias('abs_diff_response')
)
```

tubular/imputers.py
```diff
@@ -461,40 +463,41 @@ def fit(self, X: pd.DataFrame, y: pd.Series) -> pd.DataFrame:
...
             )
 
             # take first value having the minimum difference in terms of average response
-            self.impute_values_[c] = mean_response_by_levels.loc[
+            self.impute_values_[c] = mean_response_by_levels.filter(
                 mean_response_by_levels["abs_diff_response"]
                 == mean_response_by_levels["abs_diff_response"].min(),
-                c,
-            ].to_numpy()[0]
+            )[c].to_numpy()[0]
```
limlam96, line 500: S: might be able to replace `.to_numpy()[0]` with `.item(index=0)`

limlam96 (commented):
just a few comments around trying to use less pandas-y syntax :)

tubular/imputers.py
Chip2916: Ahh to be fair that doesn't actually work as it returns a df

Chip2916: So this works, but for some reason adding .to_series() (Polars method) on the end doesn't work?

limlam96: interesting, what's the error like when you add .to_series?

Chip2916: I've added this (kinda) - this returns a df when we need a series. I've tried chaining the .to_series() method but for some reason that throws an error? I've trialled the exact thing in a notebook without issue. Not sure if this is polars-y enough

Chip2916: [here](https://github.com/lvgig/tubular/pull/342/commits/01de73607f1ef09031fb121b17b6b4bdc36c9212)

tubular/imputers.py
Chip2916: [here](https://github.com/lvgig/tubular/pull/342/commits/01de73607f1ef09031fb121b17b6b4bdc36c9212)

tubular/imputers.py
Chip2916: [here](https://github.com/lvgig/tubular/pull/342/commits/01de73607f1ef09031fb121b17b6b4bdc36c9212)

tubular/imputers.py
Chip2916: Sorry my comments are all a bit delayed as I added them then got sidetracked before committing! It basically says that the dataframe doesn't have a 'to_series()' method

limlam96: ah OK weird - as there isn't a 'LazySeries' in polars the optimisations might be limited anyway, thanks for looking i'll approve!

limlam96 (approved):

PR #345, opened 2024-11-07
Feature/narwhalify modeimp

limlam96:
narwhalified mode imputer

also refactored to tighten up code. This involved new logic to add in dummy weights when they are not provided, adding a dummy column is not ideal but does have benefits in not having two paths through code - open to feedback!

https://github.com/lvgig/tubular/issues/321


limlam96:
Note, accidentally branched off of checknumericmixin branch, so would leave this til that completes

tests/imputers/test_ModeImputer.py
```diff
@@ -34,70 +41,175 @@ class TestFit(WeightColumnFitMixinTests, GenericFitTests):
...
 
     @pytest.mark.parametrize(
-        ("df", "expected"),
-        ta.pandas.row_by_row_params(
-            pd.DataFrame({"a": [np.nan, np.nan, np.nan], "b": [None, None, None]}),
-            expected_df_nan(),
+        "library",
+        ["pandas", "polars"],
+    )
+    def test_learnt_values_tied_weighted(self, library):
```
davidhopkinson26, line 94: S: should we maybe consider testing this behaviour with types other than int? e.g str?  Sort of thing where behaviour could maybe change with changes in dependencies and cause a real head scratcher.

davidhopkinson26:
fyi - pull changes from main to fix the readthedocs build failure

tests/imputers/test_ModeImputer.py
limlam96: sorry this is probably a distant memory now, but finally got back to this https://github.com/lvgig/tubular/pull/345/commits/37771b14f87e9b3654461b8294e5db4327f31344

adding more type tests did expose some issues, so was worth doing, you'll see I had to make some improvements to dataframe_init_dispatch (and also added tests)

tests/base_tests.py
```diff
@@ -585,7 +585,7 @@ def test_fit_not_changing_data_weighted(
     @pytest.mark.parametrize(
         "bad_weight_value, expected_message",
         [
-            (np.nan, "weight column must be non-null"),
```
davidhopkinson26, line 588: Same issue as in #349 that removing np.nan for all cases is perhaps not the best idea? @Chip2916 is looking into adding handling in median imputer for np.nan which would fix this, might be able to do simillar here and have both None and np.nan cases in this base test. https://github.com/lvgig/tubular/pull/349#discussion_r1861964504

If we cant do np.nan then we should probably juts be overriding for this transformers case rather than changing for the whole package.

davidhopkinson26 (changes requested):

tests/imputers/test_ModeImputer.py
davidhopkinson26: thanks for adding this

tests/base_tests.py
limlam96: ahh yes good point, want to test both 

I've added handling here that I think will work until the narwhals method becomes available?

https://github.com/lvgig/tubular/pull/345/commits

davidhopkinson26: https://github.com/lvgig/tubular/pull/345/commits/30061ed3bbf6d05100eaddb8a3ade23bc5176b48 looks like it addresses this adequately until is_nan gets implemented in narwhals.

PR #351, opened 2024-12-11
narwhalified MeanImputer

limlam96:
Converted transformer to narwhals https://github.com/lvgig/tubular/issues/344 #344

### DIFF
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -585,7 +585,7 @@ class WeightColumnFitMixinTests:
     @pytest.mark.parametrize(
         "bad_weight_value, expected_message",
         [
-            (np.nan, "weight column must be non-null"),
+            (None, "weight column must be non-null"),
             (np.inf, "weight column must not contain infinite values."),
             (-np.inf, "weight column must be positive"),
             (-1, "weight column must be positive"),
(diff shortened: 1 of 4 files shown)


tests/base_tests.py
```diff
@@ -585,7 +585,7 @@ def test_fit_not_changing_data_weighted(
     @pytest.mark.parametrize(
         "bad_weight_value, expected_message",
         [
-            (np.nan, "weight column must be non-null"),
```
davidhopkinson26, line 588: As in https://github.com/lvgig/tubular/pull/345#discussion_r1880067836

davidhopkinson26 (changes requested):
in general looks good, but with the caveat of the issue with np.nan in the fit_not_changing_data_weighted test as we have seen in Mode and Median imputers.

tests/base_tests.py
limlam96: pulled in the fix here :) https://github.com/lvgig/tubular/pull/351/commits/aed1e2c438dec42be95a2c1401c0648222165a2d

davidhopkinson26 (dismissed):
Thanks @limlam96!

limlam96:
> Sorry, just realised need to update CHANGELOG

ah yeah thanks, updated here https://github.com/lvgig/tubular/pull/351/commits/59d74fb1cb12b3c2c7b8dd3874fc4f88bbe70c29

davidhopkinson26 (approved):

PR #353, opened 2024-12-11
updated weight mixin to use new narwhals is_finite method

limlam96:
narwhals has added 'is_finite' method that means we can remove eager-frame logic from this mixin


requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.10
+# This file is autogenerated by pip-compile with Python 3.9
```
limlam96, line 2: fyi for reviewer, we switched to py3.9 as primary dev env a while ago ([here](https://github.com/lvgig/tubular/pull/326), so this was picking up a leftover 3.10 ref. Seeing this led me to ctrl-shift-f for 3.10 where I found and updated another old ref

davidhopkinson26: thanks for the explainer

davidhopkinson26 (approved):
Happy with this. Thanks for following up.

davidhopkinson26:
sorry, thought the merge conflict on CHANGELOG looked trivial so I tried to resolve quickly.  turns out there was another conflict with numpy import being removed in this PR which conflicted with the np.nan check added in #351 which wasn't obvious from the conflict resolver

cjmwills (approved):
Pretty simple first tubular review for me!

PR #381, opened 2025-01-08
narwhalified BaseMappingTransformerMixin

limlam96:
To do this had to make mapping transformers more type conscious, with introduction of new `return_dtypes` dict arg. Also updated other impacted transformers to work with this new setup

https://github.com/lvgig/tubular/issues/369

### DIFF
--- a/tubular/mapping.py
+++ b/tubular/mapping.py
@@ -143,22 +143,24 @@ class BaseMappingTransformMixin(BaseTransformer):
             Transformed input X with levels mapped accoriding to mappings dict.
 
         """
-        self.check_is_fitted(["mappings"])
-
-        if not self.return_dtypes:
-            msg = f"{self.classname()}.transform requires return_dtypes attr to be specified"
-            raise ValueError(
-                msg,
-            )
+        self.check_is_fitted(["mappings", "return_dtypes"])
 
         X = nw.from_native(super().transform(X))
         native_namespace = nw.get_native_namespace(X)
+
+        # will do a join further down, which does not preserve index
+        # polars does not care about this, but pandas does,
+        # so need to handle a bit carefully
+        if nw.get_native_namespace(X).__name__ == "pandas":
+            index = nw.to_native(X).index
+
+        # pull out column order to preserve
+        column_order = X.columns
 
         for col in self.mappings:
             mappings = self.mappings[col]
 
             # TODO - update this logic once narwhals implements map_dict
-
             # differentiate between unmapped cols and cols mapped to null
             # by including unmapped cols
             unique = X.get_column(col).unique()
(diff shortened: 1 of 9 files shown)


tubular/mapping.py
```diff
@@ -37,14 +40,35 @@ class BaseMappingTransformer(BaseTransformer):
...
+        "Int32",
+        "Int64",
+        "Float32",
+        "Float64",
+    ]
+
+    def __init__(
+        self,
+        mappings: dict[str, dict],
+        return_dtypes: dict[str, RETURN_DTYPES] | None = None,
```
davidhopkinson26, line 69: Looks like we are currently treating this as mandatory (i.e we require it to be a dict so None would cause an error)

Wondering if we should set up a default value based on mappings in init?

tubular/nominal.py
```diff
@@ -158,6 +159,13 @@ def fit(self, X: pd.DataFrame, y: pd.Series | None = None) -> pd.DataFrame:
                 k: i for i, k in enumerate(col_values, self.start_encoding)
             }
 
+            # if more levels than int8 type can handle, then error
+            if len(self.mappings[c]) > 127:
```
davidhopkinson26, line 163: should we also raise a warning if the user tries to set return_dtypes since it will be overwritten?

davidhopkinson26 (changes requested):

tubular/mapping.py
liamholmes31: good suggestion, actioned here https://github.com/azukds/tubular/pull/381/commits/1fc778d91eb9c6b229d6d326348fd927334e8772

liamholmes31:
Bit of scope creep, but I realised the I hadn't added type checks for return_dtypes.. but have recently started using a tool called beartype elsewhere that removes the need for these, so felt much easier to take the plunge and include here than to write additional checks.

https://github.com/beartype/beartype

See what you think
https://github.com/azukds/tubular/pull/381/commits/1fc778d91eb9c6b229d6d326348fd927334e8772

tubular/nominal.py
liamholmes31: so started work on this, but found that these two transformers actually don't accept return_dtypes as an input, because BaseNominalTransformer rejects it - so should be all good!

pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
     "scikit-learn>=1.2.0",
     "narwhals >= 1.17.0",
     "polars >= 1.9.0",
+    "beartype",
```
davidhopkinson26, line 23: Q: sure this can be completely unpinned?

tests/base_tests.py
```diff
@@ -453,7 +453,7 @@ def test_unexpected_kwarg_error(
             ),
         ):
             uninitialized_transformers[self.transformer_name](
-                unexpected_kwarg="spanish inquisition",
+                unexpected_kwarg=True,
```
davidhopkinson26, line 456: ...but it was such a good pun

davidhopkinson26 (approved):
Changes look good

pyproject.toml
liamholmes31: added a >= pin to be safe https://github.com/azukds/tubular/pull/381/commits/042d629e167dd0b3cf0896455564b91feac35164

davidhopkinson26 (approved):

PR #382, opened 2025-01-10
Refactored DateTimeTransformer to use narwhals

favouribude1:
Currently havin one test failing. 
image

### DIFF
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -5,8 +5,10 @@ from __future__ import annotations
 import datetime
 import warnings
 
+import narwhals as nw
 import numpy as np
 import pandas as pd
+import polars as pl
 
 from tubular.base import BaseTransformer
 from tubular.mixins import DropOriginalMixin, NewColumnNameMixin, TwoColumnMixin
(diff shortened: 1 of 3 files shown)


tubular/dates.py
```diff
@@ -540,23 +542,56 @@ def __init__(
...
         # are not yet date/datetime
-        X = BaseTransformer.transform(self, X)
+        X = nw.from_native(BaseTransformer.transform(self, X))
 
-        X[self.new_column_name] = pd.to_datetime(
-            X[self.columns[0]],
-            **self.to_datetime_kwargs,
-        )
+        native_X = X.to_native()
+        if isinstance(native_X, pd.DataFrame):
```
limlam96, line 561: narwhals is designed to avoid having to implement branching polars/pandas handling like this, the idea is to make use of narwhals methods, which will translate between pandas/polars behind the scenes. In this case, would try something like
```
time_unit=self.to_datetime_kwargs.get('time_unit', None)
time_zone=self.to_datetime_kwargs.get('time_zone', None)
df=df.with_columns(nw.col(columns[0]).cast(nw.Datetime(time_unit=time_unit, time_zone=time_zone)))
```

PR #383, opened 2025-01-10
Feature/nw group rare levels transformer

limlam96:
As part of this, had to make transformer more opinionated and not accept null columns - it will now error and direct users to use an imputer first in this case

https://github.com/lvgig/tubular/issues/372

### DIFF
--- a/tubular/nominal.py
+++ b/tubular/nominal.py
@@ -1,10 +1,12 @@
 """This module contains transformers that apply encodings to nominal columns."""
 from __future__ import annotations
 
+import copy
 import warnings
 from typing import TYPE_CHECKING, Literal
 
 import narwhals as nw
+import narwhals.selectors as ncs
 import numpy as np
 import pandas as pd
 
@@ -301,7 +303,7 @@ class GroupRareLevelsTransformer(BaseTransformer, WeightColumnMixin):
 
     """
 
-    polars_compatible = False
+    polars_compatible = True
 
     FITS = True
 
@@ -329,6 +331,10 @@ class GroupRareLevelsTransformer(BaseTransformer, WeightColumnMixin):
 
         WeightColumnMixin.check_and_set_weight(self, weights_column)
 
+        if not isinstance(rare_level_name, str):
+            msg = f"{self.classname()}: rare_level_name must be a str"
+            raise ValueError(msg)
+
         self.rare_level_name = rare_level_name
 
         if not isinstance(record_rare_levels, bool):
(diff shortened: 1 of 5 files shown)


tubular/nominal.py
```diff
@@ -343,7 +349,71 @@ def __init__(
 
         self.unseen_levels_to_rare = unseen_levels_to_rare
 
-    def fit(self, X: pd.DataFrame, y: pd.Series | None = None) -> pd.DataFrame:
+    @nw.narwhalify
+    def _check_strlike_columns(self, X: FrameT) -> None:
```
davidhopkinson26, line 353: N: would put an underscore between str_like personally.  Had to double take when I first read it as one word.

tubular/nominal.py
```diff
@@ -343,7 +349,71 @@ def __init__(
...
+    def _check_for_nulls(self, X: FrameT) -> None:
+        """check that transformer being called on only non-null columns.
+
+        Note, found including nulls to be quite complicated due to:
+        - categorical variables make use of NaN not None
+        - pl/nw categorical variables do not allow categories to be edited,
+        so adjusting requires converting to str as interim step
+        - NaNs are converted to nan, introducing complications
+
+        As this transformer is generally used post imputation, elected to remove null
```
davidhopkinson26, line 395: P: this seems reasonable

tubular/nominal.py
```diff
@@ -366,141 +436,131 @@ def fit(self, X: pd.DataFrame, y: pd.Series | None = None) -> pd.DataFrame:
...
-                        X[c].isin(self.non_rare_levels[c]),
-                        X[c],
-                        self.rare_level_name,
-                    ),
-                    index=X.index,
-                )
+            categorical = False
+            if str(X.schema[c]) == "Categorical":
+                categorical = True
+                X = X.with_columns(nw.col(c).cast(nw.String))
```
davidhopkinson26, line 545: Q: Does this mean the transformer will always cast to string?

liamholmes31: so it always casts to category type to str as an interim step, but then casts back to categorical before returning (line 563)

tubular/nominal.py
```diff
@@ -343,7 +349,71 @@ def __init__(
...
+        X : pd/pl.DataFrame
+            Data to transform
+
+        """
+
+        strlike_columns = list(
+            set(self.columns).intersection(
+                set(
+                    X.select(
+                        ncs.string(),
```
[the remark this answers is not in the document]
liamholmes31: hm yeah from what I can see, isn't yet implemented https://narwhals-dev.github.io/narwhals/api-reference/selectors/#narwhals.selectors.string

davidhopkinson26 (approved):

PR #387, opened 2025-01-30
converted BaseDatetimeTransformer to narwhals

cjmwills:
Made BaseDatetimeTransformer polars compatible and updated tests. Kept the structure of test_BaseDatetimeTransformer similar to avoid error with converting pandas data frame to date type. Explanation in code comments.

#375

### DIFF
--- a/tests/dates/test_BaseDatetimeTransformer.py
+++ b/tests/dates/test_BaseDatetimeTransformer.py
@@ -1,7 +1,9 @@
 from copy import deepcopy
 
+import narwhals as nw
 import numpy as np
 import pandas as pd
+import polars as pl
 import pytest
 
 from tests.base_tests import (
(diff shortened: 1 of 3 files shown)


tests/dates/test_BaseDatetimeTransformer.py
```diff
@@ -34,32 +41,43 @@ def test_non_datetypes_error(
...
             col = columns[i]
-            df[col] = bad_value
-            # force date test to wanted type
-            if bad_type == "Date":
-                df[col] = df[col].astype("date32[pyarrow]")
-
-            x = uninitialized_transformers[self.transformer_name](
-                **args,
-            )
+            # force date test to wanted type if pandas df, to avoid
```
adamsardar, line 61: S: Worth including a similar comment into the `dates.py` base class? It took me a while to understand why `check_columns_are_date_or_datetime()` is structured the way that it is without this comment ... which is in the tests.

adamsardar: I think that `check_columns_are_date_or_datetime` in general could do with some deep thinking, and that should be done when considering the other date transformers in the `dates.py` file. Won't hold on this item.

adamsardar (approved):
Nice start to to dates.py refactor, which will be quite the task in the fullness of time.

Happy with this PR,

PR #388, opened 2025-02-19
Optional wanted_values feature added to OHE

Boluwatife28:
An optional feature, "wanted_levels," has been integrated into the OneHotEncodingTransformer. This feature allows users to specify which levels in a column they wish to encode. 
If "wanted_levels" is set to None, the transformer will encode all levels present in the columns. 
Additionally, new tests have been added to the test suite to validate the performance of this feature.

### DIFF
--- a/tests/nominal/test_OneHotEncodingTransformer.py
+++ b/tests/nominal/test_OneHotEncodingTransformer.py
@@ -26,6 +26,19 @@ class TestInit(
     def setup_class(cls):
         cls.transformer_name = "OneHotEncodingTransformer"
 
+    @pytest.mark.parametrize(
+            "values", 
+            [{'b': ['a', 'b']}, 'a', ['a', 'b'], 123, True],
+    )
+    def test_wanted_values_is_dict(self, values):
+        df= d.create_df_1()
+ 
+        with pytest.raises(
+            TypeError,
+            match= 'OneHotEncodingTransformer: Wanted_values should be a dictionary',
+        ):
+            self.OneHotEncodingTransformer(wanted_values = values)
+
 
 class TestFit(GenericFitTests):
     """Generic tests for transformer.fit()"""
--- a/tests/nominal/test_OneHotEncodingTransformer.py
+++ b/tests/nominal/test_OneHotEncodingTransformer.py
@@ -26,51 +26,48 @@ class TestInit(
     def setup_class(cls):
         cls.transformer_name = "OneHotEncodingTransformer"
 
-   
     # Tests for wanted_values parameter
 
     @pytest.mark.parametrize(
-            "values", 
-            [ "a", ["a", "b"], 123, True],
+        "values",
+        ["a", ["a", "b"], 123, True],
     )
     def test_wanted_values_is_dict(self, values, minimal_attribute_dict):
         args = minimal_attribute_dict[self.transformer_name]
-        args["wanted_values"]=values
- 
+        args["wanted_values"] = values
+
         with pytest.raises(
             TypeError,
-            match= "OneHotEncodingTransformer: Wanted_values should be a dictionary",
+            match="OneHotEncodingTransformer: Wanted_values should be a dictionary",
         ):
             OneHotEncodingTransformer(**args)
 
-
     @pytest.mark.parametrize(
-            "values",
-            [
-                {1:["a", "b"]},
-                {True:["a"]},
-                {("a",):["b", "c"]},
-            ]
+        "values",
+        [
+            {1: ["a", "b"]},
+            {True: ["a"]},
+            {("a",): ["b", "c"]},
+        ],
     )
     def test_wanted_values_key_is_str(self, values, minimal_attribute_dict):
         args = minimal_attribute_dict[self.transformer_name]
-        args["wanted_values"]= values
- 
+        args["wanted_values"] = values
+
         with pytest.raises(
             TypeError,
-            match= "OneHotEncodingTransformer:  Key in 'wanted_values' should be a string",
+            match="OneHotEncodingTransformer:  Key in 'wanted_values' should be a string",
         ):
             OneHotEncodingTransformer(**args)
-
 
     @pytest.mark.parametrize(
         "values",
         [
             {"a": "b"},
-            {"a":("a","b")},
+            {"a": ("a", "b")},
             {"a": True},
             {"a": 123},
-        ]
+        ],
     )
     def test_wanted_values_value_is_list(self, values, minimal_attribute_dict):
         args = minimal_attribute_dict[self.transformer_name]
@@ -82,7 +79,6 @@ class TestInit(
         ):
             OneHotEncodingTransformer(**args)
 
-
     @pytest.mark.parametrize(
         "values",
         [
@@ -90,18 +86,17 @@ class TestInit(
             {"a": ["b", True]},
             {"a": ["b", None]},
             {"a": ["b", ["a", "b"]]},
-        ]
+        ],
     )
     def test_wanted_values_entries_are_str(self, values, minimal_attribute_dict):
-        args= minimal_attribute_dict[self.transformer_name]
-        args["wanted_values"]= values
+        args = minimal_attribute_dict[self.transformer_name]
+        args["wanted_values"] = values
 
         with pytest.raises(
             TypeError,
-            match= "OneHotEncodingTransformer: Entries in 'wanted_values' list should be a string"
+            match="OneHotEncodingTransformer: Entries in 'wanted_values' list should be a string",
         ):
             OneHotEncodingTransformer(**args)
-            
 
 
 class TestFit(GenericFitTests):
@@ -126,6 +121,23 @@ class TestFit(GenericFitTests):
             match="OneHotEncodingTransformer: column b has nulls - replace before proceeding",
         ):
             transformer.fit(df)
+
+    @pytest.mark.parametrize(
+            "library",
+            ["pandas", "polars"]
+    )
+    def test_fit_missing_levels_warning(self, library):
+        """ Test OneHotEncodingTransformer.fit triggers a warning for missing levels."""
+        df = d.create_df_1(library=library)
+
+        transformer = OneHotEncodingTransformer(columns=["b"], wanted_values={"b": ["f", "g"]})
+
+        with pytest.warns(
+            UserWarning,
+            match= ("OneHotEncodingTransformer: column b includes user-specified values .* not found in the dataset"),
+        ):
+            transformer.fit(df)
+
 
     @pytest.mark.parametrize(
         "library",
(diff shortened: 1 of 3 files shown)


tubular/nominal.py
```diff
@@ -1184,9 +1189,23 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
             # for consistency
             levels_list.sort()
 
-            self.categories_[c] = levels_list
+            # filter if 'values' is provided
```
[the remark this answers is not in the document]
liamholmes31: Stretch task: for bullet point 2, would good to raise a warning when there are columns in levels which are not in the data

```
import warnings
warnings.warn('blablabla')
```

tubular/nominal.py
```diff
@@ -1260,6 +1288,22 @@ def transform(self, X: FrameT) -> FrameT:
                 for missing_level in missing_levels[c]
             )
 
+            # find user-defined values not present in dataset
```
liamholmes31, line 1291: S: I think this chunk of new code is already covered by lines 1273-1289 above :) looks like we actually don't really need to change transform luckily!

tubular/nominal.py
```diff
@@ -1184,12 +1191,33 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
...
+                final_categories = levels_list
+            else:
+                final_categories = selected_values
 
+            self.categories_[c] = final_categories
             self.new_feature_names_[c] = self._get_feature_names(column=c)
 
         return self
 
+    def warn_missing_levels(self, X: FrameT, c: str, missing_levels: dict[str, list[str]])-> None:
```
liamholmes31, line 1207: S: think the next challenge is to set this function up so that it can be used in both fit and transform :) looks like it will need to return the missing_levels dictionary, as this gets used later in transform

tubular/nominal.py
Boluwatife28: Yes I noticed that too.
Quick question - Do you think we since lines 1270-1281 are now in the new warn_missing_levels function we don't necessarily need them verbatim in the transform method again. We could just make use of the function. What is your opinion on that?

tubular/nominal.py
Boluwatife28: Okay, I will work on that.

tubular/nominal.py
```diff
@@ -1238,19 +1290,11 @@ def transform(self, X: FrameT) -> FrameT:
                     f"{self.classname()}: column %s has nulls - replace before proceeding"
                     % c,
                 )
-
             # print warning for unseen levels
-            present_levels = set(X.select(nw.col(c).unique()).get_column(c).to_list())
```
liamholmes31, line 1243: S: I would drop the present_levels argument and reintroduce this line, and then can call `warn_missing_levels` below :)

tubular/nominal.py
```diff
@@ -1184,12 +1211,37 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
...
 
+            self.categories_[c] = final_categories
             self.new_feature_names_[c] = self._get_feature_names(column=c)
+        missing_levels = {}
+        present_levels = set(X.select(nw.col(c).unique()).get_column(c).to_list())
+        missing_levels = self.warn_missing_levels(present_levels, c, missing_levels)
 
         return self
 
+    def warn_missing_levels(
```
liamholmes31, line 1232: N: there's a convention to start functio names with "_" when we don't want users to directly call them (the term for this is 'private method/function', I think that would fit here :)

tubular/nominal.py
```diff
@@ -1087,6 +1087,9 @@ class OneHotEncodingTransformer(
         Names of columns to transform. If the default of None is supplied all object and category
         columns in X are used.
 
+    values: dict[str, list[str] or None , default = None
```
liamholmes31, line 1090: N: typo, 'wanted_values'

liamholmes31: Maybe add a final sentence to the description: 'format {col1: [level_1, level_2, ...], ...}'

tubular/nominal.py
```diff
@@ -1184,12 +1211,40 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
...
+            )
+
+            if selected_values is None:
+                final_categories = levels_list
+            else:
+                final_categories = selected_values
 
+            self.categories_[c] = final_categories
             self.new_feature_names_[c] = self._get_feature_names(column=c)
+        missing_levels = {}
```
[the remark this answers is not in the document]
liamholmes31: this is so that the missing level check happens for each column

tubular/nominal.py
```diff
@@ -1184,12 +1211,40 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
...
+        present_levels: list,
+        c: str,
+        missing_levels: dict[str, list[str]],
+    ) -> list:
+        # print warning for missing levels
+        missing_levels[c] = list(
+            set(self.categories_[c]).difference(present_levels),
+        )
+        if missing_levels:
+            warning_msg = f"{self.classname()}: column {c} includes user-specified values not found in the dataset"
```
liamholmes31, line 1243: S: can we add the missing levels to the warning message (so include `{missing_levels[c]}` in the message str)

liamholmes31:
to get this PR ready for next reviewer, could you please:

- update the title to be more descriptive
- add a short description summarising the work + linking to the relevant issue :)

tubular/nominal.py
Boluwatife28: Sure!

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -50,6 +122,23 @@ def test_nulls_in_X_error(self, library):
...
+    )
+    def test_fit_missing_levels_warning(self, library):
+        """ Test OneHotEncodingTransformer.fit triggers a warning for missing levels."""
+        df = d.create_df_1(library=library)
+
+        transformer = OneHotEncodingTransformer(columns=["b"], wanted_values={"b": ["f", "g"]})
+
+        with pytest.warns(
+            UserWarning,
+            match= ("OneHotEncodingTransformer: column b includes user-specified values .* not found in the dataset"),
```
liamholmes31, line 137: S: looks good, but can we replace the .* with the actual list?

liamholmes31:
We use a Changelog file (https://github.com/azukds/tubular/blob/main/CHANGELOG.rst) to keep track of what changes between versions, could you please add a line explaining the change made here :)

tubular/nominal.py
```diff
@@ -1137,6 +1141,29 @@ def __init__(
             **kwargs,
         )
 
+        if wanted_values is not None:
+            if not isinstance(wanted_values, dict):
+                msg = f"{self.classname()}: Wanted_values should be a dictionary"
```
liamholmes31, line 1146: N: could you change to lowercase "wanted_values" to match format elsewhere

tubular/nominal.py
```diff
@@ -1184,12 +1212,44 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
...
+
+            if selected_values is None:
+                final_categories = levels_list
+            else:
+                final_categories = selected_values
 
+            self.categories_[c] = final_categories
             self.new_feature_names_[c] = self._get_feature_names(column=c)
 
+            present_levels = set(X.select(nw.col(c).unique()).get_column(c).to_list())
```
liamholmes31, line 1228: N: think this could be achieved with fewer method calls like:
```
present_levels=set(X.get_column(c).unique().to_list())
```

tubular/nominal.py
```diff
@@ -1184,12 +1212,44 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
...
+
         return self
 
+    def _warn_missing_levels(
+        self,
+        present_levels: list,
+        c: str,
+        missing_levels: dict[str, list[str]],
+    ) -> list:
+        # print warning for missing levels
```
liamholmes31, line 1243: B: need to add a docstring for this one, can follow the format of the other docstrings :)

tubular/nominal.py
```diff
@@ -1184,12 +1212,44 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
...
+            )
+
         return self
 
+    def _warn_missing_levels(
+        self,
+        present_levels: list,
+        c: str,
+        missing_levels: dict[str, list[str]],
+    ) -> list:
```
liamholmes31, line 1242: B: think the typehint is wrong here, should match 'missing_levels'

tubular/nominal.py
```diff
@@ -1184,12 +1212,44 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
...
+        # print warning for missing levels
+        missing_levels[c] = list(
+            set(self.categories_[c]).difference(present_levels),
+        )
+        if len(missing_levels) > 0:
```
liamholmes31, line 1247: B: think this should be `if len(missing_levels[c])>0:`

liamholmes31: this has highlighted that we should probably add another test for a case where 'wanted_values' is provided but no warning is raised, this may be helpful to action that :)
https://stackoverflow.com/questions/45671803/how-to-use-pytest-to-assert-no-warning-is-raised

tubular/nominal.py
```diff
@@ -1242,15 +1302,12 @@ def transform(self, X: FrameT) -> FrameT:
             # print warning for unseen levels
             present_levels = set(X.select(nw.col(c).unique()).get_column(c).to_list())
```
liamholmes31, line 1303: N: same comment about how we could do this with fewer methods:
`present_levels=set(X.get_column(c).unique().to_list())`

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -319,3 +432,43 @@ def test_unseen_categories_encoded_as_all_zeroes(self, library):
...
                 df_expected_row,
             )
+
+
+    @pytest.mark.parametrize(
+        "library",
+        ["pandas", "polars"],
+    )
+    def test_transform_missing_levels_encoded_as_all_zeroes(self, library):
+        """Test OneHotEncodingTransformer.transform triggers a warning for missing levels."""
```
liamholmes31, line 442: N: docstring needs updating here

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -319,3 +432,43 @@ def test_unseen_categories_encoded_as_all_zeroes(self, library):
                 df_transformed_row[column_order],
                 df_expected_row,
             )
+
+
+    @pytest.mark.parametrize(
+        "library",
+        ["pandas", "polars"],
+    )
+    def test_transform_missing_levels_encoded_as_all_zeroes(self, library):
```
liamholmes31, line 441: N: would maybe rename to 'test_transform_output_with_wanted_values_arg'

tubular/nominal.py
Boluwatife28: Q: for the additional test, do you mean a case where all levels in the user-specified 'wanted_levels' are present in the data hence no missing_level warning is raised?

liamholmes31: yes exactly :) as that would have failed while the issue that the first comment in this thread mentions was present, and alerted us that something was up

tubular/nominal.py
```diff
@@ -1231,12 +1259,56 @@ def fit(self, X: FrameT, y: nw.Series | None = None) -> FrameT:
...
+
+        Parameters
+        ----------
+        present_levels: list
+            List of levels observed in the data.
+        c: str
+            The column name being checked for missing user-specified levels.
+        missing_levels: dict[str, list[str]]
+            Dictionary containing missing user-specified levels for each column.
+
```
liamholmes31, line 1300: Last comment, can you just add a section on the returned values to this docstring please:
```
Returns
        -------
           missing_levels : dict[str, list[str]]
               dictionary of form col: missing_level info, with this info added/updated for column c
```

Boluwatife28: Sure, I will include that.

Boluwatife28: Done https://github.com/azukds/tubular/blob/fb5d7046e9e672f190b991901d315b076b838ac2/tubular/nominal.py#L1300-L1303

liamholmes31 (approved):
Approved! first PR under your belt, nice one Bolu :)

PR #390, opened 2025-03-06
Feature/narwhalify arb imp

alphanumericmale:
Narwhalification of the arbitrary imputer https://github.com/azukds/tubular/issues/315

### DIFF
--- a/tests/imputers/test_ArbitraryImputer.py
+++ b/tests/imputers/test_ArbitraryImputer.py
@@ -13,18 +13,16 @@ from tests.imputers.test_BaseImputer import GenericImputerTransformTests
 from tubular.imputers import ArbitraryImputer
 
 
-# Dataframe used exclusively in this testing script
-@pytest.fixture(params=["pandas", "polars"])
-def downcast_df(request):
+def impute_df_with_several_types(library="pandas"):
     """
     Fixture that returns a DataFrame with columns suitable for downcasting
     for both pandas and polars.
     """
     data = {
-        "a": [1, 2, 3, 4, 5],
-        "b": [1.0, 2.0, 3.0, 4.0, 5.0],
+        "a": ["a", "b", "c", "d", None],
+        "b": [1.0, 2.0, 3.0, 4.0, None],
+        "c": [True, False, False, None, True],
     }
-    library = request.param
 
     return u.dataframe_init_dispatch(data, library)
 
@@ -62,23 +60,99 @@ class TestTransform(GenericImputerTransformTests, GenericTransformTests):
     def setup_class(cls):
         cls.transformer_name = "ArbitraryImputer"
 
-    def test_impute_value_preserve_dtype(self, downcast_df):
-        """Test that downcast dtypes are preserved after imputation."""
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize(
+        ("column", "col_type", "impute_value"),
+        [
+            ("a", "String", 1),
+            ("a", "Categorical", True),
+            ("b", "Float32", "bla"),
+            ("c", "Boolean", 500),
+        ],
+    )
+    def test_type_mismatch_errors(
+        self,
+        column,
+        col_type,
+        impute_value,
+        library,
+    ):
+        """Test that dtypes are preserved after imputation."""
 
-        df_nw = nw.from_native(downcast_df)
+        df = impute_df_with_several_types(library=library)
 
-        df_nw = df_nw.with_columns(
-            nw.col("a").cast(nw.Int8),
-            nw.col("b").cast(nw.Float32),
+        df = nw.from_native(df)
+
+        df = df.with_columns(
+            nw.col(column).cast(getattr(nw, col_type)),
         )
 
-        x = ArbitraryImputer(impute_value=1, columns=["a", "b"])
-        df_transformed_native = x.transform(df_nw.to_native())
+        df = nw.to_native(df)
+
+        transformer = ArbitraryImputer(impute_value=impute_value, columns=[column])
+
+        if col_type in ["Categorical", "String"]:
+            msg_required_impute_value_type = "str"
+            msg_col_type = "Categorical or String"
+
+        elif col_type == "Boolean":
+            msg_required_impute_value_type = "bool"
+            msg_col_type = "Boolean"
+
+        else:
+            msg_required_impute_value_type = "numeric"
+            msg_col_type = "Numeric"
+
+        msg = rf"""
+                {self.transformer_name}: Attempting to impute non-{msg_required_impute_value_type} value {transformer.impute_value} into
+                {msg_col_type} type columns, this is not type safe,
+                please use {msg_required_impute_value_type} impute_value for these columns
+                \(this may require separate ArbitraryImputer instances for different column types\)
+                """
+
+        with pytest.raises(
+            TypeError,
+            match=msg,
+        ):
+            transformer.transform(df)
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize(
+        ("column", "col_type", "impute_value"),
+        [
+            ("a", "String", "z"),
+            ("a", "Categorical", "z"),
+            ("b", "Float32", 1),
+            ("c", "Boolean", True),
+        ],
+    )
+    def test_impute_value_preserve_dtype(
+        self,
+        column,
+        col_type,
+        impute_value,
+        library,
+    ):
+        """Test that dtypes are preserved after imputation."""
+
+        df = impute_df_with_several_types(library=library)
+
+        df_nw = nw.from_native(df)
+
+        df_nw = df_nw.with_columns(
+            nw.col(column).cast(getattr(nw, col_type)),
+        )
+
+        transformer = ArbitraryImputer(impute_value=impute_value, columns=[column])
+        df_transformed_native = transformer.transform(df_nw.to_native())
 
         df_transformed_nw = nw.from_native(df_transformed_native)
 
-        assert df_transformed_nw["a"].dtype == df_nw["a"].dtype
-        assert df_transformed_nw["b"].dtype == df_nw["b"].dtype
+        expected_dtype = df_nw[column].dtype
+        actual_dtype = df_transformed_nw[column].dtype
+        assert (
+            actual_dtype == expected_dtype
+        ), f"{self.transformer_name}: dtype changed unexpectedly in transform, expected {expected_dtype} but got {actual_dtype}"
 
     @pytest.mark.parametrize(
         ("library", "expected_df_4", "impute_values_dict"),
--- a/tests/imputers/test_ArbitraryImputer.py
+++ b/tests/imputers/test_ArbitraryImputer.py
@@ -1,5 +1,4 @@
 import narwhals as nw
-import pandas as pd
 import pytest
 
 import tests.utils as u
@@ -14,16 +13,19 @@ from tubular.imputers import ArbitraryImputer
 
 
 # Dataframe used exclusively in this testing script
-def create_downcast_df(library="pandas"):
-    """Create a dataframe with mixed dtypes to use in downcasting tests."""
-    df = pd.DataFrame(
-        {
-            "a": [1, 2, 3, 4, 5],
-            "b": [1.0, 2.0, 3.0, 4.0, 5.0],
-        },
-    )
+@pytest.fixture(params=["pandas", "polars"])
+def downcast_df(request):
+    """
+    Fixture that returns a DataFrame with columns suitable for downcasting
+    for both pandas and polars.
+    """
+    data = {
+        "a": [1, 2, 3, 4, 5],
+        "b": [1.0, 2.0, 3.0, 4.0, 5.0],
+    }
+    library = request.param
 
-    return u.dataframe_init_dispatch(df, library)
+    return u.dataframe_init_dispatch(data, library)
 
 
 class TestInit(ColumnStrListInitTests):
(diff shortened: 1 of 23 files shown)


tubular/imputers.py
```diff
@@ -113,42 +111,44 @@ def __init__(
...
-        Additions
-        ---------
-        * Preserving the datatypes of columns
-        * Finding the target column dtype and cast imputer values as same dtype
         """
+
         self.check_is_fitted(["impute_value"])
         self.columns_check(X)
+
+        X = nw.from_native(X)
```
liamholmes31, line 133: Q: think you shouldn't need this with the narwhalify decorator?

tests/imputers/test_ArbitraryImputer.py
```diff
@@ -12,15 +14,17 @@
 
 
 # Dataframe used exclusively in this testing script
-def create_downcast_df():
+def create_downcast_df(library="pandas"):
     """Create a dataframe with mixed dtypes to use in downcasting tests."""
-    return pd.DataFrame(
+    df = pd.DataFrame(
```
davidhopkinson26, line 19: S: Think the way to do this is to define the data as a dictionary before passing to dataframe_init_dispatch.

tubular/imputers.py
```diff
@@ -113,42 +111,44 @@ def __init__(
...
+            # Handle categorical column cases explicitly using Polars' syntax with `nw`
             if (
-                "category" in X[c].dtype.name
+                X[c].dtype == nw.Categorical
                 and self.impute_value not in X[c].cat.categories
             ):
-                X[c] = X[c].cat.add_categories(
-                    self.impute_value,
-                )  # add new category
+                X[c] = X[c].cat.add_categories(self.impute_value)  # add new category
```
liamholmes31, line 142: S: narwhals/polars don't seem to support adding categories in the same way as pandas does, what I would suggest is:

- check if column is categorical, if so store this in a var e.g. `CATEGORICAL=True` and convert to str
- apply BaseImputer.transform
- if CATEGORICAL, convert the col back to category before returning (this part already happens anyway with the dtype conversions at the end)

tests/imputers/test_ArbitraryImputer.py
davidhopkinson26: This would also be better as an @pytest.fixture so you can pass it as an arg to tests instead of calling within the body of tests

liamholmes31, line 19: S: need to pass a dict into the dataframe_init_dispatch function, so would remove the DataFrame call here

tubular/imputers.py
davidhopkinson26, line 133: shouldn't need the form native call when we are using the @narwhalify decorator

tubular/imputers.py
```diff
@@ -113,42 +111,44 @@ def __init__(
...
         self.columns_check(X)
+
+        X = nw.from_native(X)
+
+        new_col_expressions = []
         for c in self.columns:
+            # Handle categorical column cases explicitly using Polars' syntax with `nw`
             if (
-                "category" in X[c].dtype.name
+                X[c].dtype == nw.Categorical
```
davidhopkinson26, line 139: N: prefer using the polars syntax for selecting columns now that we can:

`X.select(c).dtype

tubular/imputers.py
alphanumericmale: removed

tubular/imputers.py
```diff
@@ -113,42 +111,44 @@ def __init__(
...
                 and self.impute_value not in X[c].cat.categories
             ):
-                X[c] = X[c].cat.add_categories(
-                    self.impute_value,
-                )  # add new category
+                X[c] = X[c].cat.add_categories(self.impute_value)  # add new category
+
+            # Apply fill_null() properly within Narwhals
+            new_col_expressions.append(nw.col(c).fill_null(self.impute_values_[c]))
```
davidhopkinson26, line 145: think we should use super().transform(X) here for consistency?

tubular/imputers.py
alphanumericmale: removed

tubular/imputers.py
alphanumericmale: I think what i've pushed covers runs in to the same issue as this would - converting null values to strings turns them into a value that aren't imputed correctly

tests/imputers/test_ArbitraryImputer.py
alphanumericmale: have implemented

alphanumericmale: have implemented

tubular/imputers.py
alphanumericmale: dont think this is relevant anymore after recent push

tubular/imputers.py
```diff
@@ -113,42 +111,44 @@ def __init__(
...
+
+        X = nw.from_native(X)
+
+        new_col_expressions = []
         for c in self.columns:
+            # Handle categorical column cases explicitly using Polars' syntax with `nw`
             if (
-                "category" in X[c].dtype.name
+                X[c].dtype == nw.Categorical
                 and self.impute_value not in X[c].cat.categories
```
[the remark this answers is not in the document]
alphanumericmale: thanks, i believe this section works now

tubular/imputers.py
liamholmes31: damn!

here's the new proposed solution we discussed earlier:

```
X=X.with_columns(
        nw.when(nw.col(column).is_null())
        .then(nw.lit(impute_value))
        .otherwise(nw.col(column).cast(nw.String))
        .cast(nw.Categorical)
        .alias(column)
)
```

liamholmes31:
I think one of the other issues is coming from here:
https://github.com/azukds/tubular/blob/9308197248d24d5452315b742843b2c3f67007bc/tests/imputers/test_BaseImputer.py#L222C1-L222C17

Should be:
```
expected_df_3["c"] = expected_df_3["c"].cat.add_categories(
                transformer.impute_values_["c"],
            )
```

tubular/imputers.py
```diff
@@ -113,43 +109,45 @@ def __init__(
...
-        # casting imputer value as same dtype as original column
+            original_dtypes[c] = X[c].dtype
+
+        # If originally categorical, cast to nw.String so fill won't fail
+        X_transformed = X
+
         for c in self.columns:
-            dtype = X[c].dtype  # get the dtype of original column
-            X_transformed[c] = X_transformed[c].astype(dtype)
+            X_transformed = X_transformed.with_columns(
```
liamholmes31, line 144: this is good, but needs to be inside an 'if c is categorical' type statement

then in the else part, we would do:
`X_transformed=super().transform(X)`

tubular/imputers.py
```diff
@@ -113,43 +109,45 @@ def __init__(
...
-                )  # add new category
-
-        # Calling the BaseImputer's transform method to impute the values
-        X_transformed = super().transform(X)
-
-        # casting imputer value as same dtype as original column
+            original_dtypes[c] = X[c].dtype
+
+        # If originally categorical, cast to nw.String so fill won't fail
+        X_transformed = X
```
liamholmes31, line 141: can delete lines 140/141

tubular/imputers.py
```diff
@@ -113,43 +109,45 @@ def __init__(
...
+
         for c in self.columns:
```
liamholmes31, line 143: we already have a loop like this on line 137, so would just move this logic to the end of that loop

Boluwatife28: Okay, done that
 https://github.com/azukds/tubular/blob/f28ef505c4002c5988965a6cb12cd525326a3be6/tubular/imputers.py#L142-L143

liamholmes31:
Note, also:

- cleaned up some leftover debug prints in package
- updated narwhals to avoid CI fails from deprecated argument

tests/imputers/test_BaseImputer.py
cjmwills, line 1: S: there's a few "if not polars compatible" statements here, can they be deleted now?

tests/imputers/test_ArbitraryImputer.py
```diff
@@ -61,40 +62,81 @@ class TestTransform(GenericImputerTransformTests, GenericTransformTests):
     def setup_class(cls):
         cls.transformer_name = "ArbitraryImputer"
 
-    # Unit testing to check if downcast datatypes of columns is preserved after imputation is done
-    def test_impute_value_preserve_dtype(self):
-        """Testing downcast dtypes of columns are preserved after imputation using the create_downcast_df dataframe.
+    def test_impute_value_preserve_dtype(self, downcast_df):
```
cjmwills, line 65: Q: the imputer here is applied but no actual imputation is happening given there are no missing values in the dataframe. Is that the behaviour we're wanting to test here?

tests/imputers/test_ArbitraryImputer.py
```diff
@@ -12,38 +14,37 @@
...
+    return u.dataframe_init_dispatch(data, library)
 
 
 class TestInit(ColumnStrListInitTests):
     """Generic tests for transformer.init()."""
 
-    @classmethod
-    def setup_class(cls):
-        cls.transformer_name = "ArbitraryImputer"
+    # overload some inherited arg tests that have been replaced by beartype
```
cjmwills, line 35: Q: is there a plan for removing these tests in future as we switch to using beartype?

tubular/imputers.py
```diff
@@ -87,71 +89,68 @@ class ArbitraryImputer(BaseImputer):
...
+    polars_compatible = True
     FITS = False
 
+    @beartype
     def __init__(
         self,
-        impute_value: float | str,
-        columns: str | list[str],
-        **kwargs: dict[str, bool],
+        impute_value: Union[int, float, str, bool],
```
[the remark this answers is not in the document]
liamholmes31: yeah I think this should be OK, i imagine it was just left out by oversight previously, unless you can think of a reason to exclude?

We have a test that makes sure the falsey value False works, so one potential issue covered there

tests/imputers/test_ArbitraryImputer.py
liamholmes31: yeah, I think i'll set up a load of tickets for beartype conversions once we're done with the narwhals work, and we'll do those from the top level (BaseTransformer) down which will remove the need for these overloads

Am just making some of the conversions now where transformer updates would require more arg checks, as converting to beartype is easier

tubular/imputers.py
```diff
@@ -87,71 +89,68 @@ class ArbitraryImputer(BaseImputer):
...
+            for col in self.columns:
+                if str(original_dtypes[col]) == "Categorical" and (
+                    self.impute_value not in X[col].cat.categories
+                ):
+                    X[col] = X[col] = X["c"].cat.add_categories(
+                        self.impute_value,
+                    )
+            X = nw.from_native(X)
+
+        return nw.from_native(super().transform(X))
```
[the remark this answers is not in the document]
liamholmes31: so super.transform returns either a polars or pandas df, but you're right that at the point of return it's unneccessary, so I did remove here:
https://github.com/azukds/tubular/pull/390/commits/2b7704979e193d99666594a440e7f2b4e3d538b4

but then while making further changes around type handling it was necessary to bring back
https://github.com/azukds/tubular/pull/390/commits/d08906abbd05171bc6c325aea45c96d200615247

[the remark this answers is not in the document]
cjmwills: Ah right thanks

tubular/imputers.py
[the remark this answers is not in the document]
cjmwills: No reason I can think of to be fair, and like you said the test is a good check for it. I've also just noticed the docstring for this will need updating to include bool. Currently it's `impute_value : int or float or str`

tests/imputers/test_ArbitraryImputer.py
cjmwills: Yep makes sense, thanks

tests/imputers/test_ArbitraryImputer.py
cjmwills: Thanks, good idea! There looks like unexpected behaviour in `test_impute_value_preserve_dtype`, will leave a separate comment for that

tests/imputers/test_ArbitraryImputer.py
```diff
@@ -61,40 +60,157 @@ class TestTransform(GenericImputerTransformTests, GenericTransformTests):
...
+                \(this may require separate ArbitraryImputer instances for different column types\)
+                """
+
+        with pytest.raises(
+            TypeError,
+            match=msg,
+        ):
+            transformer.transform(df)
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
```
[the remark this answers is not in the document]
cjmwills: The problem being the tests pass because the dtype has still been preserved

tubular/imputers.py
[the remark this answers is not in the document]
liamholmes31: ah thanks, updated here https://github.com/azukds/tubular/pull/390/commits/bc08b91b9b8b25f0a57d7d01ae69ca956b17cae9

tests/imputers/test_ArbitraryImputer.py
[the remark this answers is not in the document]
liamholmes31: ah good catch, was a couple of things going on here:

- faulty setup in the tests, casting input columns like [True, None] to bool yielded [True, False]
- issues with type handling, have made this even stricter

Sorry this is split over a couple of commits as committed once before the weekend

https://github.com/azukds/tubular/pull/390/commits/ffde8638fd764715fe4326790319f075d50e8760

https://github.com/azukds/tubular/pull/390/commits/404da17934f411f0bb56cd3a8892b90c39c79439

tubular/imputers.py
```diff
@@ -80,78 +84,199 @@ class ArbitraryImputer(BaseImputer):
...
+        num_columns = set(self.columns).intersection(X.select(ncs.numeric()).columns)
+        bool_columns = set(self.columns).intersection(X.select(ncs.boolean()).columns)
+        str_columns = set(self.columns).intersection(X.select(ncs.string()).columns)
+        null_columns = set(self.columns).intersection(
+            [col for col, dtype in X.schema.items() if (dtype == nw.Unknown)],
+        )
+
+        # start with object columns, which can be a massive nuisance from pandas
+        pandas_object_cols_to_polars_types = {}
+        if len(object_columns) > 0 and nw.get_native_namespace(X).__name__ == "pandas":
```
cjmwills, line 147: B: I wasn't able to get this code block to activate. When using a pandas df as input it doesn't get picked up when determining `object_columns`, possibly because it's searching for `nw.Object`.

cjmwills: Also had similar issue for `null_columns` where it always seemed to be empty for pandas dfs.

tests/imputers/test_ArbitraryImputer.py
```diff
@@ -61,40 +64,316 @@ class TestTransform(GenericImputerTransformTests, GenericTransformTests):
...
+    @pytest.mark.parametrize(
+        ("column", "col_type", "impute_value", "expected_values"),
+        [
+            ("a", "String", "z", ["a", "b", "c", "d", "z"]),
```
cjmwills, line 127: Q: was the boolean row removed on purpose here? There used to be `("c", "Boolean", True, [True, False, False, True, True])` which checked for dtype preservation of bool. I think it's covered in `test_edge_cases` but line 147 has an if statement that looks for bool cols.

tests/imputers/test_ArbitraryImputer.py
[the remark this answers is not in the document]
cjmwills: Thanks! Have added a couple of follow up question but will close this thread

tests/imputers/test_ArbitraryImputer.py
liamholmes31: yeah sorry should have made clear - I moved to test edge cases as the behaviour is slightly different, a bool column will come in a object for pandas and then the type is purposefully changed to boolean at the end, so the input_type==output_type check will fail, but the behaviour is as intended (so felt it sat better in 'edge_cases'

will remove the if statement here

tubular/imputers.py
liamholmes31: re null being empty for pandas dfs, this is actually expected as pandas doesn't have a dedicated null column type in the way that polars does, so an all null pandas column would be e.g. str for [None] and float for [np.NaN], and go through the str/numeric handling

the object one I will look into now!

tests/imputers/test_ArbitraryImputer.py
liamholmes31: https://github.com/azukds/tubular/pull/390/commits/0b91057f75558ba8d99baa9c7e7e5b4f651f9ca8

tubular/imputers.py
liamholmes31: just did a run of 'test_edge_cases' with some extra prints/a breakpoint, and for the pandas version of case:
`([True, False, None], "Boolean", True, [True, False, True]),`

The column is a mix on bool/null and so comes in as object, then did appear in object_columns as expected - do you remember the case you tried that didn't work?

cjmwills: I was adding a column of strings to the `cali_df` example in the docs. But think I was getting confused, as this comes in as an object column in pandas but is picked up by `str_columns`, which actually seems correct but I was expecting to see it in the `object_columns`.

cjmwills:
We should be ready to merge this now, thanks for all your work on that!

cjmwills (approved):

Issue #391, opened 2025-03-11
[Bug]: Imputers do not impute 0 value in v1.4.1


nedwebster:
### What happened?

Imputers have no effect when imputing zero in version 1.4.1

Line 58 seems to be the issue, since the 0 evaluates to false.
https://github.com/azukds/tubular/blob/9308197248d24d5452315b742843b2c3f67007bc/tubular/imputers.py#L56-L61

Happy to put in a quick fix to replace the line with:

```python
if self.impute_values_[c] is not None
```

Side note, cool to see so much progress and development in Tubular. I'm still using ~3 years on :)

### Environment

OS: MacOS
Python 3.10
Tubular 1.4.1
Scikit-learn: 1.6.1

### Minimum reproducible code

```python
import pandas as pd
from sklearn.pipeline import Pipeline
from tubular.imputers import ArbitraryImputer

data = pd.DataFrame({"feature_1": [1, None]})
pipeline = Pipeline([("imputer", ArbitraryImputer(columns=["feature_1"], impute_value=0))])

pipeline.fit(data)
transformed_data = pipeline.transform(data)

print(transformed_data)
```

### Relevant error output

```shell
Expected output:

   feature_1
0        1.0
1        0.0



Actual output:

   feature_1
0        1.0
1        NaN

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

liamholmes31:
heya Ned, good to hear from ya - hope you're well :)

Thanks for raising, looks like a clear bug - would be happy to review and get a release out quickly for you if you are happy to PR that!

davidhopkinson26:
Hi Ned, Thanks for raising this!

Yes, that looks like a good fix so please do go ahead and raise a PR and we can get a release out with the fix.

Good to hear you are using the package.  Please keep reporting issues if you find them!  Contributions are always welcome :)

nedwebster:
PR is up with just the change. You might want it to be covered by the tests, but I'll leave that up to you!

PR #392, opened 2025-03-18
fix imputer bug when imputing 0

nedwebster:
Resolving issue: https://github.com/azukds/tubular/issues/391


nedwebster:
I cannot replicate the failing test locally, not sure what is going on there.

PR #393, opened 2025-03-18
bugfix for arb imputer falsey values

liamholmes31:
https://github.com/azukds/tubular/issues/391

Wasn't able to add test for impute_value=False with current setup due to casting that arbitrary imputer performs, but will add this to the narwhals refactor PR


davidhopkinson26 (approved):
changes look good.  Thanks @liamholmes31 and @nedwebster!

PR #394, opened 2025-03-18
Feature/nw datediffleapyeartransformer

ChristinaKatsiafa:
This is a pull request to convert datediffleapyeartransformer to narwhals.


tubular/dates.py
```diff
@@ -381,27 +381,42 @@ def calculate_age(self, row: pd.Series) -> int:
...
-        X : pd.DataFrame
-            Transformed data with new_column_name column.
+        X : pd/pl.DataFrame
+            Data containing self.columns
 
         """
 
-        X = super().transform(X)
+        X = nw.from_native(super().transform(X))
+        age = pd.DataFrame(
```
davidhopkinson26, line 404: Q: was it intentional to use pd.DataFrame here?

davidhopkinson26: think we need to refactor calculate_age to narwhals as part of this

davidhopkinson26 (changes requested):
thanks for this @ChristinaKatsiafa.  Think there is still a bit to do, as I think you need to also narwhalify calculate_age so we can remove the pandas specific logic in transform.

tubular/dates.py
```diff
@@ -341,67 +341,69 @@ def __init__(
...
+                )
+                .otherwise(
+                    nw.col(self.new_column_name),
+                )
+                .cast(nw.Int64)
+                .alias(self.new_column_name),
+            )
 
-        X[self.new_column_name] = X.apply(self.calculate_age, axis=1)
+        X = X.to_native()
```
davidhopkinson26, line 404: Q: not sure I get why we are doing this last step, what dtype is coming out of the above operation?  Understand we were getting issues with the dtype in the expected dataframes for tests when nulls present but did not realise we were also casting here.

CHANGELOG.rst
```diff
@@ -38,6 +38,7 @@ Changed
 - narwhalified DropOriginalMixin `#352 <https://github.com/lvgig/tubular/issues/352>_`
 - narwhalified BaseMappingTransformer `#367 <https://github.com/lvgig/tubular/issues/367>_`
 - narwhalified BaseMappingTransformerMixin. As part of this made mapping transformers more 
+- narwhalified DateDiffLeapYearTransformer.
```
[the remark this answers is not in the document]
ChristinaKatsiafa: Yes this is correct. So I started that as a practice while I picked the DateDifferenceTransformer. So you told me to do 2 PRs and not to create an issue.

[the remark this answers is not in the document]
ChristinaKatsiafa: This is added at the end now

tubular/dates.py
ChristinaKatsiafa: This is because narwhals returns the results as int64[pyarrow] instead of Int64. So I just force the change.

tubular/dates.py
```diff
@@ -341,67 +341,69 @@ def __init__(
...
-            Transformed data with new_column_name column.
+        X : pd/pl.DataFrame
+            Data containing self.columns
 
         """
 
-        X = super().transform(X)
+        X = nw.from_native(super().transform(X))
+
+        X = X.with_columns(
```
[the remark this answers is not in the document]
ChristinaKatsiafa: Comment added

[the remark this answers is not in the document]
davidhopkinson26: thanks, much clearer

tubular/dates.py
ChristinaKatsiafa: The change is ready to review :)

davidhopkinson26: looks good, thanks!

davidhopkinson26 (approved):

PR #398, opened 2025-03-26
narwhalify SetValueTransformer

davidhopkinson26:
Narwhalifying SetValueTransformer for polars compatibility.

Also extending testing to cover more values being set (was previously just testing setting a string value)

Draft because for integers we get int32 return while expected is  currently int64. <- edited test to cast to int32 as transformer does.

### DIFF
--- a/tests/misc/test_SetValueTransformer.py
+++ b/tests/misc/test_SetValueTransformer.py
@@ -1,5 +1,5 @@
+import polars as pl
 import pytest
-import test_aide as ta
 
 import tests.test_data as d
 from tests.base_tests import (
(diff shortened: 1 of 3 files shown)


tests/misc/test_SetValueTransformer.py
```diff
@@ -34,29 +55,22 @@ class TestTransform(GenericTransformTests):
...
         cls.transformer_name = "SetValueTransformer"
 
-    def expected_df_1():
-        """Expected output of test_value_set_in_transform."""
-        df = d.create_df_2()
-
-        df["a"] = "a"
-        df["b"] = "a"
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize("value", ["a", 1, 1.0, None])
```
ChristinaKatsiafa, line 59: I can see that in the transformer we allow all types and we dont have options for nans. So should we also add a missing value test here?

ChristinaKatsiafa (changes requested):

tubular/misc.py
```diff
@@ -42,7 +48,8 @@ def __init__(
 
         super().__init__(columns=columns, copy=copy, **kwargs)
 
-    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+    @nw.narwhalify
+    def transform(self, X: FrameT) -> FrameT:
```
ChristinaKatsiafa, line 52: Please update the docstring to match pd/pl.DataFrame

tests/misc/test_SetValueTransformer.py
davidhopkinson26: yep fair point.  Was kind of expecting this to break for polars but it works still! [1251699](https://github.com/azukds/tubular/pull/398/commits/1251699d4592180ec2073fa2fc1994d87cec2fb3)

tubular/misc.py
davidhopkinson26: [3989f78](https://github.com/azukds/tubular/pull/398/commits/3989f780dc37bd6eeaa3732a8391408d5ddf1a65)

ChristinaKatsiafa (approved):
I am happy with the changes :)

PR #401, opened 2025-04-01
Feature/nw betweendatestransformer

ChristinaKatsiafa:
This is a draft PR so David can have a look at the error I get from DropOriginalTransformMixinTests for polars. All othet tests passes and there is only one failing but I cant figure out why.


davidhopkinson26:
This one was a bit of a head scratcher but I think I have figured it out.

The minimal dataframe for between dates transformer does not include any additional columns so this test breaks down as 'other_columns' is an empty list so indexing with it does weird things and i think in the case of polars returns the entire dataframe.

Need to update the config in two places:

- in test_data, add another column, say "d" to the dataframe in  create_is_between_dates_df_1 so there is something there to be left alone!
- in conftest update the minimal_attributes dict for BetweenDatesTransformer to assign the new column as a new column (say "e").  At present the drop original mixin will drop the original column even if that is the one which has been assigned as the target.

davidhopkinson26 (dismissed):
Thanks Christina, looks good!

davidhopkinson26 (approved):

PR #408, opened 2025-04-09
Feature/1dkmeans

ChristinaKatsiafa:
I selected the name OneDKmeansTransformer and added this into the numeric transformers. The build and tests are ready. Please let me know if I missed any necessary tests or you would like a different structure.


tubular/numeric.py
```diff
@@ -911,3 +912,191 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+            msg = f"{self.classname()}: X should not contain missing values."
+            raise ValueError(msg)
+
+        kmeans = KMeans(
+            n_clusters=self.n_clusters,
+            n_init=self.n_init,
+            **self.kmeans_kwargs,
+        )
+
+        native_namespace = nw.get_native_namespace(X).__name__
```
davidhopkinson26, line 1047: Q: Might be missing something but I can't see why we need to specify this and do the to_native call below? Can't we just work with the polars style handling in narwhals like usual regardless of native namespace?

tubular/numeric.py
```diff
@@ -911,3 +912,191 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+
+        Returns
+        -------
+        X : pl/pd.DataFrame
+            Input X with additional cluster column added.
+        """
+        X = super().transform(X)
+
+        X = nw.from_native(X)
+        native_namespace = nw.get_native_namespace(X).__name__
```
davidhopkinson26, line 1087: same here, can we just leave backend as None?

davidhopkinson26 (changes requested):
Thanks for this Christina! I have taken a look at the changes in numeric.py and like the general form but I think we might be doing quite a few unnescessary calls to from_native and defining a backend for new_series.  Happy to go over this with you to understand better.

tubular/numeric.py
ChristinaKatsiafa: Hey. I tried that but I got an error ValueError: `backend` must be specified in `new_series`. this is mainly because the groups are returned as a numpy array so when this is passed into a series the backend needs to be specified. Finally removing to_native was giving me ValueError: output array is read-only for some reason. The solutions are either use to_native or to_numpy.

tubular/numeric.py
ChristinaKatsiafa: See the above comment please

tubular/numeric.py
```diff
@@ -911,3 +912,191 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+
+    n_clusters : int, default = 8
+        The number of clusters to form as well as the number of centroids to generate.
+
+    n_init "auto" or int, default="auto"
+        Number of times the k-means algorithm is run with different centroid seeds.
+        The final results is the best output of n_init consecutive runs in terms of inertia.
+        Several runs are recommended for sparse high-dimensional problems (see `Clustering sparse data with k-means <https://scikit-learn.org/stable/auto_examples/text/plot_document_clustering.html#kmeans-sparse-high-dim>`__).
+
+        When n_init='auto', the number of runs depends on the value of init: 10 if using init='random' or init is a callable;
```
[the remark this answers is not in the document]
davidhopkinson26: yep happy with this

tubular/numeric.py
davidhopkinson26: what happens if you just always set the backend to polars and rely on narwhalify converting back to pandas at return if needed?

ChristinaKatsiafa: All pandas tests fail. i think that behind the scenes it will convert them to pandas or polars and then work with either polars or pandas. So when I have polars series and try to add this into pandas then it fails.

davidhopkinson26: OK, fair enough.  It does seem to be necessary.  Would suggest we raise an issue on the narwhals board as the implementation seems to suggest you should be able to pass None so this may be a bug.

davidhopkinson26 (approved):

PR #419, opened 2025-05-21
narwhalified ToDatetimeTransformer

liamholmes31:
converted transformer to narwhals.

Also made usability improvements, to allow multiple columns to be processed by single transformer

https://github.com/azukds/tubular/issues/379

### DIFF
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import datetime
 import warnings
 import zoneinfo
-from typing import TYPE_CHECKING, Union
+from typing import TYPE_CHECKING, Optional, Union
 
 import narwhals as nw
 import narwhals.selectors as ncs
@@ -541,9 +541,15 @@ class ToDatetimeTransformer(BaseGenericDateTransformer):
     def __init__(
         self,
         columns: Union[str, list[str]],
-        time_format: str,
+        time_format: Optional[str] = None,
         **kwargs: dict[str, bool],
     ) -> None:
+        if not time_format:
+            warnings.warn(
+                "time_format arg has not been provided, so datetime format will be inferred",
+                stacklevel=2,
+            )
+
         self.time_format = time_format
 
         super().__init__(
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -5,12 +5,13 @@ from __future__ import annotations
 import datetime
 import warnings
 import zoneinfo
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Union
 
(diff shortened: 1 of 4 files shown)


tubular/dates.py
```diff
@@ -538,65 +533,40 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
-                if type(k) is not str:
-                    msg = f"{self.classname()}: unexpected type ({type(k)}) for to_datetime_kwargs key in position {i}, must be str"
-                    raise TypeError(msg)
-
-        self.to_datetime_kwargs = to_datetime_kwargs
-
-        # This attribute is not for use in any method, use 'columns' instead.
-        # Here only as a fix to allow string representation of transformer.
-        self.column = column
+        self.format = time_format
```
davidhopkinson26: N: any particular reason for the mismatch in names here? Would just go with `format` personally. I think  time is sufficiently implied by the context.

tubular/dates.py
```diff
@@ -538,65 +533,40 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
 
+    @beartype
     def __init__(
         self,
-        column: str,
-        new_column_name: str,
-        drop_original: bool = False,
-        to_datetime_kwargs: dict[str, object] | None = None,
+        columns: Union[str, list[str]],
+        time_format: str,
```
davidhopkinson26: Q: do we want to take advantage of the fact that to_datetime can infer the format if None is provided? https://narwhals-dev.github.io/narwhals/api-reference/expr_str/#narwhals.expr.ExprStringNamespace.to_datetime

Can see a case for demanding specification here but might be good to leave None as an option for quick analysis. maybe raise a warning that datetime format is being inferred?

tubular/dates.py
liamholmes31: ah it was becuase of this one:
`A002 Argument `format` is shadowing a Python builtin`
have gone the other way and set all to 'time_format'

tubular/dates.py
liamholmes31: good shout, have enabled here + added some tests https://github.com/azukds/tubular/pull/419/commits/64dc9a159050a0d0bbf561c9954f6ef98c096fab

tubular/dates.py
liamholmes31: https://github.com/azukds/tubular/pull/419/commits/7fbc285458ddca24910d6d71d2f61fb479186694

tubular/dates.py
davidhopkinson26: brilliant

PR #423, opened 2025-06-17
fixed mapping transformer issues with bool and null mappings

liamholmes31:
mapping transformers were struggling with mappings containing bools/nulls, this was due to:

- pandas crazy type handling for bools, resulting in nulls in bool cols being converted to False
- the merge mapping logic failing for null values, which were not being merged on

Updated logic to:
- carefully force pandas to use best bool type at each stage
- handle null mappings separately as imputations
    - considered removing null mappings altogether, but they are needed for inverse pipelines 

Example Failure Cases
Example failure 1:
df=pd.DataFrame({'a': [True, False, None]})
mappings={True:1, False:0, None:0}
return_dtypes={'a': 'Int8'}
mapping_transformer=MappingTransformer(mappings, return_dtypes)
mapping_transformer.transform(df)->
pd.DataFrame(
    {
    'a': [
        1,
        0,
        None # mapping merge has failed on None, resulting in None instead of 0
    ]
    }
)
---------
Example Failure 2
df=pd.DataFrame({'a': [1, 0, -1]})
mappings={1:True, 0:False, -1:None}
return_dtypes={'a': 'Int8'}
mapping_transformer=MappingTransformer(mappings, return_dtypes)
mapping_transformer.transform(df)->
pd.DataFrame(
  {
  'a': [
      True,
      False,
      # when the mapping values are put into bool series
      # the none value is converted to False, instead of None
      False,
  ]
  }
)

### DIFF
--- a/tests/_utils/test_new_narwhals_series_with_best_pandas_types.py
+++ b/tests/_utils/test_new_narwhals_series_with_best_pandas_types.py
@@ -1,0 +1,115 @@
+import narwhals as nw
+import pandas as pd
+import polars as pl
+import pytest
+from pandas.api.types import CategoricalDtype
+from pandas.testing import assert_series_equal as assert_series_equal_pandas
+from polars.testing import assert_series_equal as assert_series_equal_polars
+
+from tubular._utils import (
+    new_narwhals_series_with_best_pandas_types,  # noqa: PLC2701, importing from private module to test
+)
+
+
+class TestNewNarwhalsSeriesWithBestPandasTypes:
+    @pytest.mark.parametrize(
+        ("values", "dtype"),
+        [
+            ([1, 0, 0], "Int64"),
+            ([1, 0, None], "Float64"),
+            (["a", "b", None], "String"),
+            ((["a", None, "c"], "Categorical")),
+            ([True, False, True], "Boolean"),
+            ([True, False, None], "Boolean"),
+        ],
+    )
+    def test_polars_unaffected(self, values, dtype):
+        "test that polars Series are initialised as usual"
+        name = "a"
+        output = new_narwhals_series_with_best_pandas_types(
+            name=name,
+            values=values,
+            backend="polars",
+            dtype=getattr(nw, dtype),
+        )
+
+        expected = pl.Series(
+            name=name,
+            values=values,
+            dtype=getattr(pl, dtype),
+        )
+
+        assert_series_equal_polars(expected, output.to_native())
+
+    @pytest.mark.parametrize(
+        ("values", "polars_dtype", "expected_pandas_dtype"),
+        [
+            ([1, 0, 0], "Int64", "Int64"),
+            ([1, 0, None], "Float64", "Float64"),
+            ([True, False, True], "Boolean", "boolean"),
+            ([True, False, None], "Boolean", "boolean"),
+        ],
+    )
+    def test_pandas_output(self, values, polars_dtype, expected_pandas_dtype):
+        "test that polars Series are initialised as usual"
+        name = "a"
+        output = new_narwhals_series_with_best_pandas_types(
+            name=name,
+            values=values,
+            backend="pandas",
+            dtype=getattr(nw, polars_dtype),
+        )
+
+        expected = pd.Series(
+            name=name,
+            data=values,
+            dtype=expected_pandas_dtype,
+        )
+
+        assert_series_equal_pandas(expected, output.to_native())
+
+    def test_pandas_output_category(self):
+        """test that polars Series are initialised as usual
+        for category type which requires special handling"""
+        name = "a"
+        categories = ["a", "b", "c"]
+        cat_type = CategoricalDtype(
+            pd.Series(data=categories, dtype="string"),
+            ordered=False,
+        )
+        output = new_narwhals_series_with_best_pandas_types(
+            name=name,
+            values=categories,
+            backend="pandas",
+            dtype=nw.Categorical,
+        )
+
+        expected = pd.Series(
+            name=name,
+            data=categories,
+            dtype=cat_type,
+        )
+
+        assert_series_equal_pandas(expected, output.to_native())
+
+    def test_pandas_output_string(self):
+        """test that polars Series are initialised as usual
+        for string type which requires special handling"""
+        name = "a"
+        values = ["a", None, "c"]
+        output = new_narwhals_series_with_best_pandas_types(
+            name=name,
+            values=values,
+            backend="pandas",
+            dtype=nw.String,
+        )
+
+        expected = pd.Series(
+            name=name,
+            data=values,
+            dtype="string",
+        ).fillna(pd.NA)
+
+        print(expected)
+        print(output.to_native())
+        assert_series_equal_pandas(expected, output.to_native())
--- a/tests/_utils/test_new_narwhals_series_with_best_pandas_types.py
+++ b/tests/_utils/test_new_narwhals_series_with_best_pandas_types.py
@@ -110,6 +110,4 @@ class TestNewNarwhalsSeriesWithBestPandasTypes:
             dtype="string",
         ).fillna(pd.NA)
 
-        print(expected)
-        print(output.to_native())
         assert_series_equal_pandas(expected, output.to_native())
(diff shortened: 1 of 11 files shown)


tubular/_utils.py
```diff
@@ -46,3 +51,46 @@ def _assess_pandas_object_column(pandas_df: pd.DataFrame, col: str) -> tuple[str
     # we may wish to add more cases in future, but starting with these
 
     return pandas_col_type, polars_col_type
+
+
+def new_narwhals_series_with_best_pandas_types(
```
nesa1995: N: Could we rename this function to new_narwhals_series_with_optimal_pandas_dtypes

tubular/_utils.py
```diff
@@ -46,3 +51,46 @@ def _assess_pandas_object_column(pandas_df: pd.DataFrame, col: str) -> tuple[str
...
+    nw.Series: new narwhals series
+    """
+
+    if backend == "pandas":
+        series = pd.Series(name=name, data=values)
+        series = nw.maybe_convert_dtypes(nw.from_native(series, allow_series=True))
+        # maybe_convert_dtype already handles object->string, casting
+        # again would convert back, so avoid
+        # also avoid directly casting back to object..
+        if dtype != nw.String and dtype != nw.Object and dtype != nw.Boolean:
```
nesa1995: N: Can simplify this to if dtype not in {nw.String, nw.Object, nw.Boolean}:

tests/_utils/test_new_narwhals_series_with_best_pandas_types.py
```diff
@@ -0,0 +1,115 @@
...
+
+        expected = pd.Series(
+            name=name,
+            data=values,
+            dtype=expected_pandas_dtype,
+        )
+
+        assert_series_equal_pandas(expected, output.to_native())
+
+    def test_pandas_output_category(self):
```
nesa1995: Q: Could these two tests be done in the above with adding to the fixture?

tests/_utils/test_new_narwhals_series_with_best_pandas_types.py
```diff
@@ -0,0 +1,115 @@
...
+            dtype=nw.String,
+        )
+
+        expected = pd.Series(
+            name=name,
+            data=values,
+            dtype="string",
+        ).fillna(pd.NA)
+
+        print(expected)
```
nesa1995: N: Shall we remove the print statements?

tubular/mapping.py
```diff
@@ -67,15 +68,30 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
+        mappings: dict[str, dict[Any, Any]],
         return_dtypes: Union[dict[str, RETURN_DTYPES], None] = None,
         **kwargs: Optional[bool],
     ) -> None:
         if not len(mappings) > 0:
             msg = f"{self.classname()}: mappings has no values"
             raise ValueError(msg)
 
+        null_mappings = {col: None for col in mappings}
+        for col in mappings:
```
nesa1995, line 80: Wondering if we can condense this and use list comprehension here to gather the null_keys:

for col, col_mappings in mappings.items():

    null_keys = [key for key in col_mappings if pd.isna(key)]

    if len(null_keys) > 1:
        raise ValueError(
            f"Multiple mappings have been provided for null values in column '{col}'. "
            "Transformer is set up to handle nan/None/NA as one."
        )
    
        # Assign the mapping for the null key if it exists
    if null_keys:
        null_mappings[col] = col_mappings[null_keys[0]]

nesa1995 (changes requested):

tubular/_utils.py
nesa1995: N: Could we update the comment above as slightly confusing to follow

tubular/_utils.py
liamholmes31: good idea, actioned here https://github.com/azukds/tubular/pull/423/commits/fe8bcbe8e576e09e909d6256ff9b99b193389fd5

tubular/_utils.py
liamholmes31: updated here, let me know your thoughts! https://github.com/azukds/tubular/pull/423/commits/2db1bfa52ad5725e6d462ab59c3ff8aa6a71af8e

tests/_utils/test_new_narwhals_series_with_best_pandas_types.py
liamholmes31: thanks removed here https://github.com/azukds/tubular/pull/423/commits/a0664a1534334d8de8b1e63788907ea46a49ab55

tests/_utils/test_new_narwhals_series_with_best_pandas_types.py
liamholmes31: sure combined here https://github.com/azukds/tubular/pull/423/commits/8559777e8e884bc226555f42cf7154f16a63719e

tubular/mapping.py
liamholmes31: much nicer https://github.com/azukds/tubular/pull/423/commits/5392ade17cfd458f01c4dd8bdc711c9c37bd724f

tubular/_utils.py
nesa1995: Much easier to follow, thank you!

tubular/mapping.py
```diff
@@ -148,15 +164,15 @@ def transform(self, X: FrameT) -> FrameT:
             Transformed input X with levels mapped accoriding to mappings dict.
 
         """
-        self.check_is_fitted(["mappings", "return_dtypes"])
+        self.check_is_fitted(["mappings", "return_dtypes", "null_mappings"])
 
         X = nw.from_native(super().transform(X))
         native_backend = nw.get_native_namespace(X)
 
         # will do a join further down, which does not preserve index
```
adamsardar, line 172: Q: isn't this what `maybe_get_index` is for? ([docs](https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.maybe_get_index)). I notice their steer though:

> This is only really intended for backwards-compatibility purposes, for example if your library already aligns indices for users. If you're designing a new library, we highly encourage you to not rely on the Index. For non-pandas-like inputs, this returns None.

Why are we keeping track of an index? Feels like a feature that is easy to get wrong.

tests/mapping/test_MappingTransformer.py
```diff
@@ -223,6 +223,58 @@ def test_excess_mapping_values(self, mapping, mapped_col, return_dtypes, library
         ):
             x.transform(df)
 
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_expected_output_boolean(self, library):
```
adamsardar: I'm struggling with this review because I can't quite grok what the issues was in the first place, nor how that issue precipitated the bug. There isn't an example of the issues and the tests don't explain/show where the bug might creep in.

On this specific bit of code, some comments or clearer docstring to outline what the assertion is. Something about nullable bools?

Could you add some clarity as to what is being tested? 'tricky bool case' - why is it tricky? What is it to watch out for?

tubular/_utils.py
```diff
@@ -46,3 +51,47 @@ def _assess_pandas_object_column(pandas_df: pd.DataFrame, col: str) -> tuple[str
...
+
+    if backend == "pandas":
+        series = pd.Series(name=name, data=values)
+        series = nw.maybe_convert_dtypes(nw.from_native(series, allow_series=True))
+        # for int/float values, we may still want to cast to better type
+        # (e.g. int8)
+        # but for bool/str values, maybe_convert_types has already
+        # cast to improved typing, so avoid further casting for these
+        if dtype not in [nw.String, nw.Object, nw.Boolean]:
+            series = series.cast(dtype)
```
adamsardar, line 92: This will result in a heck of a lot of casts and affect performance.

I can't tell if this is the right approach because I can't see what the original problem was/is.

tests/mapping/test_MappingTransformer.py
liamholmes31: added some specific examples here https://github.com/azukds/tubular/pull/423/commits/4b0917a0c9cd95b37b27af052a61bdf58e4302f9

Issue #385, opened 2025-01-23
Add ColumnAggregationTransformer class


liamholmes31:
What?
Create new class in the numeric file (this transformer will have some functionality that generalises to non-numeric columns, e.g. mode aggregation, but still think it fits best here)

We want this class to have options for:

- min/max/mean/median/mode aggregations
- to work on either column or row level

To this end, I think we will ultimately want two transformers (RowAggregationTransformer and ColumnAggregationTransformer), which share much logic in a BaseAggregationTransformer class, I think this first class will need to implement:

- an init method, which:
    - checks and sets arguments:
        - columns, a list of columns to aggregate
        - aggregations, a list of aggregations to apply (min/mean/max/median/mode)
        - drop_original, a bool arg indicating whether to drop original columns at the end
        - level, str being one of 'row' or 'column'
        - these functions will need a way of creating new_col_names, possibly by arg or possibly just automated (e.g. 'col1_sum')

The two child classes will then just need to implement

- init method, for RowAggregationTransformer, also needs a key arg (giving the name of the datas key column)
- a transform method, which:
    - takes in a dataframe, and creates new columns using the agg method
    - for ColumnAggregationTransformer, this would look like df.with_columns(nw.col(col1)+nw.col(col2)+...)
    - For RowAggregationTransformer, this would look like df.group_by(nw.col(key)).agg(nw.col(col1).sum(), nw.col(col2).sum(), ...)

will also need to test this new object/methods, focus on outputs rather than implementations - e.g. are argument errors raised correctly? are the output columns as expected?

Why?
Help us to move away from soon to be deprecated DataFrameMethodTransformer

liamholmes31:
yeah fair comment, had to reread the ticket to remember which name I had intended for which so a bad sign.

happy with your suggestion, or if we want to go even more explicit:
- AggregateRowOverColumnsTransformer
- AggregateColumnOverGroupedRowsTransformer

thoughts?

liamholmes31:
https://github.com/azukds/tubular/pull/441

PR #424, opened 2025-06-17
Feature/row aggregator

hariinduu:
Added a row aggregation transformer with few different aggregations to work on row level

#385


tubular/aggregations.py
```diff
@@ -0,0 +1,162 @@
+import pandas as pd
+from beartype import beartype
+from mixins import DropOriginalMixin
+
+
+class BaseAggregationTransformer:
```
liamholmes31: Q: can this inherit from tubular BaseTransformer?

tubular/aggregations.py
```diff
@@ -0,0 +1,162 @@
...
+    def __init__(
+        self,
+        columns: list[str],
+        aggregations: list[str],
+        drop_original: bool = False,
+        level: str = "row",
+    ) -> None:
+        valid_aggregations = {"min", "max", "mean", "median", "mode", "sum", "count"}
+
+        if not all(agg in valid_aggregations for agg in aggregations):
```
liamholmes31: S: as discussed on our call, would be better to handle these checks with beartype

tubular/aggregations.py
liamholmes31: and DropOriginalMixin

tubular/aggregations.py
liamholmes31: can typehint aggregations like Literal["min", "max", "mean", "median", "mode", "sum", "count"]

tubular/aggregations.py
```diff
@@ -0,0 +1,162 @@
...
+        drop_original: bool = False,
+    ) -> None:
+        super().__init__(
+            columns=columns,
+            aggregations=aggregations,
+            drop_original=drop_original,
+            level="row",
+        )
+        self.key = key
+        self.set_drop_original_column(drop_original)
```
liamholmes31: S: this can be handled in the base class

tubular/aggregations.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @beartype
+    def transform(
```
liamholmes31, line 177: need to add the nw.narwhalify decorator here

tubular/aggregations.py
```diff
@@ -0,0 +1,162 @@
...
+        self,
+        df: pd.DataFrame,
```
liamholmes31: These type hints will need to be narwhals FrameT type

tubular/aggregations.py
```diff
@@ -0,0 +1,162 @@
...
+        ------
+        ValueError
+            If the key column is not found in the DataFrame.
+        """
+        if self.key not in df.columns:
+            msg = f"key '{self.key}' not found in dataframe columns"
+            raise ValueError(msg)
+
+        aggregation_dict = {col: self.aggregations for col in self.columns}
+        grouped_df = df.groupby(self.key).agg(aggregation_dict)
```
liamholmes31: looks good for pandas, but this will need to use narwhals syntax:
https://narwhals-dev.github.io/narwhals/api-reference/group_by/#narwhals.group_by.GroupBy.agg
```
grouped_df=df.group_by(self.key).agg(getattr(nw.col(col), agg)() for agg in aggs for col, aggs in aggregation_dict.items())
```
^ something like this should work

tubular/aggregations.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # Flatten MultiIndex columns
+        grouped_df.columns = [
```
liamholmes31: S: think this part and the reset index won't be needed in narwhals

tubular/aggregations.py
```diff
@@ -0,0 +1,162 @@
...
+            "_".join(col).strip() for col in grouped_df.columns.to_numpy()
+        ]
+
+        # Reset index to merge with original DataFrame
+        grouped_df = grouped_df.reset_index()
+
+        # Merge the aggregated results back with the original DataFrame
+        df = df.merge(grouped_df, on=self.key, how="left")
```
liamholmes31: narwhals uses .join fyi 
https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.join

tubular/aggregations.py
hariinduu: Done

tubular/aggregations.py
hariinduu: Done

tubular/aggregations.py
hariinduu: Aggregations passed as a list and used beartype. Also with levels

tubular/aggregations.py
hariinduu: Changed

tubular/aggregations.py
hariinduu: Changed

tubular/aggregations.py
hariinduu: Removed both

tubular/aggregations.py
hariinduu: Added

tubular/aggregations.py
hariinduu: Changed

pyproject.toml
```diff
@@ -115,6 +115,7 @@ ignore = [
     "B905", # the below two rules seem to make changes that error for python3.9
     "UP038", 
     "PYI041", # clashes with beartype
+    "F821",
```
liamholmes31: Q: how come this has been added?

tubular/aggregations.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Attributes
+    ----------
+    columns : list[str]
+        Columns to apply the transformations to.
+    aggregations : list[str]
+        Aggregation methods to apply.
+    drop_original : bool
+        Indicator for dropping original columns.
+    level : str
```
liamholmes31: Q: does this arg do anything in the end? might not be needed now we've decided to have 2 separate child classes

tubular/aggregations.py
```diff
@@ -0,0 +1,158 @@
...
+        grouped_df = df.group_by(self.key).agg(
+            *(
+                getattr(nw.col(col_name), agg)().alias(f"{col_name}_{agg}")
+                for col_name, aggs in aggregation_dict.items()
+                for agg in aggs
+            ),
+        )
+
+        # Merge the aggregated results back with the original DataFrame
+        df = df.join(grouped_df, on=self.key, how="left")
```
liamholmes31: G: in other versions of this, we have also deduped the df at this point. But I think I might prefer this, it seems more in the spirit of Do-One-Thing to do it this way and handle deduping elsewhere!

tests/aggregations/test_baseaggregationtransformer.py
```diff
@@ -0,0 +1,45 @@
+import pytest
+from beartype.roar import BeartypeCallHintParamViolation
+
+from tubular.aggregations import BaseAggregationTransformer
+
+
+class TestBaseAggregationTransformerInit:
```
liamholmes31: S: will want this one to inherit the following two from base_tests
 - DropOriginalInitMixinTests,
 - ColumnStrListInitTests,

liamholmes31: will also need to add this to this class, so that the inherited tests will work:

```
@classmethod
def setup_class(cls):
        cls.transformer_name = "BaseAggregationTransformer"
```

liamholmes31: and will need to add the two new classes to minimal_attribute_dict in conftest

tests/aggregations/test_baseaggregationtransformer.py
```diff
@@ -0,0 +1,45 @@
...
+    """Tests for BaseAggregationTransformer initialization."""
+
+    def test_valid_initialization(self):
+        """Test initialization with valid parameters."""
+        columns = ["col1", "col2"]
+        aggregations = ["min", "max", "mean"]
+        transformer = BaseAggregationTransformer(columns, aggregations)
+        assert transformer.columns == columns
```
liamholmes31: S: can remove these attr asserts, we've moved away from doing these as ties us to a specific implementation

tests/aggregations/test_baseaggregationtransformer.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def test_invalid_level_error(self):
+        """Test that an error is raised for invalid level."""
+        columns = ["col1", "col2"]
+        aggregations = ["min", "max"]
+        with pytest.raises(BeartypeCallHintParamViolation):
+            BaseAggregationTransformer(columns, aggregations, level="invalid")
+
+
+class TestBaseAggregationTransformerMethods:
```
liamholmes31: N: would be specific and rename TestBaseAggregationTransformerCreateNewColNames

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,150 @@
+import narwhals as nw
+import pandas as pd
+import pytest
+from beartype.roar import BeartypeCallHintParamViolation
+
+from tubular.aggregations import AggregateRowOverColumnsTransformer
+
+
+class TestAggregateRowOverColumnsTransformerInit:
```
liamholmes31: S: can you set this up to inherit from TestAggregateRowOverColumnsTransformerInit?

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,150 @@
...
+    """Tests for AggregateRowOverColumnsTransformer initialization."""
+
+    def test_valid_initialization(self):
```
liamholmes31: S: once the inheritance is added, will be able to delete this

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,150 @@
...
+        aggregations = ["min", "max", "mean"]
+        key = "group_key"
+        transformer = AggregateRowOverColumnsTransformer(columns, aggregations, key)
+        assert transformer.columns == columns
+        assert transformer.aggregations == aggregations
+        assert transformer.key == key
+        assert transformer.drop_original is False
+        assert transformer.level == "row"
+
+    def test_invalid_aggregation_type_error(self):
```
liamholmes31: S: once the inheritance is added, will be able to delete this

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,150 @@
...
+
+        transformer = AggregateRowOverColumnsTransformer(columns, aggregations, key)
+        with pytest.raises(
+            ValueError,
+            match=f"key '{key}' not found in dataframe columns",
+        ):
+            transformer.transform(df)
+
+
+class TestAggregateRowOverColumnsTransformerMethods:
```
liamholmes31: N: would rename `TestAggregateRowOverColumnsTransformerMethodsTransform`

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,150 @@
...
+    def test_invalid_aggregation_type_error(self):
+        """Test that an exception is raised for invalid aggregation type."""
+        columns = ["col1", "col2"]
+        aggregations = ["invalid_agg"]
+        key = "group_key"
+
+        with pytest.raises(BeartypeCallHintParamViolation):
+            AggregateRowOverColumnsTransformer(columns, aggregations, key)
+
+    def test_invalid_key_error(self):
```
liamholmes31: S: this belongs in the class below

liamholmes31: S: good first pass, but for all of these transform tests we'll need to edit to make them run for both pandas/polars:
- can see examples e.g. here https://github.com/azukds/tubular/blob/main/tests/imputers/test_BaseImputer.py
- need to add another parametrise for argument 'library' that takes values ['pandas', 'polars']
- when creating dataframes, need to use the dataframe_init_dispatch function found here https://github.com/azukds/tubular/blob/main/tests/utils.py
- when comparing dataframes, need to use the assert_frame_equal dispatch function found here https://github.com/azukds/tubular/blob/main/tests/utils.py
- do not need to convert dfs to narwhals before calling transform, narwhalify handles this

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,150 @@
...
+                "col1": [1, 2, 3],
+                "col2": [4, 5, 6],
+            },
+        )
+
+        # Convert to narwhals DataFrame using the conversion function
+        df = nw.from_native(pd_df)
+
+        # Assert the type to ensure it's a Narwhals DataFrame
+        assert isinstance(df, nw.DataFrame), "DataFrame should be a Narwhals DataFrame"
```
liamholmes31: N: don't need this check

pyproject.toml
hariinduu: It gives a ruff issue which refers to 'undefined name' when variable or function name is used but has not been defined or imported. In below part where 'col' is not actually defined within narwhals

getattr(nw.col(col_name), agg)().alias(f"{col_name}_{agg}")

tests/aggregations/test_baseaggregationtransformer.py
hariinduu: Renamed

tests/aggregations/test_baseaggregationtransformer.py
hariinduu: Yes, if we change the implementation in the future, it might fail. Removed the test.

tests/aggregations/test_baseaggregationtransformer.py
hariinduu: Done. Added verbose as a parameter to BaseAggregationTransformer

tests/aggregations/test_aggregation_transformers.py
hariinduu: Renamed

tests/aggregations/test_aggregation_transformers.py
hariinduu: Deleted

tests/aggregations/test_aggregation_transformers.py
hariinduu: Created a BaseTransformerTest class and inherited from that

tests/aggregations/test_aggregation_transformers.py
hariinduu: deleted

tests/aggregations/test_aggregation_transformers.py
hariinduu: Removed

tubular/aggregations.py
hariinduu: Yes, not needed and removed

pyproject.toml
liamholmes31: just tested, I think this was actually coming from the type hints and is no longer an issue with your new approach, so can remove this :)

tubular/aggregations.py
```diff
@@ -0,0 +1,182 @@
...
+
+from tubular.base import BaseTransformer
+from tubular.mixins import DropOriginalMixin
+
+_VALID_AGGREGATIONS = frozenset(
+    ["min", "max", "mean", "median", "mode", "sum", "count"],
+)
+
+ListOfAggregations = Annotated[
+    List[Literal["min", "max", "mean", "median", "mode", "sum", "count"]],
```
liamholmes31: S: think you can just type hint as list here, since you validate the entries in the Is part anyway

liamholmes31: would suggest setting up like this:
```
class AggregationOptions(str, Enum):
    MAX = "max"
    ...

ListOfAggregations = Annotated[
    List,
    Is[
        lambda list_value: all(
            entry in AggregationOptions._value2member_map_ for entry in list_value
        )
    ],
]
```

tubular/aggregations.py
```diff
@@ -0,0 +1,182 @@
...
+
+    @beartype
+    def __init__(
+        self,
+        columns: list[str],
+        aggregations: ListOfAggregations,
+        drop_original: bool = False,
+        verbose: bool = False,
+    ) -> None:
+        if not columns:
```
liamholmes31: S: if you call super().__init__() it will handle columns and verbose for you, so could delete the checking and setting here

tubular/aggregations.py
```diff
@@ -0,0 +1,182 @@
...
+            msg = "Columns list cannot be empty."
+            raise ValueError(msg)
+
+        self.columns = columns
+        self.aggregations = aggregations
+        self.drop_original = drop_original
```
liamholmes31: S: don't need to set this here, as you call the mixin method that handles this

tubular/aggregations.py
```diff
@@ -0,0 +1,182 @@
...
+    ----------
+    key : str
+        Column used for grouping during aggregation.
+    """
+
+    @beartype
+    def __init__(
+        self,
+        columns: list[str],
+        aggregations: list[str],
```
liamholmes31: S: can use the same type hint here

tubular/aggregations.py
```diff
@@ -0,0 +1,182 @@
...
+    @nw.narwhalify
+    def transform(
+        self,
+        df: FrameT,
+    ) -> FrameT:
+        """Transforms the dataframe by aggregating rows over specified columns.
+
+        Parameters
+        ----------
+        df : FrameT
```
liamholmes31: would put 'df: pl/pd.DataFrame', as this is more easily understood for readers (same for other parts of docstring)

tubular/aggregations.py
```diff
@@ -0,0 +1,182 @@
...
+                getattr(nw.col(col_name), agg)().alias(f"{col_name}_{agg}")
+                for col_name, aggs in aggregation_dict.items()
+                for agg in aggs
+            ),
+        )
+
+        # Merge the aggregated results back with the original DataFrame
+        df = df.join(grouped_df, on=self.key, how="left")
+
+        df = df.with_columns(
```
liamholmes31: Q: is this left here from debugging?

tests/conftest.py
```diff
@@ -286,6 +286,17 @@ def minimal_attribute_dict():
             "new_column_name": "c",
             "pd_method_name": "add",
         },
+        "BaseAggregationTransformer": {
+            "columns": ["col1", "col2"],
```
liamholmes31: S: it will probably make things easier if these follow the same 'a', 'b' test column naming convention as the other transformers in this file

tests/conftest.py
```diff
@@ -286,6 +286,17 @@ def minimal_attribute_dict():
...
+            "aggregations": ["min", "max"],
+            "drop_original": False,
+        },
+        "AggregateRowOverColumnsTransformer": {
+            "columns": ["col1", "col2"],
+            "aggregations": ["min", "max"],
+            "key": "key_column",
```
liamholmes31: similarly would use e.g. 'c'

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,116 @@
+import pytest
+
+from tests import utils as u
+from tubular.aggregations import AggregateRowOverColumnsTransformer
+
+
+class BaseTransformerTest:
+    """Base test class for transformer initializations."""
+
+    def setup_transformer(self, columns, aggregations, key, drop_original=False):
```
liamholmes31: N: would say we don't need this method, you should be able to accomplish the same with the 'minimal_attribute_dict' and 'uninitialized_transformers' fixtures

liamholmes31: think this whole class could be deleted actually

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,116 @@
...
+        """Utility method to initialize a transformer."""
+        return AggregateRowOverColumnsTransformer(
+            columns,
+            aggregations,
+            key,
+            drop_original,
+        )
+
+
+class TestAggregateRowOverColumnsTransformerInit(BaseTransformerTest):
```
liamholmes31: needs to inherit TestBaseAggregationTransformerInit

liamholmes31: also needs this method:
```
@classmethod
    def setup_class(cls):
        cls.transformer_name = "AggregateRowOverColumnsTransformer"
```

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,116 @@
...
+
+        transformer = self.setup_transformer(columns, aggregations, key)
+        with pytest.raises(
+            ValueError,
+            match=f"key '{key}' not found in dataframe columns",
+        ):
+            transformer.transform(df)
+
+
+class TestAggregateRowOverColumnsTransformerMethodsTransform(BaseTransformerTest):
```
liamholmes31: S: needs to inherit 'GenericTransformTests' from base_tests.py, and also needs the setup_class method mentioned above

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,116 @@
...
+        ],
+    )
+    def test_transform(self, library, aggregations, drop_original, expected_data):
+        """Test transform method aggregates rows correctly."""
+        columns = ["col1", "col2"]
+        key = "group_key"
+
+        # Create a pandas DataFrame first
+        df_dict = {
+            "col1": [1, 2, 3, 4],
```
liamholmes31: S: is often a good idea to include some nulls to test that they don't break things

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,116 @@
...
+        df = u.dataframe_init_dispatch(df_dict, library)
+
+        transformer = self.setup_transformer(columns, aggregations, key, drop_original)
+        transformed_df = transformer.transform(df)
+
+        # Create expected DataFrame using the library parameter
+        expected_df = u.dataframe_init_dispatch(expected_data, library)
+
+        # Compare the transformed DataFrame with the expected DataFrame using the dispatch function
+        u.assert_frame_equal_dispatch(transformed_df, expected_df)
```
liamholmes31: S: things can often break for single rows, so it's worth including a test that looks at this scenario too

tests/aggregations/test_baseaggregationtransformer.py
```diff
@@ -0,0 +1,60 @@
...
+        cls.transformer_name = "BaseAggregationTransformer"
+
+    def test_invalid_aggregation_error(self):
+        """Test that an error is raised for invalid aggregation methods."""
+        columns = ["col1", "col2"]
+        invalid_aggregations = ["invalid", "max"]
+        with pytest.raises(BeartypeCallHintParamViolation):
+            BaseAggregationTransformer(columns, invalid_aggregations)
+
+    def test_columns_empty_list_error(
```
liamholmes31: S: can delete this, will be inherited from ColumStrListInitTests

tests/aggregations/test_baseaggregationtransformer.py
```diff
@@ -0,0 +1,60 @@
...
+        minimal_attribute_dict,
+        uninitialized_transformers,
+    ):
+        args = minimal_attribute_dict[self.transformer_name].copy()
+        args["columns"] = []
+        with pytest.raises(ValueError):
+            uninitialized_transformers[self.transformer_name](**args)
+
+    @pytest.mark.parametrize("drop_orginal_column", (0, "a", ["a"], {"a": 10}, None))
+    def test_drop_column_arg_errors(
```
liamholmes31: S: can delete, is inherited from DropOriginalInitMixinTests

tubular/aggregations.py
```diff
@@ -0,0 +1,182 @@
...
+            raise ValueError(msg)
+
+        self.columns = columns
+        self.aggregations = aggregations
+        self.drop_original = drop_original
+        self.verbose = verbose
+
+        self.set_drop_original_column(drop_original)
+
+        if self.verbose:
```
liamholmes31: N: I would remove this for now, I think we'll add more verbose functionality to BaseTransformer at a later date so will be handled there

tests/aggregations/test_baseaggregationtransformer.py
```diff
@@ -0,0 +1,60 @@
...
+
+class TestBaseAggregationTransformerCreateNewColNames:
+    """Tests for methods in BaseAggregationTransformer."""
+
+    def test_create_new_col_names(self):
+        """Test create_new_col_names method returns correct column names."""
+        columns = ["col1", "col2"]
+        aggregations = ["min", "max", "mean"]
+        transformer = BaseAggregationTransformer(columns, aggregations)
+        prefix = "agg"
```
liamholmes31: N: this is a bit confusing as in the code this would be a column name, e.g. 'col1', so suggest matching that here

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,116 @@
...
+            key,
+            drop_original,
+        )
+
+
+class TestAggregateRowOverColumnsTransformerInit(BaseTransformerTest):
+    """Tests for AggregateRowOverColumnsTransformer initialization."""
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_invalid_key_error(self, library):
```
liamholmes31: N: think this test belongs in the transform tests class

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,116 @@
...
+                ["min", "max", "mean", "median", "sum", "count"],
+                False,
+                {
+                    "col1": [1, 2, 3, 4],
+                    "col2": [5, 6, 7, 8],
+                    "group_key": ["A", "B", "A", "B"],
+                    "col1_min": [1, 2, 1, 2],
+                    "col1_max": [3, 4, 3, 4],
+                    "col1_mean": [2.0, 3.0, 2.0, 3.0],
+                    "col1_median": [2.0, 3.0, 2.0, 3.0],
```
liamholmes31: N: would be a more meaningful test of median if we included another row of test data, so then median!=mean

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,116 @@
...
+        ):
+            transformer.transform(df)
+
+
+class TestAggregateRowOverColumnsTransformerMethodsTransform(BaseTransformerTest):
+    """Tests for methods in AggregateRowOverColumnsTransformer."""
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize(
+        "aggregations, drop_original, expected_data",
```
liamholmes31: S: tests around drop_original will be inherited from DropOriginalTransformTests, so don't need to worry about that here.
I would drop this argument and the second set of values included below

tubular/aggregations.py
hariinduu: Changed

tubular/aggregations.py
hariinduu: Yes, since it inherits from BaseTransformer. Modified.

tubular/aggregations.py
hariinduu: Removed

tubular/aggregations.py
hariinduu: Alright. Removed

tubular/aggregations.py
hariinduu: Done

tubular/aggregations.py
hariinduu: Updated df and returns

tests/aggregations/test_baseaggregationtransformer.py
hariinduu: Removed

tests/aggregations/test_baseaggregationtransformer.py
hariinduu: Even though DropOriginalInitMixinTests covers the 'drop_original' checks, this test is here to make sure to catch BeartypeCallHintParamViolation. If we drop this we will have to edit DropOriginalInitMixinTests to expect BeartypeCallHintParamViolation instead of TypeError for non-boolean 'drop_original' values.

tests/aggregations/test_baseaggregationtransformer.py
hariinduu: Changed and gave column name as prefix. Hope it is clear now

tests/aggregations/test_aggregation_transformers.py
hariinduu: Relocated to the transform tests class

tests/aggregations/test_aggregation_transformers.py
hariinduu: Added this method

tests/aggregations/test_aggregation_transformers.py
hariinduu: Deleted the whole class

tests/aggregations/test_aggregation_transformers.py
hariinduu: Added anthor row

tests/conftest.py
hariinduu: changed

tests/conftest.py
hariinduu: changed

tests/aggregations/test_aggregation_transformers.py
hariinduu: Done

tests/aggregations/test_aggregation_transformers.py
hariinduu: Dropped them

tests/aggregations/test_aggregation_transformers.py
hariinduu: Added a single row test

tests/aggregations/test_aggregation_transformers.py
hariinduu: Added another test with null values

pyproject.toml
hariinduu: Just removed

tubular/aggregations.py
hariinduu: Yes, removed

tests/aggregations/test_baseaggregationtransformer.py
hariinduu: Fixed this with modifications to the conftest.py

tests/aggregations/test_aggregation_transformers.py
hariinduu: Done

tests/aggregations/test_baseaggregationtransformer.py
liamholmes31: Ah OK i'm with you, this is because we haven't converted DropOriginalMixin to beartype yet so what you've done makes sense - could you just add a comment above saying 'NOTE - can delete this test once DropOriginalMixin is converted to beartype'

tests/conftest.py
```diff
@@ -349,6 +357,14 @@ def minimal_dataframe_lookup(request) -> dict[str, pd.DataFrame]:
     for transformer in other_nan_transformers:
         min_df_dict[transformer] = nan_df
 
+    min_df_dict["AggregateRowOverColumnsTransformer"] = pd.DataFrame(
+        {
```
liamholmes31: S: can you add this to test_data as 'create_row_aggregation_df' and import from there (copy the style of the other functions in that file)

tubular/aggregations.py
```diff
@@ -0,0 +1,192 @@
...
+        pd.DataFrame or pl.DataFrame
+            Transformed DataFrame with aggregated columns.
+
+        Raises
+        ------
+        ValueError
+            If the key column is not found in the DataFrame.
+        """
+
+        if df is None or not isinstance(
```
liamholmes31: S: I think this check isn't needed as narwhals will handle it for us

tubular/aggregations.py
```diff
@@ -0,0 +1,192 @@
...
+            df,
+            (pd.DataFrame, pl.DataFrame, nw.DataFrame, nw.LazyFrame),
+        ):
+            msg = f"{self.classname()}: Input must be a pandas, polars, or narwhals DataFrame, got {type(df).__name__}"
+            raise ValueError(msg)
+
+        if df.shape[0] == 0:
```
liamholmes31: S: this check would be accomplished by calling super.transform(X)

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class TestAggregateRowOverColumnsTransformerMethodsTransform(GenericTransformTests):
+    """Tests for methods in AggregateRowOverColumnsTransformer."""
+
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "AggregateRowOverColumnsTransformer"
+
+    def setup_method(self, method):
```
liamholmes31: S: doesn't look like this is used anywhere, can delete

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,242 @@
...
+        df = u.dataframe_init_dispatch(self.df_dict, library)
+
+        # transformer = transformer_setup(columns, aggregations, key, drop_original)
+        transformer = uninitialized_transformers[self.transformer_name](**args)
+        transformed_df = transformer.transform(df)
+
+        # Create expected DataFrame using the library parameter
+        expected_df = u.dataframe_init_dispatch(expected_data, library)
+
+        if library == "polars":
```
liamholmes31: S: can you include a comment here explaining that this is because polars uses more efficient types for count stats (and also in other sections with this logic)

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,242 @@
...
+                    nw.col("a_count").cast(nw.UInt32),
+                    nw.col("b_count").cast(nw.UInt32),
+                ],
+            ).to_native()
+
+        # Compare the transformed DataFrame with the expected DataFrame using the dispatch function
+        u.assert_frame_equal_dispatch(transformed_df, expected_df)
+
+    @pytest.mark.parametrize("invalid_input", [1, True, "a", [], {}, None])
+    def test_non_pd_type_error(self, invalid_input, initialized_transformers):
```
liamholmes31: S: can delete once you add the super.transform(X) call

tubular/aggregations.py
liamholmes31: correction, will be handled if you add a super.transform(X) call

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,242 @@
...
+        transformer = initialized_transformers[self.transformer_name]
+
+        with pytest.raises(
+            ValueError,
+            match=f"{self.transformer_name}: Input must be a pandas, polars, or narwhals DataFrame, got {type(invalid_input).__name__}",
+        ):
+            transformer.transform(invalid_input)
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_single_row(
```
liamholmes31, line 133: S: can delete once you add the super.transform(X) call

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,242 @@
...
+    ):
+        """Test transform method with a single-row DataFrame."""
+        args = copy.deepcopy(minimal_attribute_dict[self.transformer_name])
+        args["columns"] = ["a", "b"]
+        args["aggregations"] = ["min", "max", "mean", "median", "sum", "count"]
+        args["key"] = "c"
+
+        # Create a single-row DataFrame
+        single_row_df_dict = {
+            "a": [1],
```
liamholmes31: S: could you change this value to None, so that a single row None test case is included

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,242 @@
...
+        ):
+            transformer.transform(df)
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize(
+        "aggregations, expected_data",
+        [
+            # Test cases for "min", "max", "mean", "median", "sum", and "count"
+            (
+                ["min", "max", "mean", "median", "sum", "count"],
```
liamholmes31: N: could you include mode in here please, should cover all cases really

liamholmes31: (same for other tests)

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,242 @@
...
+                    "b_max": [9, 5, 9, 5, 9],
+                    "b_mean": [5.0, 4.0, 5.0, 4.0, 5.0],
+                    "b_median": [4.0, 4.0, 4.0, 4.0, 4.0],
+                    "b_sum": [15, 8, 15, 8, 15],
+                    "b_count": [3, 2, 3, 2, 3],
+                },
+            ),
+        ],
+    )
+    def test_transform(
```
liamholmes31: N: would call something like 'test_transform_basic_case_outputs'

tubular/aggregations.py
```diff
@@ -0,0 +1,192 @@
...
+        prefix : str
+            Prefix to use for generating new column names.
+
+        Returns
+        -------
+        list[str]
+            List of new column names with the specified prefix and aggregation type.
+        """
+        return [f"{prefix}_{agg}" for agg in self.aggregations]
+
```
liamholmes31, line 94: S: we might need some extra logic here to handle categorical columns, in a transform method:

```
import narwhals.selectors as ncs

class CategoricalAggregationOptions(str, Enum):

    MODE = "mode"
    COUNT = "count"

def transform(self, X):

non_numeric_columns = list(
            set(self.columns).difference(set(X.select(ncs.numeric()).columns)),
        )

numeric_aggs_requested=any(agg not in CategoricalAggregationOptions.values for agg in self.aggregations)

if not len(non_numeric_columns)==0 and numeric_aggs_requested:
    raise TypeError(..)

return super.transform(X)

```

and would also want to add a test for this error to a new TestBaseAggregationTransformerTransform test class (this class would inherit from GenericTransformTests)

tests/aggregations/test_aggregation_transformers.py
```diff
@@ -0,0 +1,242 @@
...
+
+class TestAggregateRowOverColumnsTransformerInit(TestBaseAggregationTransformerInit):
+    """Tests for AggregateRowOverColumnsTransformer initialization."""
+
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "AggregateRowOverColumnsTransformer"
+
+
+class TestAggregateRowOverColumnsTransformerMethodsTransform(GenericTransformTests):
```
liamholmes31: once you've created TestBaseAggregationTransformerTransform, this class will now inherit from that instead

liamholmes31: S: could we add one more test case, looking at categorical variables for mode and count please

liamholmes31: actually would be good if this test included 3 columns for:

- categorical
- string
- bool

tubular/aggregations.py
liamholmes31: thinking about it, I think this method should also error if non-numeric/cat/bool types are passed along with cat agg methods, so like:
```
cat_columns = set(self.columns).intersection(
            X.select(ncs.categorical()).columns,
        )
bool_columns = set(self.columns).intersection(X.select(ncs.boolean()).columns)
str_columns = set(self.columns).intersection(X.select(ncs.string()).columns)
numeric_columns=X.select(ncs.numeric()).columns

non_categorical_agg_friendly_columns=set(self.columns).difference(cat_columns).difference(bool_columns).difference(str_columns).difference(numeric_columns)
if not numeric_aggs_requested and len(non_categorical_agg_friendly_columns)>0:
    raise TypeError(...)
```

tubular/aggregations.py
```diff
@@ -0,0 +1,192 @@
...
+    columns : list[str]
+        Columns to apply the transformations to.
+    aggregations : list[str]
+        Aggregation methods to apply.
+    drop_original : bool
+        Indicator for dropping original columns.
+    verbose : bool
+        Indicator for verbose output.
+    """
+
```
liamholmes31, line 62: can you add here `polars_compatible=True` 

some tests are dependent on this class attr to run their polars versions!

tubular/aggregations.py
```diff
@@ -0,0 +1,192 @@
...
+        Column name to group by for aggregation.
+    drop_original : bool, optional
+        Whether to drop the original columns after transformation. Default is False.
+
+    Attributes
+    ----------
+    key : str
+        Column used for grouping during aggregation.
+    """
+
```
liamholmes31, line 156: can you add here `polars_compatible=True` 

some tests are dependent on this class attr to run their polars versions!

tests/aggregations/test_aggregation_transformers.py
hariinduu: I think this method is required as it initializes self.df_dict, which is used to set up test data across multiple test methods

tests/aggregations/test_aggregation_transformers.py
hariinduu: Renamed

tests/aggregations/test_baseaggregationtransformer.py
hariinduu: Added a note

tests/aggregations/test_aggregation_transformers.py
hariinduu: This revealed a scenario where 'sum' will come as Null if we have only 'None' values to consider. Therefore added a seperate part for the transformer to handle them and make it zero if the sum is Null. Or just thinking whether to keep it as null and modify the the test.

tubular/aggregations.py
hariinduu: Added

tubular/aggregations.py
hariinduu: Added

tests/aggregations/test_aggregation_transformers.py
hariinduu: Deleted

tubular/aggregations.py
hariinduu: Added super.transform(X) and deleted the test

tests/aggregations/test_aggregation_transformers.py
hariinduu: Added comments

tubular/aggregations.py
hariinduu: Yes, this check is performed in BaseTransformer. Therefore deleted

tubular/aggregations.py
```diff
@@ -0,0 +1,276 @@
...
+        aggregations: ListOfAggregations,
+        drop_original: bool = False,
+        verbose: bool = False,
+    ) -> None:
+        super().__init__(columns=columns, verbose=verbose)
+
+        self.aggregations = aggregations
+
+        self.set_drop_original_column(drop_original)
+        self.polars_compatible = True
```
liamholmes31: this should be a class attr, can see an example here https://github.com/azukds/tubular/blob/24a5ac33857628fc7e4692565b8b32acce8dac46/tubular/capping.py#L20

(applies to the other one too)

tubular/aggregations.py
```diff
@@ -0,0 +1,276 @@
...
+        # print("Numeric columns in DataFrame:", numeric_columns_in_df)
+        # print("Non-numeric columns identified:", non_numeric_columns)
+        # print("Numeric aggregations requested:", numeric_aggs_requested)
+
+        if len(non_numeric_columns) > 0 and numeric_aggs_requested:
+            raise TypeError(
+                "Numeric aggregation methods requested for non-numeric columns.",
+            )
+
+        if not numeric_aggs_requested and len(non_categorical_agg_friendly_columns) > 0:
```
liamholmes31: S: this checking should go in a .transform method of the base class

tubular/aggregations.py
hariinduu: Added as a class attribute

tubular/aggregations.py
hariinduu: Added as a class attribute

tubular/aggregations.py
hariinduu: Added as a class attribute

tubular/aggregations.py
hariinduu: with the mode, the issue is that it can return multiple values so doesn't count as an aggregation - we would need a .first method that narwhals doesn't have, so we decided to remove the option for categorical columns

tests/aggregations/test_aggregation_transformers.py
hariinduu: Regarding the mode option the issue is that it can return multiple values so doesn't count as an aggregation - we would need a .first method that narwhals doesn't have, so we removed this option

tubular/aggregations.py
hariinduu: Removed that part as we got rid of categorical aggregations

tests/aggregations/test_aggregation_transformers.py
liamholmes31: I think sum as null makes sense to me here! i'll edit that while you're off :)

tests/aggregations/test_aggregation_transformers.py
liamholmes31: have decided to drop categorical support, closing

tests/aggregations/test_aggregation_transformers.py
liamholmes31: ah OK this did expose what I think is a narwhals bug where sum of nulls is 0 for pandas and none for polars, think both should be none.. to get this over the line, will remove sum as an option and we can look to reintroduce later!

tests/conftest.py
liamholmes31: actioned https://github.com/azukds/tubular/pull/424/commits/b6f75a6a59f66308e40cbbc302b9a0356fe849b3

tests/aggregations/test_aggregation_transformers.py
liamholmes31: removed here https://github.com/azukds/tubular/pull/424/commits

tests/aggregations/test_aggregation_transformers.py
liamholmes31: removed here https://github.com/azukds/tubular/pull/424/commits/eecbb610c1e35524f526ae9da1603a8300e79074

tests/aggregations/test_aggregation_transformers.py
liamholmes31: actioned here https://github.com/azukds/tubular/pull/424/commits/1d217c095769fa3f8414cbdad814973407f54604

tubular/aggregations.py
liamholmes31: added base transform checks here https://github.com/azukds/tubular/pull/424/commits/1d217c095769fa3f8414cbdad814973407f54604

tests/aggregations/test_aggregation_transformers.py
liamholmes31: issue here https://github.com/azukds/tubular/issues/438

nesa1995 (approved):

PR #426, opened 2025-06-24
Feature/nw dtinfoextractor

liamholmes31:
narwhalified datetime info extractor https://github.com/azukds/tubular/issues/378

### DIFF
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -142,7 +142,7 @@ class DropOriginalInitMixinTests:
         args["drop_original"] = drop_orginal_column
 
         with pytest.raises(
-            TypeError,
+            BeartypeCallHintParamViolation,
             match=f"{self.transformer_name}: drop_original should be bool",
         ):
             uninitialized_transformers[self.transformer_name](**args)
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -5,12 +5,13 @@ from __future__ import annotations
 import datetime
 import warnings
 import zoneinfo
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional, Union
 
 import narwhals as nw
 import narwhals.selectors as ncs
 import numpy as np
 import pandas as pd
+from beartype import beartype
 
 from tubular.base import BaseTransformer
 from tubular.mixins import DropOriginalMixin, NewColumnNameMixin, TwoColumnMixin
@@ -1004,7 +1005,7 @@ class DatetimeInfoExtractor(BaseDatetimeTransformer):
 
     """
 
-    polars_compatible = False
+    polars_compatible = True
 
     TIME_OF_DAY = "timeofday"
     TIME_OF_MONTH = "timeofmonth"
@@ -1056,27 +1057,20 @@ class DatetimeInfoExtractor(BaseDatetimeTransformer):
         DAY_OF_WEEK: "weekday",
     }
 
+    @beartype
     def __init__(
         self,
-        columns: str | list[str],
-        include: str | list[str] | None = None,
-        datetime_mappings: dict[str,] | None = None,
+        columns: Union[str, list[str]],
+        include: Optional[Union[str, list[str]]] = None,
+        datetime_mappings: Optional[dict[str, list[int]]] = None,
         drop_original: bool = False,
         **kwargs: dict[str, bool],
     ) -> None:
         if include is None:
             include = self.INCLUDE_OPTIONS
-        else:
-            if type(include) is not list:
-                msg = f"{self.classname()}: include should be List"
-                raise TypeError(msg)
 
         if datetime_mappings is None:
             datetime_mappings = {}
-        else:
-            if type(datetime_mappings) is not dict:
-                msg = f"{self.classname()}: datetime_mappings should be Dict"
-                raise TypeError(msg)
 
         super().__init__(
             columns=columns,
@@ -1091,10 +1085,7 @@ class DatetimeInfoExtractor(BaseDatetimeTransformer):
                 raise ValueError(msg)
 
         if datetime_mappings != {}:
-            for key, mapping in datetime_mappings.items():
-                if type(mapping) is not dict:
-                    msg = f"{self.classname()}: values in datetime_mappings should be dict"
-                    raise TypeError(msg)
+            for key, _ in datetime_mappings.items():
                 if key not in include:
                     msg = f"{self.classname()}: keys in datetime_mappings should be in include"
                     raise ValueError(msg)
@@ -1170,30 +1161,46 @@ class DatetimeInfoExtractor(BaseDatetimeTransformer):
         msg = f"{self.classname()}: value for {include_option} mapping in self._map_values should be an integer value in {min(self.RANGE_TO_MAP[include_option])}-{max(self.RANGE_TO_MAP[include_option])}"
         raise ValueError(msg)
 
-    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+    @nw.narwhalify
+    def transform(self, X: FrameT) -> FrameT:
         """Transform - Extracts new features from datetime variables.
 
         Parameters
         ----------
-        X : pd.DataFrame
+        X : pd/pl.DataFrame
             Data with columns to extract info from.
 
         Returns
         -------
-        X : pd.DataFrame
+        X : pd/pl.DataFrame
             Transformed input X with added columns of extracted information.
         """
-        X = super().transform(X)
+        X = nw.from_native(super().transform(X))
+        backend = nw.get_native_namespace(X)
 
         for col in self.columns:
             for include_option in self.include:
-                X[col + "_" + include_option] = getattr(
-                    X[col].dt,
-                    self.DATETIME_ATTR[include_option],
-                ).apply(
-                    self._map_values,
-                    include_option=include_option,
+                X = X.with_columns(
+                    getattr(
+                        nw.col(col).dt,
+                        self.DATETIME_ATTR[include_option],
+                    )().map_batches(
+                        lambda s: nw.new_series(
+                            name=col,
+                            values=[self._map_values(v, include_option) for v in s],
+                            backend=backend,
+                        ).alias(
+                            col + "_" + include_option,
+                        ),
+                    ),
                 )
+                # X[col + "_" + include_option] = getattr(
+                #     X[col].dt,
+                #     self.DATETIME_ATTR[include_option],
+                # ).apply(
+                #     self._map_values,
+                #     include_option=include_option,
+                # )
 
         # Drop original columns if self.drop_original is True
         return DropOriginalMixin.drop_original_column(
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -98,11 +98,12 @@ class BaseGenericDateTransformer(
         """
 
         type_dict = {}
-        datetime_type = "datetime64"
-        date_type = "date32[pyarrow]"
-        allowed_types = [datetime_type]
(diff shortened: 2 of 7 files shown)


tests/base_tests.py
```diff
@@ -142,7 +142,7 @@ def test_drop_column_arg_errors(
         args["drop_original"] = drop_orginal_column
 
         with pytest.raises(
-            TypeError,
+            BeartypeCallHintParamViolation,
             match=f"{self.transformer_name}: drop_original should be bool",
```
davidhopkinson26: Q: surprised this is passing with the match specified like this since beartype raises a different error? any idea what is going on here?

tubular/dates.py
```diff
@@ -982,79 +1002,78 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
-                msg = f"{self.classname()}: datetime_mappings should be Dict"
-                raise TypeError(msg)
+
+        if datetime_mappings != {}:
+            for key in datetime_mappings:
+                if key not in include:
+                    msg = f"{self.classname()}: keys in datetime_mappings should be in include"
+                    raise ValueError(msg)
+
+        if datetime_mappings is None:
```
davidhopkinson26: duplicated statement from line 1066

tubular/dates.py
```diff
@@ -982,79 +1002,78 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
-                raise TypeError(msg)
 
         if datetime_mappings is None:
             datetime_mappings = {}
-        else:
-            if type(datetime_mappings) is not dict:
-                msg = f"{self.classname()}: datetime_mappings should be Dict"
-                raise TypeError(msg)
+
+        if datetime_mappings != {}:
```
davidhopkinson26: S: Could this be simplified to `if datetime_mappings:` and then an else block to set to {} if it is?

tubular/dates.py
```diff
@@ -982,79 +1002,78 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
+            for key in datetime_mappings:
+                if key not in include:
+                    msg = f"{self.classname()}: keys in datetime_mappings should be in include"
+                    raise ValueError(msg)
+
+        if datetime_mappings is None:
+            datetime_mappings = {}
```
davidhopkinson26: should this be setting to the default values?

tubular/dates.py
```diff
@@ -1063,26 +1082,37 @@ def __init__(
...
+        self.mapping_transformers = {
+            include_option: MappingTransformer(
+                mappings={
+                    col + "_" + include_option: self.inverted_datetime_mappings[
+                        include_option
+                    ]
+                    for col in self.columns
+                },
+                return_dtypes={
+                    col + "_" + include_option: "Categorical" for col in self.columns
```
davidhopkinson26: do we want the return of mappings to be categorical?  Would have expected integer here based on the default mapping values?  though I guess a user supplied map could include categoricals.

tests/base_tests.py
davidhopkinson26: ah, this test is in fact failing! think we just need to validate that we have a BeartypeCallHintParamViolation as verifying the exact message leaves us subject to tests failing if they make small changes to the message

tubular/dates.py
liamholmes31: looks like this happens inside _process_provided_mappings.

I suppose a user might just provide custom mappings for one of the options, so the code records mappings_provided separately and then either grabs these or looks up the defaults later on.

Could probably be streamlined.. and an open question as to what extent we actually need to support custom mappings

tests/base_tests.py
liamholmes31: thanks, looks like I actually left that in by accident so have reverted here https://github.com/azukds/tubular/pull/426/commits/b0eea4a684fa752d96723a4d05dbba0f575392cf

tubular/dates.py
liamholmes31: thanks, removed here https://github.com/azukds/tubular/pull/426/commits/63ef88072fb8f6528318e2ef70ad833fda4db104

tubular/dates.py
liamholmes31: thanks, updated here https://github.com/azukds/tubular/pull/426/commits/0694ea3ce198190be5688f763c0d43a59c4e5b20

I have made further changes to how mapping are handled, but will link in another comment

tubular/dates.py
liamholmes31: have done a bit of a refactor of how mappings are handled that I think makes more sense here https://github.com/azukds/tubular/pull/426/commits/f5cb9f9dd0050cddc28a023e932a5af310c7f324

basically the old code was having to bend over backwards to invert the user provided mappings, and I think this was just because I didn't know about the unpack operator when I originally wrote this years ago, hopefully it's all a cleaner setup now

tubular/dates.py
liamholmes31: hopefully the new mapping setup makes it clearer that we are returning cat/enum types!
https://github.com/azukds/tubular/pull/426/commits/f5cb9f9dd0050cddc28a023e932a5af310c7f324

liamholmes31:
ready for review

tubular/dates.py
```diff
@@ -1168,124 +1180,159 @@ def __init__(
...
+                if not pd.isna(val)
+            }
+            for col in self.columns
+            for include_option in self.include
+        }
+
+        # set up list of paired condition/outcome tuples for mapping
+        conditions_and_outcomes = {
+            col + "_" + include_option: [
+                self.mapping_transformer._create_mapping_conditions_and_outcomes(
```
cjmwills: S: given this is a static method, should we instead call it directly from the class, rather than the instance? And same for `_combine_mappings_into_expression`

```suggestion
                MappingTransformer._create_mapping_conditions_and_outcomes(...)
```

tubular/dates.py
```diff
@@ -1007,6 +1011,29 @@ def transform(self, X: FrameT) -> FrameT:
...
+    list,
+    Is[
+        lambda list_value: all(
+            entry in DatetimeInfoOptions._value2member_map_ for entry in list_value
+        )
+    ],
+]
+
+
 class DatetimeInfoExtractor(BaseDatetimeTransformer):
```
cjmwills, line 1037: S: looks like parts of this docstring will need updating, particularly the `dayofweek: 0-6` reference which has changed now

tubular/dates.py
```diff
@@ -2,21 +2,25 @@
```
cjmwills, line 2: N: nothing to do with this PR, but the docstring on line 1 references the wrong module!

tubular/dates.py
```diff
@@ -1007,6 +1011,29 @@ def transform(self, X: FrameT) -> FrameT:
...
+    Is[lambda s: s in DatetimeInfoOptions._value2member_map_],
+]
+DatetimeInfoOptionList = Annotated[
+    list,
+    Is[
+        lambda list_value: all(
+            entry in DatetimeInfoOptions._value2member_map_ for entry in list_value
+        )
+    ],
+]
```
cjmwills, line 1034: P: this is very well structured, not seen some of these functions before!

tubular/dates.py
```diff
@@ -1087,79 +1114,64 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
+    @beartype
     def __init__(
         self,
-        columns: str | list[str],
-        include: str | list[str] | None = None,
-        datetime_mappings: dict[str,] | None = None,
-        drop_original: bool = False,
+        columns: Union[str, list[str]],
+        include: Optional[DatetimeInfoOptionList] = None,
+        datetime_mappings: Optional[dict[DatetimeInfoOptionStr, dict[int, str]]] = None,
```
cjmwills, line 1147: Q: I can't seem to run an example where I supply `datetime_mappings`, is that the same for you? The beartype validation requires the mapping dict to be `dict[int, str]` so it fails validation (and that isn't the format I'd expect it to be passed in as). This is using the `datetime_mappings` in the `examples/` folder so that will need updating either way

CHANGELOG.rst
```diff
@@ -22,8 +22,10 @@ Each individual change should have a link to the pull request after the descript
 Changed
 ^^^^^^^
 
+- feat: narwhalified DatetimeInfoExtractor `#378 <https://github.com/azukds/tubular/issues/378>_`
+- placeholder 
 - feat: optimisations for MeanResponseTransformer, further optimisations for  MappingTransformer `#451 <https://github.com/azukds/tubular/issues/451>_`
-- feat: optimisation chnages to DatetimeSinusoidCalculator, added 'return_native_override' argument to DatetimeSinusoidCalculator, reduced with_columns being called many times. https://github.com/azukds/tubular/issues/465
+- feat: optimisation changes to DatetimeSinusoidCalculator, added 'return_native_override' argument to DatetimeSinusoidCalculator, reduced with_columns being called many times. https://github.com/azukds/tubular/issues/465
 - placeholder
```
cjmwills: Q: are the versions in this changelog correct? Didn't we release 1.4.8 recently?

tubular/dates.py
cjmwills: And looking at the tests, the only ones that reference `datetime_mappings` are checking it errors correctly, rather than it gives the expected output. Will be worth adding some for that if so.

CHANGELOG.rst
liamholmes31: fixed in merge conflicts here :) https://github.com/azukds/tubular/pull/426/commits/4f64167c78367d10e3c8917c0f9d74e468707472

tubular/dates.py
liamholmes31: this call has now been removed, in this commit https://github.com/azukds/tubular/pull/426/commits/6f4b9754fd2904437b41029a0e2d54b0ef7312cc

tubular/dates.py
liamholmes31: good spot, have added a test for this so should all be working now https://github.com/azukds/tubular/pull/426/commits/28e2e58b681632a600bd8d4edfaf59d8afae3a6c

tubular/dates.py
liamholmes31: thanks, updated here https://github.com/azukds/tubular/pull/426/commits/ebc66bc1a1dd35c40e0927f58776832660ebe110

tubular/dates.py
liamholmes31: updated here https://github.com/azukds/tubular/pull/426/commits/d33a05976e6667bebb50a3ed78850ccfa05724ed

tubular/dates.py
cjmwills: Thanks the test looks good! The typehint still seems too restrictive though, as beartype requires the dict in `datetime_mappings` to be in form `dict[int, str]`, yet the docstring says you can pass an iterable which will then be unpacked by `_process_provided_mappings`

cjmwills: Which means the example in the docstring fails `datetime_mappings = {"timeofday": {"am": range(0, 12), "pm": range(12, 24)}}`

tubular/dates.py
cjmwills: That's a neat way to replace three blocks of code!

tubular/dates.py
liamholmes31: ah thanks, the docstring was stale again here, have updated https://github.com/azukds/tubular/pull/426/commits/86a4e8ddce9615530aad02e6316765b2bc255f2b

cjmwills: Thanks! We'll need to update the notebook in `examples/` but can do that in a separate PR

PR #428, opened 2025-07-02
Feature/nw_datetimesinusoidtransformer

Boluwatife28:


uv.lock
cjmwills, line 1: Q: was this supposed to be committed?

tubular/dates.py
```diff
@@ -1382,8 +1383,19 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
             for method in self.method:
                 new_column_name = f"{method}_{desired_period}_{desired_units}_{column}"
 
-                X[new_column_name] = getattr(np, method)(
-                    column_in_desired_unit * (2.0 * np.pi / desired_period),
+                # Calculate the sine or cosine of the column in the desired unit
+                X = X.with_columns(
```
cjmwills, line 1387: P: nice use of narwhals syntax to get the code block updated!

tests/dates/test_DatetimeSinusoidCalculator.py
```diff
@@ -236,148 +237,262 @@ def setup_class(cls):
...
+                expected_row,
+            )
 
-    def test_expected_output_both_methods(self):
-        expected = d.create_datediff_test_df()
-        columns = ["a", "b"]
+    @pytest.mark.parametrize(
+        "columns, method, units, period",
+        [
+            (["a", "b"], ["sin", "cos"], "month", 12),
```
cjmwills: S: given this test is for both methods on single column, should we remove the parameter that uses cols "a" and "b" at the same time?

tubular/dates.py
Boluwatife28: Thanks :)

uv.lock
Boluwatife28: No, it wasn't. I huave removed it. :)

tests/dates/test_DatetimeSinusoidCalculator.py
Boluwatife28: ah good catch, I have modified it

cjmwills (approved):

cjmwills (approved):

Issue #244, opened 2024-05-01
Consolidate DateDiffLeapYearTransormer and DateDifferenceTransformer


davidhopkinson26:
**What**

Consolidate the DateDiffLeapYearTransormer and DateDifferenceTransformer into a single transformer with options to cover the functionality of both.

**Why**

It would be more user friendly if these two transformers which appear to offer similar functionality were consolidated.

**How**

These two transformers are semantically very similar but function in different ways which for now necessitates different logic.

DateDiffLeapYearTransormer returns integer year date differences (i.e. age) factoring in leap years.

DateDifferenceTransformer calculates difference between datetimes in days/ hours/ minutes/ seconds.  M and Y were removed as options in #123 due to [M and Y values not being supported by np.timedeltas](https://stackoverflow.com/questions/54794431/adding-one-month-to-datetime64-with-timedelta)

To resolve this issue we need to consider ways of bringing the logic of these together in a way which can support  years/ months/ days/ hours/ minutes/ seconds.  We need to be able to account for calendar months and leap years.  Ideally we would not have forking logic for this and would use a single logical framework for all.

Currently unclear how best to approach this so contributions and ideas are welcome!

Issue #282, opened 2024-07-19
Drop py38 support, pin all requirements.txt for reproducibilty


adamsardar:
**What** This is a bookmark to remind us to perform a few cleanup tasks when python38 goes out of support in October.

**Why** See [this comment](https://github.com/lvgig/tubular/pull/271#discussion_r1684384519) on PR #271 - py38 support prevents us pinning dependencies across all GA python versions:

> [scipy 1.10.1](https://pypi.org/project/scipy/1.10.1/) (released in Feb'23) supports 3.8<=python<3.12 and [scipy 1.11.1](https://pypi.org/project/scipy/1.11.1/) (released Jun'23) supports 3.9<=python<3.13. There is therefore no way to cover all of the python versions with a single pinned requirements.txt.

Python 3.8 is due to go out of support soon, so it seems timely to restore pinned dependencies for explicit reproducible of build environments.

**How** Wait for python 3.8 to go into end of life ([October 31st 2024](https://endoflife.date/python)) - or close to it - and then:
- Drop CI stages that test against py38 
  - consider adding against 3.13, which should [be available by then](https://peps.python.org/pep-0719/)
- Change the devcontainer python version
- pin `requirements.txt` - it should be possible to use package versions that work across py39 through py313 - and use it as part of the CI pipeline
- update pyproject.toml

adamsardar:
I've been thinking about this ticket. I don't think that we should use `requirements.txt` for CI - that should be a direct install from `pyproject.toml` to simulate 'real' package installations in different python versions (which will have no knowledge of `requirements.txt` from pypi). In a `pip-tools` workflow, `requirements.txt` only exists for reproducibility reasons, not for listing dependencies. For us this means that it is only used for defining the devcontainer.

JohnStokes228:
the [tubular pypi](https://pypi.org/project/tubular/) page only shows its compatibility up to python 3.8, which does not match to your [pyproject.toml](https://github.com/lvgig/tubular/blob/main/pyproject.toml), which shows you test at >=3.8

could this be updated to reflect the actual supported python versions as part of this issue?

davidhopkinson26:
> the [tubular pypi](https://pypi.org/project/tubular/) page only shows its compatibility up to python 3.8, which does not match to your [pyproject.toml](https://github.com/lvgig/tubular/blob/main/pyproject.toml), which shows you test at >=3.8
> 
> could this be updated to reflect the actual supported python versions as part of this issue?

[...]

PR #429, opened 2025-07-03
Feature/consolidate datediffleap and datediff transformers

kajol-patel:
Added options week, fortnight, lunar_month, common_year and custom_days in DateDifferenceTransformer. Added a deprecated message for DateDiffLeapYearTransormer. Created a test test_expected_output_units and a test fixture for it.

https://github.com/azukds/tubular/issues/244

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ dependencies = [
     "narwhals >= 1.31.0",
     "polars >= 1.9.0",
     "beartype >= 0.19.0",
+    "typing-extensions>=4.5.0",
 ]
 requires-python = ">=3.9"
 authors = [{ name = "LV GI Data Science Team", email="#DataSciencePackages@lv.co.uk"}]
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -76,7 +76,9 @@ threadpoolctl==3.6.0
 tomli==2.2.1
     # via pytest
 typing-extensions==4.14.0
-    # via exceptiongroup
+    # via
+    #   tubular (pyproject.toml)
+    #   exceptiongroup
 tzdata==2025.2
     # via pandas
 virtualenv==20.31.2


pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "narwhals >= 1.31.0",
     "polars >= 1.9.0",
     "beartype >= 0.19.0",
+    "deprecated>=1.2.18",
```
davidhopkinson26: Q: genuine question - do we want to introduce a dependency just to make deprecation warnings? Trying to keep dependencies of tubular quite light to avoid conflicts for users.  What benefits do we get over just using the native warnings.warn https://docs.python.org/3/library/warnings.html

uv.lock
davidhopkinson26, line 1: I think we should probably make a choice whether to use uv lock vs pip compile -> requirements-dev.txt and stick to one.  Might make sense to switch but should probably be out of scope for this PR?  Would you mind raising a separate issue/small PR for this if it is something you want to propose please?

tubular/dates.py
```diff
@@ -429,14 +433,15 @@ class DateDifferenceTransformer(BaseDateTwoColumnTransformer):
         Name given to calculated datediff column. If None then {column_upper}_{column_lower}_datediff_{units}
         will be used.
     units : str, default = 'D'
-        Numpy datetime units, accepted values are 'D', 'h', 'm', 's'
+        Accepted values are "week", "fortnight", "lunar_month", "common_year", "custom_days", 'D', 'h', 'm', 's'
```
davidhopkinson26, line 441: Q: Not sure I understand the intent here, why are we expanding functionality at the same time as deprecating the transformer?

davidhopkinson26 (commented):
Thanks for the contribution Kajol! 

Not really a full review here but I have asked some questions below to help clarify the intent.  Suspect I am just out of the loop on a couple of offline discussions.  A PR description would also probably help clear things up.

tubular/dates.py
davidhopkinson26: my bad - misread the PR due to some collapsed lines!

pyproject.toml
adamsardar: 100% we should use the python base deprecated system from [the PEP](https://peps.python.org/pep-0702/).

uv.lock
adamsardar: Yeah, agreed. Consider adding `uv.lock` to .gitignore until we take a stab at it. Perhaps as part of #282 (which might as well say 3.9 now!)

tubular/dates.py
```diff
@@ -495,11 +517,48 @@ def transform(self, X: FrameT) -> FrameT:
...
+        start_date_col = nw.col(self.columns[0])
+        end_date_col = nw.col(self.columns[1])
+
+        # truncating time for specific units
+        if self.units in UNITS_TO_TRUNCATE_TIME_FOR:
+            start_date_col = start_date_col.dt.truncate("1d")
+            end_date_col = end_date_col.dt.truncate("1d")
+
+        if self.units == "custom_days":
+            timedelta_value, timedelta_format = self.custom_days_divider, "D"
```
adamsardar, line 559: S: currently it's pretty easy to pass in a non-integer value into this field. Consider using the `@beartype` decorator so as to check types? We have already used it successfully in the package - [example](https://github.com/azukds/tubular/blob/23da87d60e63778df78a5e14d6cd2b2f0813baf3/tubular/mapping.py#L68).

tests/dates/test_DateDifferenceTransformer.py
```diff
@@ -198,13 +198,13 @@ def expected_df_3(library="pandas"):
             ),
         ],
         "D": [
-            9713.042372685186,
-            7219.957627314815,
+            9713.0,
```
adamsardar: P: dropping the long list of decimals is well worth doing

tests/dates/test_DateDifferenceTransformer.py
```diff
@@ -893,6 +893,841 @@ def create_datediff_test_nulls_df2(library="pandas"):
...
+
+    return dataframe_init_dispatch(df_dict, library=library)
+
+
+def expected_df_10(library="pandas"):
+    """Expected output for test_expected_output_units_fortnight."""
+
+    df_dict = {
+        "a": [
+            datetime.datetime(
```
adamsardar: Q: Are all these datetime examples the same, differing only by the units? In which case, consider a text fixture?

uv.lock
kajol-patel: Done [here](https://github.com/azukds/tubular/pull/429/commits/6d6dcc9ee4d9b9984a0c6ed7afd3fadbfd1d22d1)

pyproject.toml
kajol-patel: Fixed [here](https://github.com/azukds/tubular/pull/429/commits/822de36f4172be32b2c9a4ed67075f357e249e6a)
Removed the "deprecated" package. Used DeprecationWarning from warnings to display a deprecated message instead, also added a deprecated message to the docstring. 
Could not use the warnings.deprecated decorator from [PEP 702](https://peps.python.org/pep-0702/), because it was [added](https://docs.python.org/3.13/library/warnings.html#warnings.deprecated) in python version 3.13 and tubular supports older versions as well (>=3.9).

tests/dates/test_DateDifferenceTransformer.py
kajol-patel: Done [here](https://github.com/azukds/tubular/pull/429/commits/26842573ea8233d1d3ad4309b2a17434cd0edbf1)

tubular/dates.py
kajol-patel: Created separate issue for beartyping [#440](https://github.com/azukds/tubular/issues/440)

pyproject.toml
adamsardar: You're right that `@deprecated` was added into py313 (see [PEP-702](https://peps.python.org/pep-0702/)), but there is also a backport within the [typing-extensions package](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.deprecated) (since v 4.5.0).

I know that it sounds pedantic, but could you use that decorator please? The reason being that type checkers and IDEs can help the user via the decorator, but a `DeprecationWarning` from inside the method won't.

tests/dates/test_DateDifferenceTransformer.py
adamsardar: Much easier to read and more concise.

pyproject.toml
kajol-patel: Done [here](https://github.com/azukds/tubular/pull/429/commits/8d3a44d08c2fd71a425fd0a4e6977e14c3b652c0) and added typng-extensions in dependencies [here](https://github.com/azukds/tubular/pull/429/commits/2fe9f8afeec2eb1ff5a663fe3c2a2b815f6f0dc6)

adamsardar: Nice. We have tests passing from py39 through py313 now.

adamsardar (approved):
Nice feature addition. Thanks!

Issue #418, opened 2025-05-14
Beartype remaining init methods


davidhopkinson26:
following on from https://github.com/azukds/tubular/issues/399

Add beartype decorator to init methods for all transformers not already using it so it can handle type checking for us.

Doing so would allow us to remove/ reduce some mixins and mixin methods:

- NewColumnNameMixin
- SeparatorColumnMixin
- TwoColumnMixin (?)
- check_and_set_weight method
- set_drop_original_column method

We could drop the Mixin classes for the transformers but arguably still want the test mixins for these bits of type handling so should probably keep those and adapt the tests to expect a `BeartypeCallHintParamViolation` with no specific message. See #417

liamholmes31:
open PR for DropOriginalMixin https://github.com/azukds/tubular/issues/418

have also been working on the other three, will PR those after this first one

PR #430, opened 2025-07-04
Feature/beartype drop original

liamholmes31:
beartyped droporiginalmixin and associated transformers

As some transformers use multiple mixins this did mean having to overload more tests, but thought this was preferable to beartyping everything in one go

https://github.com/azukds/tubular/issues/418

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,7 @@ Each individual change should have a link to the pull request after the descript
 ------------------
 
 - feat: - feat: converted DropOriginalMixin and associated transformers to beartype
+- feat: added `get_feature_names_out` method to all transformers, which gives back list of modified/created features. This follows the sklearn API.
 
 2.0.0 (16/10/2025)
 ------------------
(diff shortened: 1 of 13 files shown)


tubular/_types.py
```diff
@@ -0,0 +1,37 @@
+from typing import Annotated, Union
+
+from beartype.vale import Is
+
+# needed as by default beartype will just randomly sample to type check elements
```
davidhopkinson26, line 36: P: think I saw that this might not be needed in future so good job separating out the complex logic so it is easy to cull in future.

tubular/mixins.py
```diff
@@ -54,28 +54,12 @@ def classname(self) -> str:
...
-        """
-        # check if 'drop_original' argument is boolean
-        if type(drop_original) is not bool:
-            msg = f"{self.classname()}: drop_original should be bool"
-            raise TypeError(msg)
-
-        self.drop_original = drop_original
-
     @nw.narwhalify
     def drop_original_column(
```
davidhopkinson26, line 75: Q: did you mean to beartype this method too?

davidhopkinson26 (dismissed):
Hi Liam, might be missing something but seems like you forgot to actually beartype the drop_original_column method?

tubular/mixins.py
liamholmes31: ah thanks - so we hit the issue we've been avoiding here that beartype doesn't like FrameT.

I'm not really sure why this method doesn't just use the attrs rather than asking for things to be repeated as args, what do you think of switching to just use attrs?

liamholmes31: some recent changes in main have unblocked this  one, so have now actioned https://github.com/azukds/tubular/pull/430/commits/623aff75edaaf2d3349e4a3e39a54d3a59085500

tests/comparison/test_EqualityChecker.py
```diff
@@ -24,6 +26,27 @@ class TestInit(
     def setup_class(cls):
         cls.transformer_name = "EqualityChecker"
 
+    # overload until we beartype the new_column_name mixin
```
hariinduu, line 29: Do we have a ticket for this?

liamholmes31: have linked this comment on the new column name draft PR so i remember :) https://github.com/azukds/tubular/pull/493

hariinduu (approved):

hariinduu (approved):

PR #441, opened 2025-07-23
Feature/column aggregator

nesa1995:
Pr for https://github.com/azukds/tubular/issues/385

### DIFF
--- a/tests/aggregations/test_BaseAggregationTransformer.py
+++ b/tests/aggregations/test_BaseAggregationTransformer.py
@@ -47,27 +47,6 @@ class TestBaseAggregationTransformerInit(
         ):  # Adjust to expect BeartypeCallHintParamViolation
             uninitialized_transformers[self.transformer_name](**args)
 
-
-class TestBaseAggregationTransformerCreateNewColNames:
-    """Tests for methods in BaseAggregationTransformer."""
-
-    def test_create_new_col_names(self):
-        """Test create_new_col_names method returns correct column names."""
-        columns = ["a", "b"]
-        aggregations = ["min", "max", "mean"]
-        transformer = BaseAggregationTransformer(columns, aggregations)
-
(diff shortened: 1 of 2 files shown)


tubular/aggregations.py
```diff
@@ -241,3 +244,88 @@ def transform(
...
+            drop_original=drop_original,
+            verbose=verbose,
+        )
+
+    @beartype
+    def transform(
+        self,
+        X: DataFrame,
+    ) -> DataFrame:
+        """Transforms the dataframe by aggregating rows over specified columns.
```
hariinduu: This should be corrected as 'Transforms the dataframe by aggregating specified columns across each row'

tests/aggregations/test_BaseAggregationTransformer.py
```diff
@@ -48,26 +47,6 @@ def test_drop_column_arg_errors(
             uninitialized_transformers[self.transformer_name](**args)
 
 
-class TestBaseAggregationTransformerCreateNewColNames:
```
hariinduu, line 51: Is below covered somewhere else?
TestBaseAggregationTransformerCreateNewColNames

tubular/aggregations.py
liamholmes31: thanks, updated docstrings here https://github.com/azukds/tubular/pull/441/commits/3db9714934ac64c69b3ce674fb61d54be94cc0c0

tests/aggregations/test_BaseAggregationTransformer.py
liamholmes31: it looks like this method was removed when it was realised that the naming conventions would need to be different between the two transformers and it wouldn't be reusable

I think it is still covered by the output tests, as these check that column names are matching expectation (as well as values)

PR #442, opened 2025-07-24
User warnings silenced and tests

alphanumericmale:
Have made use of the inherited attribute 'verbose' from the BaseImputer class in order to silence the large number of Mapping Imputer User warnings experienced from missing levels in date series. 

The Default Value is False, which may need reviewing, as this also silences the warning for missing mapping dictionaries which share the same logic and warning.

### DIFF
--- a/tubular/mapping.py
+++ b/tubular/mapping.py
@@ -340,17 +340,18 @@ class MappingTransformer(BaseMappingTransformer, BaseMappingTransformMixin):
             values_to_be_mapped = set(self.mappings[col].keys())
             values_in_df = set(X.get_column(col).unique())
 
-            if len(values_to_be_mapped.intersection(values_in_df)) == 0:
-                warnings.warn(
-                    f"{self.classname()}: No values from mapping for {col} exist in dataframe.",
-                    stacklevel=2,
-                )
+            if self.verbose:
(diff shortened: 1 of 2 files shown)


tubular/mapping.py
```diff
@@ -340,17 +340,18 @@ def transform(
             values_to_be_mapped = set(self.mappings[col].keys())
             values_in_df = set(X.get_column(col).unique())
 
-            if len(values_to_be_mapped.intersection(values_in_df)) == 0:
-                warnings.warn(
-                    f"{self.classname()}: No values from mapping for {col} exist in dataframe.",
-                    stacklevel=2,
-                )
+            if self.verbose:
```
hariinduu, line 343: S: Should we ensure that verbose has a default value set in the class? (e.g., self.verbose = False). This will make sure that warnings are suppressed by default unless explicitly enabled

bayesandconfused: It's quite tricky to track back but this class inherits the verbose attribute from BaseTransformer and it is set to True there. Currently this is the only use for that verbose attribute but if it expands we can separate out the functionality a bit.

bayesandconfused: apologies ignore above it is set to False as default not True

hariinduu (approved):

PR #443, opened 2025-07-25
speed optimisations for base transformer and imputers file

liamholmes31:
PR addressing live performance issues, key issues were:

- layered narwhalify calls becoming impactful in live setting
- lots of use of methods which instantiate objects or copies, e.g. with_columns, select, schema
- copies being create in BaseTranformer
- moved fit/transform type checking to beartype

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,14 +16,6 @@ Subsections for each version can be one of the following;
 
 Each individual change should have a link to the pull request after the description of the change.
 
-1.4.6 (04/07/2025)
-------------------
-
-Changed
-^^^^^^^
-- Added deprecated warning for DateDiffLeapYearTransformer `#244 <https://github.com/azukds/tubular/issues/244>`
-- Added new units 'week', 'fornight', 'lunar_month', 'common_year' and 'custom_days' to DateDifferenceTransformer. The time component will be truncated for these units and for unit 'D'.
-
 1.4.5 (unreleased))
 ------------------
 
@@ -32,6 +24,15 @@ Changed
 - bugfix: updated env to make package importable, added basic test for this
 - feat: added BaseAggregationTransformer and AggregateRowOverColumnsTransformer classes in new aggregations module
 - narwhalified DatetimeSinusoidCalculator '#425 <https://github.com/azukds/tubular/issues/425>_' 
+- Added deprecated warning for DateDiffLeapYearTransformer `#244 <https://github.com/azukds/tubular/issues/244>`
+- Added new units 'week', 'fornight', 'lunar_month', 'common_year' and 'custom_days' to DateDifferenceTransformer. The time component will be truncated for these units and for unit 'D'.
+- feat: optimisation changes to BaseTransfomer and imputers file. Edited to reduce number of copies and type changes
+from to/from_native calls, and select/with_columns being called many times.
+- feat: added 'return_native' argument to BaseTransfomer to control whether native or narwhals types are returned,
+and limit type changes. Idea is for this to be rolled out across transformers.
+- feat: made creation of copies in BaseTransfomer optional, and default to False.
+- placeholder
+- placeholder
 - placeholder
 
 1.4.4 (24/06/2025)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,6 +116,7 @@ ignore = [
     "B905", # the below two rules seem to make changes that error for python3.9
     "UP038", 
     "PYI041", # clashes with beartype
+    "TCH001", # clashes with beartype
     ]
 
 # Allow autofix for all enabled rules (when `--fix`) is provided.
@@ -131,7 +132,11 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 # Ignore `E402` (import violations) in all `__init__.py` file.
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["E402", "F401"]
-"tests/*" = ["ANN", "S101"]
+"tests/*" = [
+    "ANN",
+    "S101",
+    "PLC2701" # complains about private imports, which we test
+    ]
 
 [tool.ruff.lint.pyupgrade]
 keep-runtime-typing=true
--- a/tests/imputers/test_ArbitraryImputer.py
+++ b/tests/imputers/test_ArbitraryImputer.py
@@ -12,6 +12,7 @@ from tests.base_tests import (
     GenericFitTests,
     GenericTransformTests,
     OtherBaseBehaviourTests,
+    ReturnNativeTests,
 )
 from tests.imputers.test_BaseImputer import GenericImputerTransformTests
 from tubular.imputers import ArbitraryImputer
@@ -47,7 +48,11 @@ class TestFit(GenericFitTests):
         cls.transformer_name = "ArbitraryImputer"
 
 
-class TestTransform(GenericImputerTransformTests, GenericTransformTests):
+class TestTransform(
+    GenericImputerTransformTests,
+    GenericTransformTests,
+    ReturnNativeTests,
+):
     """Tests for transformer.transform."""
 
     @classmethod
@@ -145,6 +150,13 @@ class TestTransform(GenericImputerTransformTests, GenericTransformTests):
         df_transformed_nw = nw.from_native(df_transformed_native)
 
         expected_dtype = df_nw[column].dtype
+
+        native_namespace = nw.get_native_namespace(df_nw).__name__
+
+        # for pandas categorical are converted to enum
+        if col_type == "Categorical" and native_namespace == "pandas":
+            expected_dtype = nw.Enum
+
         actual_dtype = df_transformed_nw[column].dtype
         assert (
             actual_dtype == expected_dtype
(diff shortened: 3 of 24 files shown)


tubular/mapping.py
```diff
@@ -148,8 +149,9 @@ class attribute, indicates whether transformer has been converted to polars/pand
 
     polars_compatible = True
 
+    @beartype
```
liamholmes31, line 152: Hadn't intended to touch this file, but tests got angry when BaseTransformer was moved to beartype, so had to do the same here

tubular/dates.py
```diff
@@ -471,7 +471,7 @@ def __init__(
             "m",
             "s",
         ] = "D",
-        copy: bool | None = None,
+        copy: bool = False,
```
liamholmes31, line 474: Removed None option for copy in BaseTransformer, so had to update other references

tubular/types.py
```diff
@@ -0,0 +1,19 @@
+from typing import Union
+
+import narwhals as nw
+import pandas as pd
+import polars as pl
+
+DataFrame = Union[
```
liamholmes31, line 7: enables beartype (which does not like FrameT)

tubular/_utils.py
```diff
@@ -1,10 +1,53 @@
...
+
+    if not isinstance(X, nw.DataFrame):
+        X = nw.from_native(X)
+
+    return X
+
+
+@beartype
+def _narwhalify_y_if_needed(y: Optional[Series] = None) -> Optional[nw.Series]:
+    """narwhalifies series, if series is not already narwhals
```
nesa1995, line 35: and wondering about the naming, shall we rename to something like _convert_df_to_narwhals and _convert_series_to_narwhals?

tests/utils/test_narwhalify_X_if_needed.py
```diff
@@ -0,0 +1,49 @@
...
+    df = dataframe_init_dispatch(dataframe_dict=df_dict, library=library)
+
+    df = nw.from_native(df)
+
+    output = _narwhalify_X_if_needed(df)
+
+    assert isinstance(
+        output,
+        nw.DataFrame,
+    ), "df has not been converted to narwhals as expected"
```
nesa1995: Q: should we be testing that a call has been made to from_native instead as this test isn't useful? We already know it is a narwhals df

tests/utils/test_narwhalify_y_if_needed.py
```diff
@@ -0,0 +1,65 @@
...
+
+    series = df["a"]
+
+    series = nw.from_native(series, allow_series=True)
+
+    output = _narwhalify_y_if_needed(series)
+
+    native_namespace = nw.get_native_namespace(series).__name__
+
+    assert isinstance(
```
nesa1995: Similar to previous comment

tubular/base.py
```diff
@@ -33,12 +34,15 @@ class BaseTransformer(TransformerMixin, BaseEstimator):
         Columns to apply the transformer to. If a str is passed this is put into a list. Value passed
         in columns is saved in the columns attribute on the object.
 
-    copy : bool, default = True
+    copy : bool, default = False
```
nesa1995, line 41: Q: the default was true before? so guessing we were making copies potentially without realising?

tests/utils/test_narwhalify_X_if_needed.py
liamholmes31: hmm, so in my mind the output that's being  tested here is 'does nothing to narwhals frames' right? e.g. the test would fail if the frame was edited in some way, so I think it is testing  something valid (if slightly trivial!)

tubular/base.py
liamholmes31: yeah exactly. I can imagine this could possibly be useful in some analysis contexts, but in live I don't think this is ever desired behaviour so have changed to default as false

tests/utils/test_narwhalify_X_if_needed.py
nesa1995: I guess I was thinking we want to check that we don't execute that line of code but I suppose ensuring we get the output as expected would also make sure it hasn't been edited in an unexpected way.

CHANGELOG.rst
```diff
@@ -32,6 +24,15 @@ Changed
...
 - narwhalified DatetimeSinusoidCalculator '#425 <https://github.com/azukds/tubular/issues/425>_' 
+- Added deprecated warning for DateDiffLeapYearTransformer `#244 <https://github.com/azukds/tubular/issues/244>`
+- Added new units 'week', 'fornight', 'lunar_month', 'common_year' and 'custom_days' to DateDifferenceTransformer. The time component will be truncated for these units and for unit 'D'.
+- feat: optimisation changes to BaseTransfomer and imputers file. Edited to reduce number of copies and type changes
+from to/from_native calls, and select/with_columns being called many times.
+- feat: added 'return_native' argument to BaseTransfomer to control whether native or narwhals types are returned,
+and limit type changes. Idea is for this to be rolled out across transformers.
+- feat: made creation of copies in BaseTransfomer optional, and default to False.
+- placeholder
+- placeholder
```
nesa1995, line 33: N: Might need to be formatted as not in bullet points and have "- placeholder - placeholder.. "

pyproject.toml
```diff
@@ -131,7 +132,11 @@ dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 # Ignore `E402` (import violations) in all `__init__.py` file.
 [tool.ruff.lint.per-file-ignores]
 "__init__.py" = ["E402", "F401"]
-"tests/*" = ["ANN", "S101"]
+"tests/*" = [
+    "ANN",
+    "S101",
+    "PLC2701" # complains about private imports, which we test
```
nesa1995, line 138: Q: Should we make those public functions instead?

tests/imputers/test_ArbitraryImputer.py
```diff
@@ -263,6 +282,64 @@ def test_polars_unknown_type_output(self, impute_value, impute_val_type):
...
+    # have to overload this one, as has slightly different categorical type handling
+    @pytest.mark.parametrize(
+        ("library", "expected_df_3", "impute_values_dict"),
+        [
+            ("pandas", "pandas", {"b": "g", "c": "f"}),
+            ("polars", "polars", {"b": "g", "c": "f"}),
+        ],
+        indirect=["expected_df_3"],
+    )
+    def test_expected_output_3(
```
nesa1995: N: Not sure about having a test named in this way as not very descriptive

tests/imputers/test_BaseImputer.py
```diff
@@ -277,7 +268,7 @@ def test_imputation_with_falsey_values(
         impute_value,
         expected,
     ):
-        """Test that transform is giving the expected output when applied to object and categorical columns."""
+        """Test that transform is giving the expected output when imputation value is falsey."""
```
nesa1995, line 281: N: small typo of "false"

nesa1995 (changes requested):

tubular/base.py
nesa1995: Makes sense, thanks :)

tests/imputers/test_BaseImputer.py
liamholmes31: this is actually intentional :) https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-how-is-it-different-from-true-and-false

tests/utils/test_narwhalify_X_if_needed.py
liamholmes31: yes exactly! I think this is more in line with the testing outputs not implementation approach

CHANGELOG.rst
liamholmes31: why is this sorry?

tubular/_utils.py
liamholmes31: renamed :) also added the extra util we discussed https://github.com/azukds/tubular/pull/443/commits/a7b5d403ddf1184ea8db630e25edd598923a80be

tests/imputers/test_ArbitraryImputer.py
liamholmes31: fair point, renamed :) https://github.com/azukds/tubular/pull/443/commits/a890859f1dc480a04cf7a3906bee7c50dc54ce96

pyproject.toml
liamholmes31: hmm, I think they do fit the bill of private functions, as we would only really expect developers to work with them?

This toml change only turns off the lint in the tests file, which is OK I think?

tests/imputers/test_ArbitraryImputer.py
liamholmes31: had to rename some others too because of overloading https://github.com/azukds/tubular/pull/443/commits/c830b7c8732016ec6c715097cfdec24103e25a27

CHANGELOG.rst
nesa1995: This is currently how it looks: https://github.com/azukds/tubular/blob/feature/opt_arb_imp/CHANGELOG.rst#:~:text=bugfix%3A%20updated%20env,placeholder%20%2D%20placeholder%20%2D%20placeholder

pyproject.toml
nesa1995: Happy to leave as is :) would have thought it might be okay since we have the function in _utils.py

tests/imputers/test_ArbitraryImputer.py
nesa1995: this looks great! definitely more helpful

CHANGELOG.rst
liamholmes31: ah thanks, should be fixed now! https://github.com/azukds/tubular/pull/443/commits/bc67bfa65b6ccc1ce631c8bf4721f3f18044edbd

Issue #446, opened 2025-07-25
optimise GroupRareLevelsTransformer.transform


liamholmes31:
What?
Things slowing us down are:

- use of X.select and ncs in type checks in _check_str_like_columns, better to look for types in X.schema
    - make sure to only call X.schema once throughout transform and reuse, as this operation is expensive
- nested narwhalify calls, remove decorator and replace with _narwhalify_X_if_needed function as in linked PR 
- add beartype where needed and use new DataFrame type to replace FrameT, as in linked PR
- we want to minimise .select calls and .with_columns calls, preferably to just 1 at the end
    - in transform there are 3 calls in the main logic, two casts and the rare grouping
    - it's actually even worse, because these are inside a for loop
    - We want to combine all of this into one expression and a single call, see the linked PR for an example
- similar applies to the .select in the for_nulls check, try to do this with a single expression and no for loop

https://github.com/azukds/tubular/pull/443 

Why?
optimise transformer

PR #452, opened 2025-07-31
Speed optimisations for rare levels transformer

hariinduu:
PR addressing below issues in GroupRareLevelsTransformer.transform method

1. use of X.select and ncs in type checks in _check_str_like_columns, better to look for types in X.schema. make sure to only call X.schema once throughout transform and reuse, as this operation is expensive
2. Nested narwhalify calls, remove decorator and replace with _narwhalify_X_if_needed function as in linked PR
3. Add beartype where needed and use new DataFrame type to replace FrameT, as in linked PR
4. We want to minimise .select calls and .with_columns calls, preferably to just 1 at the end
- in transform there are 3 calls in the main logic, two casts and the rare grouping
- it's actually even worse, because these are inside a for loop
- We want to combine all of this into one expression and a single call, see the linked PR for an example
5. similar applies to the .select in the for_nulls check, try to do this with a single expression and no for loop

#446

### DIFF
--- a/tubular/nominal.py
+++ b/tubular/nominal.py
@@ -364,8 +364,8 @@ class GroupRareLevelsTransformer(BaseTransformer, WeightColumnMixin):
             msg = f"{self.classname()}: transformer must run on str-like columns, but got non str-like {non_str_like_columns}"
             raise TypeError(msg)
 
-    @nw.narwhalify
-    def _check_for_nulls(self, X: FrameT) -> None:
+    @beartype
+    def _check_for_nulls(self, X: DataFrame) -> None:
         """check that transformer being called on only non-null columns.
 
         Note, found including nulls to be quite complicated due to:
@@ -383,6 +383,8 @@ class GroupRareLevelsTransformer(BaseTransformer, WeightColumnMixin):
             Data to transform
 
         """
+
+        X = _convert_dataframe_to_narwhals(X)
 
         null_check_expressions = {
             col: nw.col(col).is_null().sum().alias(col) for col in self.columns


tubular/nominal.py
```diff
@@ -501,11 +502,16 @@ def transform(self, X: FrameT) -> FrameT:
             Transformed input X with rare levels grouped for into a new rare level.
 
         """
-        X = nw.from_native(BaseTransformer.transform(self, X))
+        # X = nw.from_native(BaseTransformer.transform(self, X))
```
liamholmes31: could you change this to:
```
X=BaseTransformer.transform(self, X, return_native_override=False)
```

tubular/nominal.py
```diff
@@ -522,34 +528,34 @@ def transform(self, X: FrameT) -> FrameT:
...
-                X = X.with_columns(nw.col(c).cast(nw.Categorical))
+        transform_expressions = {
+            c: transform_expressions[c].cast(schema[c])
+            if (schema[c] in [nw.Categorical, nw.Enum])
+            else transform_expressions[c]
+            for c in self.columns
+        }
 
-        return X
+        return X.with_columns(**transform_expressions)
```
liamholmes31: could you change this to:
```
X=X.with_columns(**transform_expressions)
return _return_narwhals_or_native_dataframe(X, self.return_native)
```

tests/nominal/test_GroupRareLevelsTransformer.py
```diff
@@ -290,6 +294,10 @@ def test_expected_output_weight(self, library):
 
         df_transformed = x.transform(df)
 
+        # Ensure both data frames are of the same type
+        df_transformed = nw.from_native(df_transformed).to_native()
```
liamholmes31: S: think these from_native calls won't be needed after the other changes (this also applies throughout this file)

tubular/nominal.py
```diff
@@ -501,11 +502,16 @@ def transform(self, X: FrameT) -> FrameT:
...
 
         """
-        X = nw.from_native(BaseTransformer.transform(self, X))
+        # X = nw.from_native(BaseTransformer.transform(self, X))
+        X = _narwhalify_X_if_needed(X)
 
-        self._check_str_like_columns(
-            X.with_columns(nw.col(col) for col in self.columns),
-        )
+        if X.shape[0] == 0:
```
liamholmes31: this is handle by BaseTransformer so can delete here :)

tubular/nominal.py
```diff
@@ -501,11 +502,16 @@ def transform(self, X: FrameT) -> FrameT:
             Transformed input X with rare levels grouped for into a new rare level.
 
         """
-        X = nw.from_native(BaseTransformer.transform(self, X))
+        # X = nw.from_native(BaseTransformer.transform(self, X))
+        X = _narwhalify_X_if_needed(X)
```
liamholmes31: need to update this to 
```
X=_convert_dataframe_to_narwhals(X)
```

tubular/nominal.py
hariinduu: Done

tubular/nominal.py
hariinduu: Added BaseTransformer above and deleted from here

tubular/nominal.py
hariinduu: Imported from _utils and changed here

tubular/nominal.py
hariinduu: Updated

tests/nominal/test_GroupRareLevelsTransformer.py
hariinduu: Removed from couple of places. Still it is there in few places as some of the commands won't work without it. 
For example .with_columns

tests/nominal/test_GroupRareLevelsTransformer.py
```diff
@@ -265,6 +265,9 @@ def test_expected_output_no_weight(self, library):
 
         df_transformed = x.transform(df)
 
+        # Ensure both data frames are of the same type
+        expected = nw.from_native(expected).to_native()
```
liamholmes31: N: don't need any from/to native calls here

liamholmes31: removed here https://github.com/azukds/tubular/pull/452/commits/f7858d1cd2bc13a199309d0051f9eee2cfb3a07b

nesa1995 (approved):

nesa1995 (approved):

Issue #445, opened 2025-07-25
optimise datedifference transformer


liamholmes31:
Optimisations for DateDifferenceTransformer and base date classes

Key issues are:

- lots of nested narwhalify calls
- inefficient type checking, calls .select often rather than .schema once

liamholmes31:
https://github.com/azukds/tubular/pull/454

PR #454, opened 2025-08-01
Feature/optimise datediff

liamholmes31:
Main changes are removing layered narwhals conversions, and making column checks more efficient by working with schema instead of directly with data

https://github.com/azukds/tubular/issues/445

### DIFF
--- a/tests/dates/test_BaseGenericDateTransformer.py
+++ b/tests/dates/test_BaseGenericDateTransformer.py
@@ -1,9 +1,11 @@
 import copy
+import datetime
 
 import narwhals as nw
 import numpy as np
 import polars as pl
 import pytest
+from dateutil.tz import gettz
 
 from tests.base_tests import (
     ColumnStrListInitTests,
@@ -15,6 +17,7 @@ from tests.base_tests import (
     ReturnNativeTests,
 )
 from tests.test_data import create_date_diff_different_dtypes
+from tests.utils import dataframe_init_dispatch
 from tubular.dates import TIME_UNITS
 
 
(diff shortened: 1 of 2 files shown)


tubular/dates.py
```diff
@@ -76,10 +86,10 @@ def __init__(
...
 
-    @nw.narwhalify
+    @beartype
     def check_columns_are_date_or_datetime(
         self,
-        X: FrameT,
+        X: DataFrame,
         datetime_only: bool,
     ) -> None:
         """Raise a type error if a column to be operated on is not a datetime.datetime or datetime.date object
```
Ritunjai-Sharma: S: Do we want to add more info to the docstring, especially to reflect the use of schema-related functionality? Maybe something like "This method checks the schema of the DataFrame to ensure that each column listed in `self.columns` is either a datetime or date type, depending on the `datetime_only` flag. If a column does not meet the expected type criteria, a TypeError is raised."

tubular/dates.py
```diff
@@ -95,52 +105,51 @@ def check_columns_are_date_or_datetime(
...
 
-            else:
-                col_dtype = X.get_column(col).dtype
+            elif schema[col] == nw.Date:
+                is_date = True
 
-                msg = f"{self.classname()}: {col} type should be in {allowed_types} but got {col_dtype}"
+            if (not is_datetime) and (not (not datetime_only and is_date)):
+                msg = f"{self.classname()}: {col} type should be in {type_msg} but got {schema[col]}. Note, Datetime columns should have time_unit in {TIME_UNITS} and time_zones from zoneinfo.available_timezones()"
                 raise TypeError(msg)
```
Ritunjai-Sharma, line 131: Q: In this function, we have two separate instances where a TypeError is raised in the code. I'm assuming this is because the first focuses on individual column type correctness, ensuring that each column meets the expected criteria and
the second focuses on overall type consistency, ensuring that all columns collectively have consistent and valid types? If so, can we add a code comment explaining this.

tubular/dates.py
liamholmes31: updated here :) https://github.com/azukds/tubular/pull/454/commits/5cd8d74e332abd719f71edebcd4fb051bd54b725

tubular/dates.py
liamholmes31: added a comment + more to docstring :) https://github.com/azukds/tubular/pull/454/commits/5cd8d74e332abd719f71edebcd4fb051bd54b725

tubular/dates.py
```diff
@@ -4,32 +4,26 @@
...
 
 TIME_UNITS = ["us", "ns", "ms"]
-TIME_ZONES = zoneinfo.available_timezones().union({None})
-
-DATETIME_VARIANTS = [
-    nw.Datetime(time_unit=time_unit, time_zone=time_zone)
-    for time_unit in TIME_UNITS
-    for time_zone in TIME_ZONES
-    # exclude problem/generic timezones
-    if time_zone not in ["Factory", "localtime"]
```
Ritunjai-Sharma, line 31: Since we are removing banned default timezones factory/localtime here, do we want to add a test to check/show that narwhals does error out for these?

liamholmes31: added here :) https://github.com/azukds/tubular/pull/454/commits/e476b6983e798c60f7cb7a55694945fc2b2d2461

Issue #448, opened 2025-07-28
optimise MappingTransformer


liamholmes31:
How?
Causes of issues seem to be:
- many col dtype lookups rather than single schema lookup
- layered to/from_native calls
- many with_columns calls inside loops, should prepare expressions fully and then call minimal times
- repeated calculation of column unique values, should store as var and reuse

PR #458, opened 2025-08-04
Feature/opt map imp

liamholmes31:
What?
- removed layered to/from native calls for transform
- moved to expression-first approach to avoid materialising data many times

https://github.com/azukds/tubular/issues/448

### DIFF
--- a/tests/mapping/test_MappingTransformer.py
+++ b/tests/mapping/test_MappingTransformer.py
@@ -2,6 +2,7 @@ import narwhals as nw
 import pytest
 
 import tests.test_data as d
+from tests.base_tests import ReturnNativeTests
 from tests.mapping.test_BaseMappingTransformer import (
     BaseMappingTransformerInitTests,
     BaseMappingTransformerTransformTests,
@@ -56,7 +57,7 @@ class TestFit(GenericFitTests):
         cls.transformer_name = "MappingTransformer"
 
 
-class TestTransform(BaseMappingTransformerTransformTests):
+class TestTransform(BaseMappingTransformerTransformTests, ReturnNativeTests):
     """Tests for the transform method on MappingTransformer."""
 
     @classmethod
--- a/tubular/mapping.py
+++ b/tubular/mapping.py
@@ -172,8 +172,8 @@ class BaseMappingTransformMixin(BaseTransformer):
 
     polars_compatible = True
 
+    @staticmethod
     def _create_mapping_conditions_and_outcomes(
-        self,
         col: str,
         key: str,
         mappings: dict[str, dict[str, Union[int, str, bool, float]]],
@@ -215,8 +215,8 @@ class BaseMappingTransformMixin(BaseTransformer):
             )
         )
 
+    @staticmethod
     def _combine_mappings_into_expression(
-        self,
         col: str,
         conditions_and_outcomes: dict[str, tuple[nw.Expr, nw.Expr]],
     ) -> nw.Expr:
--- a/tubular/mapping.py
+++ b/tubular/mapping.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import warnings
 from collections import OrderedDict
-from typing import TYPE_CHECKING, Any, Literal, Optional, Union, Set
+from typing import Any, Literal, Optional, Set, Union
 
 import narwhals as nw
 import numpy as np
@@ -12,12 +12,11 @@ import pandas as pd
 import polars as pl
 from beartype import beartype
 
-from tubular._utils import new_narwhals_series_with_optimal_pandas_types, _narwhalify_X_if_needed
+from tubular._utils import (
+    _narwhalify_X_if_needed,
+)
 from tubular.base import BaseTransformer
 from tubular.types import DataFrame
-
-if TYPE_CHECKING:
-    from narwhals.typing import FrameT
 
 
 class BaseMappingTransformer(BaseTransformer):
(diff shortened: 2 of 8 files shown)


liamholmes31:
Some info from the profiling (ran on 3 columns to get some different types/scenarios in):
- original (non narwhals tubular) - ~0.0088s
- pre optimised narwhals tubular + pandas - ~0.037s
- optimised narwhals tubular + pandas - ~0.022s
- optimised narwhals tubular + polars - ~0.0056s

so we are still taking a hit short term when working with pandas, but hopefully this work helps to keep it manageable until we move to polars

tubular/mapping.py
```diff
@@ -89,31 +91,41 @@ def __init__(
 
             # Assign the mapping to the single null key if it exists
             if len(null_keys) != 0:
-                null_mappings[col] = mappings[col][null_keys[0]]
+                mappings_from_null[col] = mappings[col][null_keys[0]]
 
         self.mappings = mappings
-        self.null_mappings = null_mappings
+        self.mappings_from_null = mappings_from_null
```
cjmwills, line 101: S: can we add this attribute to the docstring? It's not currently mentioned there

tubular/mapping.py
```diff
@@ -77,7 +79,7 @@ def __init__(
             msg = f"{self.classname()}: mappings has no values"
             raise ValueError(msg)
 
-        null_mappings = {col: None for col in mappings}
+        mappings_from_null = {col: None for col in mappings}
         for col in mappings:
```
cjmwills, line 87: Q: this block is designed to handle cases where the mapping key is null, but the use of beartype validation means that can't ever be possible. Won't it reject null keys from the start?

tubular/mapping.py
```diff
@@ -149,9 +172,87 @@ class attribute, indicates whether transformer has been converted to polars/pand
 
     polars_compatible = True
 
+    def _create_mapping_conditions_and_outcomes(
```
cjmwills, line 180: S: could this (and `_combine_mappings_into_expression`) be static methods and then we can remove the reference to self? Doesn't seem to be needed

tubular/mapping.py
```diff
@@ -149,9 +172,87 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
+            else (
+                nw.col(col) == key,
+                nw.lit(mappings[col][key], dtype=dtype),
+            )
+        )
+
+    def _combine_mappings_into_expression(
+        self,
+        col: str,
+        conditions_and_outcomes: list[tuple[nw.Expr, nw.Expr]],
```
cjmwills: Q: should the typehint for this be a dict rather than a list?

tubular/mapping.py
```diff
@@ -149,9 +172,87 @@ class attribute, indicates whether transformer has been converted to polars/pand
...
+        Returns
+        -------
+        nw.Expr: prepared mapping expression
+
+        """
+
+        return reduce(
+            lambda expr, condition_and_outcome: nw.when(condition_and_outcome[0])
+            .then(condition_and_outcome[1])
+            .otherwise(expr),
```
cjmwills, line 251: S: It took me a while to get my head around what was happening here, with narwhals syntax in a lambda function in a reduce function. Potentially my unfamiliarity with narwhals and the abstract nature of narwhals expressions, but some more comments to explain what's happening could be useful (doesn't need to hold up the PR though)

tubular/mapping.py
```diff
@@ -160,113 +261,93 @@ def transform(self, X: DataFrame) -> DataFrame:
...
-        # restore original index for pandas
-        if nw.get_native_namespace(X).__name__ == "pandas":
-            X = nw.to_native(X)
-            X.index = index
-
-        return X[column_order]
+        return _return_narwhals_or_native_dataframe(X, return_native)
 
 
 class MappingTransformer(BaseMappingTransformer, BaseMappingTransformMixin):
```
cjmwills, line 362: Q: is the reference to the use of pd.Series.replace in the docstring still relevant? If not we should update

tests/mapping/test_MappingTransformer.py
cjmwills, line 1: P: some great tests in here that help ensure lots of edge cases are covered!

liamholmes31:
updated description with correct issue, thanks!

tubular/mapping.py
liamholmes31: hm, does it not go through this type hint which it would pass?
```
mappings: dict[str, dict[Any, Any]]
```

tubular/mapping.py
liamholmes31: added here https://github.com/azukds/tubular/pull/458/commits/bc1a648d841a4584311d1c246d7d9fe70bc683c3

tubular/mapping.py
liamholmes31: good call, updated here https://github.com/azukds/tubular/pull/458/commits/57a41972878601f84559a66d4cb2803c4269d0f2

tubular/mapping.py
liamholmes31: updated here https://github.com/azukds/tubular/pull/458/commits/a57abc0c02a145ec1394a3fb384a9f92206a524a

tubular/mapping.py
liamholmes31: thanks updated here https://github.com/azukds/tubular/pull/458/commits/a57abc0c02a145ec1394a3fb384a9f92206a524a

tubular/mapping.py
liamholmes31: added an example comment here https://github.com/azukds/tubular/pull/458/commits/a3a21ca63f409a7faa35ed03f5126823a7c77775

tubular/mapping.py
liamholmes31: thanks updated the docstring here https://github.com/azukds/tubular/pull/458/commits/89c79dfd8c7bdfbdb749c4318eef2837e9fcb5a0

tubular/mapping.py
cjmwills: you're correct, I remember trying this and getting an error, must've been looking at the wrong thing

PR #459, opened 2025-08-04
Optimising BaseNominal

Boluwatife28:


liamholmes31:
looks like you will need to merge in changes from main :) and also some other  tests having the same  error msg issue

tubular/nominal.py
```diff
@@ -50,38 +55,44 @@ def check_mappable_rows(self, X: FrameT) -> None:
...
+        ]
 
-    @nw.narwhalify
-    def transform(self, X: FrameT) -> None:
+        if mappable_rows_count:
+            msg = f"{self.classname()}: nulls would be introduced into columns {', '.join(mappable_rows_count)} from levels not present in mapping"
+            raise ValueError(msg)
+
+    @beartype
+    def transform(self, X: DataFrame) -> DataFrame:
```
liamholmes31: could you add another argument here plz:
`return_native_override: Optional[bool]=None`

this will allow us to turn off return native when this method is being  called within another method, to keep it so that dfs are only  converted to/from native at the beginning and end

liamholmes31 (approved):

Issue #455, opened 2025-08-04
uncap narwhals version


liamholmes31:
What?
the below PR had to cap narwhals due to issues with beartype appearing in 2.0.0, want to debug these to allow us to remove the cap
https://github.com/azukds/tubular/pull/452

Why?
allow us to use latest narwhals

How?
install latest narwhals and debug test issues that appear, once solved can remove the '<' cap in our toml file

dangotbanned:
Hey `narwhals` contributor here 👋 

Just wanted to check, are these (51) failures the new issue?

- https://github.com/narwhals-dev/narwhals/actions/runs/16722924321/job/47331306849?pr=2933

liamholmes31:
Hey! 👋

thanks for reaching out, I gave both of those suggestions a go but sadly no joy. It seemed to be triggering from beartype trying to validate nw.Schema - this was in a private method so for the time being have just turned beartype off there in order to remove the cap on our end!

https://github.com/azukds/tubular/pull/460

dangotbanned:
Oh I see!

I think that can be explained by this PR that landed just before `2.0.0`

- https://github.com/narwhals-dev/narwhals/pull/2894

There's now a forward reference to `DType` in the definition of `Schema`

That *might* be resolved by importing `narwhals.dtypes.DType` instead?

Otherwise, glad to see you've found a workaround 🥳

liamholmes31:
ah thanks, can confirm this one works when imported outside the TYPE_CHECKING block :)

dangotbanned:
> can confirm this one works when imported outside the TYPE_CHECKING block :)

Perfect! happy I could help 🎉

PR #460, opened 2025-08-06
import narwhals DType to enable compatibility with narwhals 2.0.0

liamholmes31:
build pipeline was failing for later narwhals versions due to beartype not liking nw.Schema. import narwhals DType to fix this and uncapped narwhals in toml

https://github.com/azukds/tubular/issues/455


liamholmes31:
update, narwhals devs advised that importing narwhals.typing.DType would also resolve, so have now done this instead

Issue #447, opened 2025-07-28
optimise WeightsColumnMixin


liamholmes31:
How?
the check_weights_column method features many expensive .select calls, would be better to do like:
```
expr_for_null_check=..
expr_for_inf_check=..
...

checks=X.select(expr_for_null_check.alias('null_check'), expr_for_inf_check.alias('inf_check), ...)

null_check, inf_check, ... =checks.row(0)

if null_check:
...

...
```

Code also uses a np.isnan method, which I think now exists in narwhals + can be replaced

liamholmes31:
would start by looking at this PR https://github.com/azukds/tubular/pull/443

adamsardar:
Closed by #461

PR #461, opened 2025-08-06
feature/optimise WeightColumnMixin

kajol-patel:
- Refactored WeightColumnMixin.check_weights_column to combine all weight checks into a single .select call
- Replaced np.isnan usage with narwhals native is_nan

[#447](https://github.com/azukds/tubular/issues/447)

### DIFF
--- a/tubular/mixins.py
+++ b/tubular/mixins.py
@@ -4,7 +4,6 @@ from typing import TYPE_CHECKING, Optional, Union
 
 import narwhals as nw
 import narwhals.selectors as ncs
-import numpy as np
 from beartype import beartype
 
 from tubular._utils import _convert_dataframe_to_narwhals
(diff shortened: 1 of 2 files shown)


tubular/mixins.py
```diff
@@ -178,30 +177,37 @@ def check_weights_column(self, X: FrameT, weights_column: str) -> None:
             msg = f"{self.classname()}: weight column must be numeric."
```
liamholmes31, line 192: think this looks good generally, only thing left is to drop the use of narwhalify/frameT and make use of the new _convert_dataframe_to_narwhals and _return_native_or_narwhals utils + beartype instead

Example can  be found here https://github.com/azukds/tubular/pull/460

adamsardar: Resolved in b31779cf0ccd0ea34a7bb315f8f997d1537ff34e

adamsardar: I had to appease the linter before the saught green tick

Issue #451, opened 2025-07-28
optimise MeanResponseTransformer


liamholmes31:
What?
issues here seem to be from:
- layered to/from_native calls
- this one is already quite good for minimising with_columns calls, but worth seeing if we can go further!

PR #464, opened 2025-08-07
Feature/opt mean response

liamholmes31:
https://github.com/azukds/tubular/issues/451

Have tried to optimise this using the same approach as mapping transformer, but has been less successful here - my theory is because MRE replaces/casts all values then maybe the when/then mapping approach is less efficient. Might be worth trying with something similar to my first attempt on mapping transformer, which used replace_strict to update ALL values (see below)

https://github.com/azukds/tubular/commit/75356b151c69610beb00d12f9f5104debe5a49ee#diff-302f1fdb729b5127ab836f31dd03d3a0194b4b40f66031966b16ec8802215bc9

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,6 @@ venv.bak/
 
 # ruff
 .ruff_cache
+
+# remove when using uv for dependency management
+uv.lock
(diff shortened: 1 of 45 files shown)


hariinduu:
Tried the method mentioned here and couldn't resolve unit tests and unfortunately no much progress. Therefore didn't push anything here.

Issue #463, opened 2025-08-07
optimise check numeric mixin


liamholmes31:
What?
- remove narwhalify calls and replace with new _convert_dataframe_to_narwhals and return_native_or_narwhals utils
- change check to work with data schema rather than X.select, example here https://github.com/azukds/tubular/pull/452

Why?
improve live scoring time

liamholmes31:
https://github.com/azukds/tubular/pull/467#pullrequestreview-3100348701

PR #467, opened 2025-08-08
optimise check numeric mixin

hariinduu:
optimised check numeric mixin by

- removing narwhalify calls and replace with new _convert_dataframe_to_narwhals

- change check to work with data schema rather than X.select calls

#463

### DIFF
--- a/tubular/mixins.py
+++ b/tubular/mixins.py
@@ -5,9 +5,14 @@ from typing import TYPE_CHECKING
 import narwhals as nw
 import narwhals.selectors as ncs
 import numpy as np
+from beartype import beartype
+
+from tubular._utils import _convert_dataframe_to_narwhals
 
 if TYPE_CHECKING:
     from narhwals.typing import FrameT
+
+from tubular.types import DataFrame
 
 
 class CheckNumericMixin:
@@ -21,8 +26,8 @@ class CheckNumericMixin:
 
         return type(self).__name__
 
-    @nw.narwhalify
-    def check_numeric_columns(self, X: FrameT) -> FrameT:
+    @beartype
+    def check_numeric_columns(self, X: DataFrame) -> DataFrame:
         """Helper function for checking column args are numeric for numeric transformers.
 
         Args:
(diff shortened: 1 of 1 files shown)


tubular/mixins.py
```diff
@@ -21,18 +26,37 @@ def classname(self) -> str:
...
 
         """
-        non_numeric_columns = list(
-            set(self.columns).difference(set(X.select(ncs.numeric()).columns)),
-        )
+
+        X = _convert_dataframe_to_narwhals(X)
+        schema = X.schema
+
+        numeric_types = [
```
liamholmes31: S: might be worth putting this list into the types file as NumericTypes, as will probably be used elsewhere too

tubular/mixins.py
```diff
@@ -21,18 +26,37 @@ def classname(self) -> str:
...
+            nw.UInt16,
+            nw.UInt32,
+            nw.UInt64,
+            nw.UInt128,
+        ]
+
+        non_numeric_columns = [
+            col for col in self.columns if schema[col] not in numeric_types
+        ]
+
```
liamholmes31, line 53: also need to edit the return to `return _return_narwhals_or_native(df, return_native)`

tubular/mixins.py
```diff
@@ -21,18 +26,37 @@ def classname(self) -> str:
 
         return type(self).__name__
 
-    @nw.narwhalify
-    def check_numeric_columns(self, X: FrameT) -> FrameT:
+    @beartype
+    def check_numeric_columns(self, X: DataFrame) -> DataFrame:
```
liamholmes31: S: can you add an arg here `return_native_override: Optional[bool]=None`

and then in the body of the function will need
`return_native=self._process_return_native(return_native_override)`

liamholmes31: think you'll then also need to look up where this function is called, and set return_native_override=False there

tubular/mixins.py
hariinduu: Added to types and imported from there

tubular/mixins.py
hariinduu: Done

PR #470, opened 2025-08-19
fixed issue with base date class checking all columns (instead of sel…

liamholmes31:
…f.columns)

also added tests to cover


alphanumericmale (approved):
Good fix and tests.

Issue #465, opened 2025-08-07
optimise datetimesinusoidcalculator


liamholmes31:
What?
the usual things to change here:
- move away from narwhalify in transform and make use of the new utils
- remove from_native calls and make use of return_native_override=False argument when calling base/mixin classes
- move to beartype using new type classes
- change logic to build up a dict of format col:expression inside the for loop, and just call with_columns once at the end

Why?
improve live scoring time

example PR on similar work here https://github.com/azukds/tubular/pull/452

liamholmes31:
https://github.com/azukds/tubular/pull/474

PR #474, opened 2025-08-20
Optimize DatetimeSinusoidCalculator

Boluwatife28:
Optimised DatetimeSinuosoidCalculator by making use of return_native_override=False argument when calling base/mixin classes, using beartype and building col:expression inside the for loop, and just call with_columns once at the end.
Why?
improve live scoring time
https://github.com/azukds/tubular/issues/465

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,7 +37,7 @@ Changed
 - feat: optimisation changes to BaseNominalTransformer, reduced select being called many times, added 'return_native_override' argument.
 - feat: optimisation changes to WeightColumnMixin, combined all weight checks into a single .select call and used narhwals is_nan
 - feat: optimisation chnages to BaseCappingTransformer, added 'return_native_override' argument to BaseCappingTransformer and BaseNumericTransformer.
-- feat: optimisation chnages to DatetimeSinusoidCalculator, added 'return_native_override' argument to DatetimeSinusoidCalculator, reduced with_columns being called many times. https://github.com/azukds/tubular/issues/465
+
 1.4.4 (24/06/2025)
 ------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,7 @@ Changed
 ^^^^^^^
 
 - feat: optimisations for MappingTransformer and BaseMappingTransformerMixin
+- feat: optimisation chnages to DatetimeSinusoidCalculator, added 'return_native_override' argument to DatetimeSinusoidCalculator, reduced with_columns being called many times. https://github.com/azukds/tubular/issues/465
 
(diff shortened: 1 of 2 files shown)


tubular/dates.py
```diff
@@ -1488,7 +1500,7 @@ def transform(self, X: FrameT) -> FrameT:
                 new_column_name = f"{method}_{desired_period}_{desired_units}_{column}"
 
                 # Calculate the sine or cosine of the column in the desired unit
-                X = X.with_columns(
+                expr = (
```
liamholmes31, line 1507: ended up being quite challenging, but found a way to remove another early materialisation!

```
for column in self.columns:

            if not isinstance(self.units, dict):
                desired_units = self.units
            elif isinstance(self.units, dict):
                desired_units = self.units[column]
            if not isinstance(self.period, dict):
                desired_period = self.period
            elif isinstance(self.period, dict):
                desired_period = self.period[column]

            for method in self.method:
                new_column_name = f"{method}_{desired_period}_{desired_units}_{column}"

                # map_batches seems to only evaluate using the final values, e.g.
                # if our for loop uses cos last and we fed it the variable 'method',
                # cos would be used for all calls, so have to be a bit hacky 
                # and fix it down

                # Calculate the sine or cosine of the column in the desired unit
                expr = (
                    (getattr(
                        nw.col(column).dt,
                        desired_units
                        )()*(2*np.pi/desired_period))
                )

                expr=(
                    expr
                    .map_batches(
                        lambda s: getattr(np, "sin")(
                            s.to_numpy()
                        ),
                        return_dtype=nw.Float64,
                    )
                    if method=='sin'
                    else 
                    expr
                    .map_batches(
                        lambda s: getattr(np, "cos")(
                            s.to_numpy()
                        ),
                        return_dtype=nw.Float64,
                    )
                    )
                    
                expr=expr.alias(new_column_name)
                
                exprs[new_column_name] = expr

        X = X.with_columns(**exprs)
```

Boluwatife28: ah that makes sense. Thanks for pointing out! have now implemented this

CHANGELOG.rst
```diff
@@ -33,6 +33,7 @@ Changed
 ^^^^^^^
 
 - feat: optimisations for MappingTransformer and BaseMappingTransformerMixin
+- feat: optimisation chnages to DatetimeSinusoidCalculator, added 'return_native_override' argument to DatetimeSinusoidCalculator, reduced with_columns being called many times. https://github.com/azukds/tubular/issues/465
```
liamholmes31: needs to be under 1.4.7 :)

Boluwatife28: fixed now :)

PR #475, opened 2025-08-20
Feature/optimise map an mre

liamholmes31:
https://github.com/azukds/tubular/issues/451

made standard refactors to MRE (reduce narwhalification calls, move to expression first approach)

also refactored to make use of mapping transformers expression building logic, and improved this logic to only build expressions components that are strictly necessary (for mapping keys that actually appear in transform data)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -47,7 +47,6 @@ Changed
 - feat: optimisation changes to BaseTransfomer and imputers file. Edited to reduce number of copies and type changes from to/from_native calls, and select/with_columns being called many times. `#444 <https://github.com/azukds/tubular/issues/444>_`
 - feat: added 'return_native' argument to BaseTransfomer to control whether native or narwhals types are returned, and limit type changes. Idea is for this to be rolled out across transformers.
 - feat: made creation of copies in BaseTransfomer optional, and default to False.
-- feat: optimisations for MappingTransformer and BaseMappingTransformerMixin
 - feat: optimisations to BaseDatetimeTransformer, BaseDateTransformer, DateDifferenceTransformer, DropOriginalMixin, CheckNumericMixin
 - feat: optimisation changes to BaseNominalTransformer, reduced select being called many times, added 'return_native_override' argument. `#450 <https://github.com/azukds/tubular/issues/450>_``
 - chore: import narwhals.typing.DType for Github in order to uncap narwhals `#455 <https://github.com/azukds/tubular/issues/455>`


liamholmes31:
for mapping profiling, alongside results from last pr (last pr -> this pr) :

original (non narwhals tubular) - ~0.0088s
pre optimised narwhals tubular + pandas - ~0.037s
optimised narwhals tubular + pandas - ~0.022s -> 0.0084
optimised narwhals tubular + polars - ~0.0056s -> 0.0036

CHANGELOG.rst
```diff
@@ -39,6 +47,7 @@ Changed
 - feat: optimisation changes to BaseTransfomer and imputers file. Edited to reduce number of copies and type changes from to/from_native calls, and select/with_columns being called many times. `#444 <https://github.com/azukds/tubular/issues/444>_`
 - feat: added 'return_native' argument to BaseTransfomer to control whether native or narwhals types are returned, and limit type changes. Idea is for this to be rolled out across transformers.
 - feat: made creation of copies in BaseTransfomer optional, and default to False.
+- feat: optimisations for MappingTransformer and BaseMappingTransformerMixin
```
cjmwills: Q: had we made mapping transformer optimisations for the 1.4.5 release?

tubular/nominal.py
```diff
@@ -1081,65 +1105,102 @@ def transform(self, X: FrameT) -> FrameT:
...
+        # set up list of paired condition/outcome tuples for mapping
+        conditions_and_outcomes = {
+            output_col: [
+                self._create_mapping_conditions_and_outcomes(
+                    input_col,
+                    key,
+                    self.mappings,
+                    output_col=output_col,
+                )
+                for key in self.mappings[output_col]
```
cjmwills, line 1147: Q: did we see nested for loops causing many issues when we profiled? I know they can be slow, but assuming given they'll be short lists that's not going to be a big time sink

cjmwills (commented):
Not many comments to leave on this one, looks really good! Nice example of the existing unit tests giving us confidence no behaviour has changed too.

tubular/nominal.py
liamholmes31: gave the profiling another skim, the comprehensions mostly all show up at around 0.0003 seconds so not something to hugely worry about I think. The slowest is the one that has a .unique call inside (0.001s), and that's more down to the call than the comp, and think we do need these calls

CHANGELOG.rst
liamholmes31: ah yes, leftover from conflict resolution i think - thanks https://github.com/azukds/tubular/pull/475/commits/962642891912016fc772e98d46d41030cc0f3e03

tubular/nominal.py
cjmwills: Makes sense, thanks for checking

cjmwills (approved):

Issue #466, opened 2025-08-07
optimise onehotencoder


liamholmes31:
What?
need to:
- move away from narwhalify in favour of new utils functions
- remove to/from native calls with code, and make use of return_native_override=False argument when calling base classes
- move logic to build longer expressions and minimise with_columns calls
    - think we will need two with_columns calls, one for checks and one for main logic
    - for checks, see an example here https://github.com/azukds/tubular/pull/461
    - for the main logic, think we should ditch the current approach. Suggest doing something like:
    ```
    {
col + self.separator + str(missing_level):
 nw.col(col)==category) for col in self.columns for category in self.categories_[col]
}
    ```

see PR on similar work here https://github.com/azukds/tubular/pull/452

Why?
speed up live scoring

PR #477, opened 2025-08-20
Feature/opt one hot encoder

nesa1995:
Optimize one hot encoder transformer by:

Moving away from using narwhalify decorator
Removing to/from native calls withing the code and making use of return_native_override=False argument when calling base classes
Building out the expressions before using functions such as with_columns
The ticket can be found: https://github.com/azukds/tubular/issues/466

### DIFF
--- a/tubular/nominal.py
+++ b/tubular/nominal.py
@@ -1509,6 +1509,42 @@ class OneHotEncodingTransformer(
         ]
 
     @beartype
+    def _check_for_nulls(self, X: DataFrame) -> None:
+        """check that transformer being called on only non-null columns.
+
+        Note, found including nulls to be quite complicated due to:
+        - categorical variables make use of NaN not None
+        - pl/nw categorical variables do not allow categories to be edited,
+        so adjusting requires converting to str as interim step
+        - NaNs are converted to nan, introducing complications
+
+        As this transformer is generally used post imputation, elected to remove null
+        functionality.
+
+        Parameters
+        ----------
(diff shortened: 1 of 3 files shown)


tubular/nominal.py
```diff
@@ -1524,15 +1508,59 @@ def _get_feature_names(
             column + self.separator + str(level) for level in self.categories_[column]
         ]
 
-    @nw.narwhalify
-    def transform(self, X: FrameT) -> FrameT:
+    @beartype
+    def _check_for_nulls(self, X: DataFrame) -> None:
```
nesa1995: Realised maybe a different approach since we want to reduce the calls to .select and we have 2 checks to perform

alphanumericmale (commented):
Looks good - very thorough. Has it been profiled or will that be done for the next release?

liamholmes31:
Profiling results (on a simple one col, 5 row df):
Fit
- tubular pre narwhals: 0.007s
- tubular post narwhals, pre refactor, pandas: 0.006s
- tubular post narwhals, post refactor, pandas: 0.0015s
- tubular post narwhals, post refactor, polars: 0.0007s

transform
- tubular pre narwhals: 0.011s
- tubular post narwhals, pre refactor, pandas: 0.011s
- tubular post narwhals, post refactor, pandas: 0.0046s
- tubular post narwhals, post refactor, polars: 0.0015s

alphanumericmale (approved):
⚡10x ! That is a great performance gain.

Issue #438, opened 2025-07-15
add sum option to aggregation transformers


liamholmes31:
this is currently blocked by a narwhals issue where groupby sum is inconsistent for pandas/polars on all null cols. pandas will sum these to 0 and polars to None, and I think None is the expected behaviour

will raise on narwhals board as next step

```
import pandas as pd
import polars as pl
import narwhals as nw

df_dict={'a': [None], 'c': ['B']}

polars_df=pl.DataFrame(df_dict)
pandas_df=pd.DataFrame(df_dict)

print(nw.from_native(polars_df).group_by('c').agg(nw.col('a').sum().item()) ->None
print(nw.from_native(pandas_df).group_by('c').agg(nw.col('a').sum().item()) -> 0
```

liamholmes31:
opened narwhals issue here https://github.com/narwhals-dev/narwhals/issues/3022

liamholmes31:
https://github.com/azukds/tubular/pull/479

PR #479, opened 2025-08-22
optimised  agg transformer, and added sum option

liamholmes31:
standard optimisation changes (less narwhalification, fewer with_columns calls)

also made small change to add 'sum' agg option

https://github.com/azukds/tubular/issues/438

### DIFF
--- a/tests/aggregations/test_AggregateRowOverColumnsTransformer.py
+++ b/tests/aggregations/test_AggregateRowOverColumnsTransformer.py
@@ -71,7 +71,7 @@ class TestAggregateRowOverColumnsTransformerTransform(
         [
             # Test cases for "min", "max", "mean", "median",  and "count"
             (
-                ["min", "max", "mean", "median", "count"],
+                ["min", "max", "mean", "median", "count", "sum"],
                 {
                     "a": [1, 2, 3, 4, 8],
                     "b": [2, 3, 4, 5, 9],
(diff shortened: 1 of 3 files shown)


tests/aggregations/test_AggregateRowOverColumnsTransformer.py
```diff
@@ -71,7 +71,7 @@ def test_invalid_key_error(
         [
             # Test cases for "min", "max", "mean", "median",  and "count"
```
bayesandconfused: N: comment to include "sum"

tubular/aggregations.py
```diff
@@ -128,7 +143,7 @@ def transform(
             msg = f"{self.classname}: attempting to call transformer on non-numeric columns {non_numerical_columns}, which is not supported"
             raise TypeError(msg)
 
-        return X
+        return _return_narwhals_or_native_dataframe(X, return_native=return_native)
```
bayesandconfused, line 146: Q: Am i understanding this right that rather than using the @narwhalify decorator functionality we are essentially more precisely applying the narwhals dataframe conversion only at the point of returning the dataframe or transformation and this is improving the time profile?

liamholmes31: yeah this is right - in the previous setup the nested narwhalification calls were leading to many conversions to/from, and while the conversion is quite fast in single row context it did become impactful

tests/aggregations/test_AggregateRowOverColumnsTransformer.py
liamholmes31: thanks updated here https://github.com/azukds/tubular/pull/479/commits/607cbfef60cdf3dbe8237046afb8c9f76b6005e0

PR #480, opened 2025-08-27
refactored dummy weights code into mixin method

liamholmes31:
moved repeated code setting up dummy weight column into centralised mixin method, and added additional tests for this

### DIFF
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -757,82 +757,6 @@ class DummyWeightColumnMixinTests:
         ["pandas", "polars"],
         indirect=True,
     )
-    @pytest.mark.parametrize("i", [0, 1, 2, 3, 4])
-    def test_warnings_raised_for_existing_dummy_weight_columns(
-        self,
-        i,
-        minimal_attribute_dict,
-        minimal_dataframe_lookup,
-        uninitialized_transformers,
-    ):
-        """Test warnings are raised when dummy weights columns already exist"""
-
-        args = minimal_attribute_dict[self.transformer_name].copy()
-        args["weights_column"] = None
-
-        df = minimal_dataframe_lookup[self.transformer_name]
-        uninitialized_transformer = uninitialized_transformers[self.transformer_name]
-
-        transformer = uninitialized_transformer(**args)
-
-        df_dict = {}
-
-        bad_weight_values = [2] + [1] * (len(df) - 1)
-
-        # cases are:
-        # -2 -> no dummy_weights_column exists
-        # -1 -> dummy_weights_column exists but has bad values
-        # 0 -> dummy_weights_column, dummy_weights_column_0 exist with
-        # bad values
-        # ...
-        for j in range(-2, i):
-            if j == -2:
-                pass
-            elif j == -1:
-                df_dict["dummy_weights_column"] = bad_weight_values
-            else:
-                df_dict[f"dummy_weights_column_{j}"] = bad_weight_values
-
-        df = nw.from_native(df)
-        backend = nw.get_native_namespace(df)
-        new_cols = [
-            nw.new_series(name=name, values=df_dict[name], backend=backend).alias(name)
-            for name in df_dict
-        ]
-        df = df.with_columns(
-            new_cols,
-        )
-        df = df.to_native()
-
-        with pytest.warns(
-            UserWarning,
-        ) as record:
-            transformer.fit(df, df["a"])
-
-        previous_dummy_weights_columns = [
-            "dummy_weights_column",
-            *[f"dummy_weights_column_{i}" for i in range(i)],
-        ]
-        final_dummy_weights_columns = [
-            f"dummy_weights_column_{i}" for i in range(i + 1)
-        ]
-
-        for j, msg in enumerate(
-            [
-                f"{previous_dummy_weights_column} already exists in X but is not all 1, attempting to add {final_dummy_weights_column}"
-                for previous_dummy_weights_column, final_dummy_weights_column in zip(
-                    previous_dummy_weights_columns,
-                    final_dummy_weights_columns,
-                )
-            ],
-        ):
-            assert msg in str(record[j].message)
-
-    @pytest.mark.parametrize(
-        "minimal_dataframe_lookup",
-        ["pandas", "polars"],
-        indirect=True,
-    )
     def test_errors_if_too_many_failed_attempts(
         self,
         minimal_attribute_dict,
@@ -852,19 +776,8 @@ class DummyWeightColumnMixinTests:
         df_dict = {}
 
         bad_weight_values = [2] + [1] * (len(df) - 1)
-        # cases are:
-        # -2 -> no dummy_weights_column exists
-        # -1 -> dummy_weights_column exists but has bad values
-        # 0 -> dummy_weights_column, dummy_weights_column_0 exist with
-        # bad values
-        # ...
-        for j in range(-2, 5):
-            if j == -2:
-                pass
-            elif j == -1:
-                df_dict["dummy_weights_column"] = bad_weight_values
-            else:
-                df_dict[f"dummy_weights_column_{j}"] = bad_weight_values
+
+        df_dict["unit_weights_column"] = bad_weight_values
 
         df = nw.from_native(df)
         backend = nw.get_native_namespace(df)
@@ -877,7 +790,7 @@ class DummyWeightColumnMixinTests:
         )
         df = df.to_native()
 
-        msg = "Taking too long to find unused name for dummy weights column, consider renaming columns like 'dummy_weights_column' in X"
+        msg = "Attempting to insert column of unit weights named 'unit_weights_column', but an existing column shares this name and is not all 1, please rename existing column"
         with pytest.raises(
             RuntimeError,
             match=msg,
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -751,6 +751,127 @@ class WeightColumnFitMixinTests:
             transformer.fit(df, df["a"])
 
 
+class DummyWeightColumnMixinTests:
+    @pytest.mark.parametrize(
+        "minimal_dataframe_lookup",
+        ["pandas", "polars"],
+        indirect=True,
+    )
+    @pytest.mark.parametrize("i", [0, 1, 2, 3, 4])
+    def test_warnings_raised_for_existing_dummy_weight_columns(
+        self,
+        i,
+        minimal_attribute_dict,
+        minimal_dataframe_lookup,
+        uninitialized_transformers,
+    ):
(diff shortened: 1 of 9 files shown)


tubular/capping.py
```diff
@@ -190,30 +190,24 @@ def fit(self, X: FrameT, y: None = None) -> BaseCappingTransformer:
             Required for pipeline.
```
Boluwatife28, line 191: Should we beartype this fit method and use "DataFrame" instead of FrameT?

liamholmes31: wouldn't say this is completely in scope, but was able to do easily/without breaking any tests so have added :) https://github.com/azukds/tubular/pull/480/commits/23a9aed55a2a9bac277cdfbf02311bf5dbb11f0e

tests/base_tests.py
```diff
@@ -751,6 +751,127 @@ def test_zero_total_weight_error(
             transformer.fit(df, df["a"])
 
 
+class DummyWeightColumnMixinTests:
+    @pytest.mark.parametrize(
+        "minimal_dataframe_lookup",
+        ["pandas", "polars"],
+        indirect=True,
+    )
+    @pytest.mark.parametrize("i", [0, 1, 2, 3, 4])
```
nesa1995: N: Very minor and can leave, but could we combine these decorators into one?

nesa1995: Occurs throughout these file changes

tubular/mixins.py
```diff
@@ -172,6 +174,80 @@ def classname(self) -> str:
         """Method that returns the name of the current class when called."""
         return type(self).__name__
 
+    @staticmethod
+    def _create_dummy_weights_column(
+        X: DataFrame,
+        backend: Literal["pandas", "polars"],
+        return_native: bool = True,
+    ) -> tuple[DataFrame, str]:
+        """Create dummy weights column. Useful to streamline logic and just treat all
```
nesa1995: N: Update to docstring to make it easier to follow what is happening here. We are creating a dummy weight column in 2 conditions: it does not exist or it does but it isn't valid in that it should have a constant of 1. The purpose being to not recreate if it has been already created

tubular/mixins.py
```diff
@@ -172,6 +174,80 @@ def classname(self) -> str:
...
+
+        X = _convert_dataframe_to_narwhals(X)
+
+        original_dummy_weights_column = "dummy_weights_column"
+        final_dummy_weights_column = copy(original_dummy_weights_column)
+
+        # first search for existing dummy weights column
+        continue_search = True
+        i = 0
+        while continue_search is True:
```
nesa1995: N: Can be simplified to: while continue search

tubular/mixins.py
```diff
@@ -172,6 +174,80 @@ def classname(self) -> str:
...
+                    original_dummy_weights_column + "_" + str(i)
+                )
+                warnings.warn(
+                    f"{previous_dummy_weights_column} already exists in X but is not all 1, attempting to add {final_dummy_weights_column}",
+                    stacklevel=2,
+                )
+                i = i + 1
+
+            else:
+                continue_search = False
```
nesa1995: N: Comment to exit search if the column does not exist

tubular/mixins.py
```diff
@@ -172,6 +174,80 @@ def classname(self) -> str:
...
+                        return_native,
+                    ), final_dummy_weights_column
+
+                # if exists and not valid, search for alternative name
+                previous_dummy_weights_column = final_dummy_weights_column
+                final_dummy_weights_column = (
+                    original_dummy_weights_column + "_" + str(i)
+                )
+                warnings.warn(
+                    f"{previous_dummy_weights_column} already exists in X but is not all 1, attempting to add {final_dummy_weights_column}",
```
nesa1995: Q: Just for my understanding, why would we ever have a dummy weights column with the given name that would not be valid?

tests/base_tests.py
```diff
@@ -751,6 +751,127 @@ def test_zero_total_weight_error(
...
+
+        df = minimal_dataframe_lookup[self.transformer_name]
+        uninitialized_transformer = uninitialized_transformers[self.transformer_name]
+
+        transformer = uninitialized_transformer(**args)
+
+        df_dict = {}
+
+        bad_weight_values = [2] + [1] * (len(df) - 1)
+        for j in range(-2, i):
```
nesa1995, line 781: Q: Why not start with: for j in range(-1, i)?

nesa1995: QWhy not have: for j in range(-1, i):?

nesa1995: Q: Just to double check, are we searching for original_dummy_weights_column and if not then for a valid column named original_dummy_weights_column_0, original_dummy_weights_column_1.. etc?

tests/mixins/test_WeightColumnMixin.py
```diff
@@ -1,11 +1,199 @@
...
+        )
+
+        output, dummy_weights_column = obj._create_dummy_weights_column(
+            df,
+            backend=library,
+        )
+
+        assert_frame_equal_dispatch(expected, output)
+
+        print(dummy_weights_column)
```
nesa1995: N: Remove any print statements

tubular/mixins.py
liamholmes31: I don't think it would ever happen in reality, just wanted to cover all cases haha, it is probably overengineered

tests/base_tests.py
liamholmes31: hm I think leaving separate can be nicer in some cases, because it handles the combinations for you, rather than having to write out e.g.
```
[
('pandas', 0),
('polars', 0),
('pandas', 1),
....
]
```

tubular/mixins.py
liamholmes31: updated here, let me know what you think!

https://github.com/azukds/tubular/pull/480/commits/b60a0e0478fc5ac24bc3f966738186b471d42570

tubular/mixins.py
liamholmes31: updated here :) https://github.com/azukds/tubular/pull/480/commits/b60a0e0478fc5ac24bc3f966738186b471d42570

tubular/mixins.py
liamholmes31: updated here :) https://github.com/azukds/tubular/pull/480/commits/b60a0e0478fc5ac24bc3f966738186b471d42570

tests/base_tests.py
liamholmes31: added comments here, let me know if they help :) https://github.com/azukds/tubular/pull/480/commits/79210aef8b2e8958ef576c961b0b91bc6c88dad0

and yes second comment is correct!

tests/mixins/test_WeightColumnMixin.py
liamholmes31: thanks have removed :) https://github.com/azukds/tubular/pull/480/commits/0ad02d6df6437f785e67f2d6727d7fa21ed2d206

tubular/mixins.py
nesa1995: Q: should this be: exit search since column name not found? and so it moves on to create the dummy weights column

tubular/mixins.py
nesa1995: I see. I suppose I am wondering whether we do the search for another/ if we just throw an error if this is the case? It is just if we can see this being adapted somewhere in the pipeline for any scenario

tests/base_tests.py
nesa1995: Not sure if I am missunderstanding here, but I think you are adding the following columns: dummy_weights_column, dummy_weights_column_0, dummy_weights_column_1.. etc.. in your df_dict to then check that you throw the warnings up to dummy_weights_column_4. This forloop is slightly confusing and I don't think the first check is useful. Let me know what you think?

tubular/mixins.py
liamholmes31: simplified to your suggestion here, see what you think https://github.com/azukds/tubular/pull/480/commits/a073acb8d3712fb7c9b87a0a30075d79117afa7e

and fixed last tests here https://github.com/azukds/tubular/pull/480/commits/7eebdcd3fe8434ddf7ff51dcbaaf69375c8ddc20

tubular/mixins.py
liamholmes31: this has now been removed with the simplifying changes https://github.com/azukds/tubular/pull/480/commits/a073acb8d3712fb7c9b87a0a30075d79117afa7e

tests/base_tests.py
liamholmes31: this has now been removed with the simplifying changes :) https://github.com/azukds/tubular/pull/480/commits/a073acb8d3712fb7c9b87a0a30075d79117afa7e

tests/base_tests.py
nesa1995: Good point :)

tubular/mixins.py
nesa1995: This looks great :) thanks for updating

tests/base_tests.py
```diff
@@ -751,6 +751,53 @@ def test_zero_total_weight_error(
             transformer.fit(df, df["a"])
 
 
+class DummyWeightColumnMixinTests:
+    @pytest.mark.parametrize(
+        "minimal_dataframe_lookup",
+        ["pandas", "polars"],
+        indirect=True,
+    )
+    def test_errors_if_too_many_failed_attempts(
```
nesa1995: Q: Does this test need a rename and the doctring updated since we are testing that an error is thrown when the column exists but doesn't have a constant of 1?

tubular/mixins.py
nesa1995: This looks great, thank you! :)

tests/mixins/test_WeightColumnMixin.py
```diff
@@ -1,11 +1,112 @@
...
+        assert_frame_equal_dispatch(expected, output)
+
+        assert unit_weights_column == "unit_weights_column"
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_existing_column_used_if_possible(
+        self,
+        library,
+    ):
+        """Test unit weights column created as expected"""
```
nesa1995: N: Update docstring as this is used rather than recreated

liamholmes31: fixed here :) https://github.com/azukds/tubular/pull/480/commits/2adb61bdda2e7f3a322ebc09dfb8c3ec58b8d1b3

tests/base_tests.py
liamholmes31: fixed here :) https://github.com/azukds/tubular/pull/480/commits/2adb61bdda2e7f3a322ebc09dfb8c3ec58b8d1b3

Issue #473, opened 2025-08-19
Optimisations imputer fit methods


adamsardar:
**What**

Use idiomatic narwhals to avoid frequent materialisations

**Why**

Speed and memory need improving

**How**

- We have [a tight loop over columns ](https://github.com/azukds/tubular/blob/2768f765ae271a4fa656a2cf3c01ee04362918b4/tubular/imputers.py#L451) that filters the table, assigns to variable, operates on that subset, potentially multiplies by a weights column, then extracts an item. Instead create expressions, employ [`drop_nulls`](https://narwhals-dev.github.io/narwhals/api-reference/expr/#narwhals.Expr.drop_nulls) and [`to_dict()`](https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.to_dict)

Consider merging the `mean`, `median` and `mode` routines into an expression builder.

PR #481, opened 2025-08-28
Feature/opt imputers fit

liamholmes31:
- made changes to mean/median/mode imputer to primarily manipulate expressions
- this was less successful with weighted median imputer as this depends on data ordering and so needs to materialize
- cleaned up transform method of NullIndicator which was missed in previous PR

### DIFF
--- a/tubular/_stats.py
+++ b/tubular/_stats.py
@@ -13,6 +13,17 @@ def _get_median_calculation_expression(
 ) -> nw.Expr:
     """Produce expressions for calculating medians in provided dataframe.
 
+    Note, this function supports either operating on raw columns or chained expressions,
+    this is to enable chaining together longer expressions across transformers. For
+    example, we may wish to find the mode of a column that has already been mapped,
+    in which case (in pseudocode) we would do something like:
+
+    mapped_expr=nw.col('c').map_batches(...)
+    mode_expr=_get_mode_calculation_expressions(
+                columns='c',
+                initial_columns_exprs=mapped_expr
+                )
+
     Parameters
     ----------
     column: list[str]
@@ -62,6 +73,17 @@ def _get_mean_calculation_expressions(
     initial_weights_expr: Optional[nw.Expr] = None,
 ) -> dict[str, nw.Expr]:
     """Produce expressions for calculating means in provided dataframe.
+
+    Note, this function supports either operating on raw columns or chained expressions,
+    this is to enable chaining together longer expressions across transformers. For
+    example, we may wish to find the mean of a column that has already been mapped,
+    in which case (in pseudocode) we would do something like:
+
+    mapped_expr=nw.col('c').map_batches(...)
+    mean_expr=_get_mean_calculation_expressions(
+                columns=['c'],
+                initial_columns_exprs={'c': mapped_expr}
+                )
 
     Parameters
     ----------
@@ -131,6 +153,17 @@ def _get_mode_calculation_expressions(
     initial_weights_expr: Optional[nw.Expr] = None,
 ) -> dict[str, nw.Expr]:
     """Produce expressions for calculating modes in provided dataframe.
+
+    Note, this function supports either operating on raw columns or chained expressions,
+    this is to enable chaining together longer expressions across transformers. For
+    example, we may wish to find the mode of a column that has already been mapped,
+    in which case (in pseudocode) we would do something like:
+
+    mapped_expr=nw.col('c').map_batches(...)
+    mode_expr=_get_mode_calculation_expressions(
+                columns=['c'],
+                initial_columns_exprs={'c': mapped_expr}
+                )
 
     Parameters
     ----------


tubular/imputers.py
```diff
@@ -66,6 +66,32 @@ def _generate_imputation_expressions(self, expr: nw.Expr, col: str) -> nw.Expr:
...
+            list of columns in dataframe to check
+
+        Returns
+        -------
+        list[str]: list of all null columns
+
+        """
+        null_exprs = {c: nw.col(c).is_null().all() for c in columns}
+
+        null_results = X.select(**null_exprs).to_dict(as_series=False)
```
adamsardar: Q. What would happen if a lazy frame were passed in? In 'normal' polars you'd have to call `collect()` (or `.lazy().select(...).collect().to_dict()`). Unsure what the approach is for narhwlas.

tubular/imputers.py
```diff
@@ -66,6 +66,32 @@ def _generate_imputation_expressions(self, expr: nw.Expr, col: str) -> nw.Expr:
             else expr
         )
 
+    @staticmethod
+    def _get_all_null_columns(
```
adamsardar: Q. Why both with this as a static method? It makes no reference to the object; just have it as a standalone helper function?

tubular/imputers.py
```diff
@@ -428,6 +454,62 @@ def __init__(
 
         WeightColumnMixin.check_and_set_weight(self, weights_column)
 
+    @staticmethod
```
adamsardar: Q. again, why the static method. Why does `MedianImputer` need to 'own' this function?

When a message of '_get_median_calculation_expression' is passed to `MedianImputer` as opposed to some other object, it needs to know what to do in order to respond. Here there is no context, so having it as an object method isn't needed.

tubular/imputers.py
```diff
@@ -428,6 +454,62 @@ def __init__(
...
+        initial_weights_expr: nw.Expr
+            initial expression for weights column. Defaults to None,
+            and in this case nw.col(weights_column) is taken as the initial expr
+
+        Returns
+        ----------
+        median_value_exprs: dict[str, nw.Expr]
+            dict of format col: expression for calculating median
+
+        all_null_columns: list[str]
```
adamsardar: N: typo?

tubular/imputers.py
```diff
@@ -428,6 +454,62 @@ def __init__(
...
+
+        if weights_column is not None:
+            cumsum_weights_expr = initial_weights_expr.cum_sum()
+
+            median_expr = initial_column_expr.filter(
+                cumsum_weights_expr >= (initial_weights_expr.sum() / 2.0),
+            ).min()
+
+        else:
+            median_expr = initial_column_expr.filter(
```
adamsardar: S: consider using [drop_nulls()](https://narwhals-dev.github.io/narwhals/api-reference/expr/#narwhals.Expr.drop_nulls)? This keeps the expression unevaluated for as long as possible (as opposed to calling filter for every column).

tubular/imputers.py
```diff
@@ -449,35 +531,47 @@ def fit(self, X: DataFrame, y: Optional[Series] = None) -> MedianImputer:
...
         self.impute_values_ = {}
 
-        for c in self.columns:
-            # filter out null rows so their weight doesn't influence calc
-            filtered = X.filter(~nw.col(c).is_null())
-
-            # if column is only nulls, then median is None
-            if len(filtered) <= 0:
-                self.impute_values_[c] = None
+        all_null_cols = self._get_all_null_columns(X, self.columns)
```
adamsardar: N: why not just have this as a `defaultdict(lambda: None)` and warn `if all_null_cols`? Fewer lines of code, less complexity in seeing what's going on.

tubular/imputers.py
```diff
@@ -449,35 +531,47 @@ def fit(self, X: DataFrame, y: Optional[Series] = None) -> MedianImputer:
...
+        else:
+            median_exprs = {
+                c: self._get_median_calculation_expression(c, None)
+                for c in not_all_null_columns
+            }
+            results_dict = X.select(
+                **median_exprs,
+            ).to_dict(as_series=False)
+
+            for c in not_all_null_columns:
```
adamsardar: S: the dict `.update()` [method](https://www.w3schools.com/python/ref_dictionary_update.asp) would be more pythonic. Also, having to pull the first value out - are there times that two cvalues are returned by the computation above?

adamsardar: Noticed your comment below about return being `key: [value]`. narwhals doesn't have a `to_dicts` (just `to_dict`). You could still do this:

```python
self.impute_values_.update({key: value[0] for key, value in d.items()})
```
Which more efficient and cuts down on the for loop bloat.

tubular/imputers.py
```diff
@@ -525,6 +619,64 @@ def __init__(
...
+            and in this case nw.col(weights_column) is taken as the initial expr
+
+        Returns
+        ----------
+        mean_value_exprs: dict[str, nw.Expr]
+            dict of format col: expression for calculating means
+
+        """
+
+        if initial_columns_exprs is None:
```
adamsardar: I've stared at this a few times and have no idea what is going on. I'd hate to have to bugfix this ...

Any chance that another layout might make it more clear?

tubular/imputers.py
liamholmes31: just had a play with testing a LazyFrame, turns out there's a lot of code throughout the package that is currently broken for these. Suggest we spinout lazyframe support into a wider issue/stream of work

tubular/imputers.py
liamholmes31: refactor ideas weren't leaping out to me, so have just added more in way of docstring and commentary for now - let me know your thoughts https://github.com/azukds/tubular/pull/481/commits/9717558719c653760a0b11c1fabdce53c6ae1f27

tubular/imputers.py
liamholmes31: thanks, have made this change for median and  mean imputers
https://github.com/azukds/tubular/pull/481/commits/3b7ec594ff25fd20be1c1aea45a204bb679fe497

https://github.com/azukds/tubular/pull/481/commits/c493ab495b34e61a25ad1bd00f099c1cf35c0c01

tubular/imputers.py
liamholmes31: thanks, have made this change for median and mode imputers
https://github.com/azukds/tubular/pull/481/commits/21027055d910bac4ee1acb5a95b2573ec81701d0

https://github.com/azukds/tubular/pull/481/commits/f0c2e3854824c235df12c8a19d2472c32f5a431e

I think we do want transformers to explicitly store info on columns mapped to null so that a user can see this, so have added touches to these dict entries to preserve this behaviour, which means we do still end up with some looping

tubular/imputers.py
liamholmes31: thanks, have done this throughout file where applicable https://github.com/azukds/tubular/pull/481/commits/79c829e86d7a99fb1ecc5b392169a8545d448217

tubular/imputers.py
liamholmes31: thanks, removed https://github.com/azukds/tubular/pull/481/commits/56cf01c24487589ba48790593e3959c001182a72

tubular/imputers.py
liamholmes31: yeah fair point, have converted to functions and moved out to a separate file https://github.com/azukds/tubular/pull/481/commits/c55faa2c30df7242576faad2def85685a21912f9

https://github.com/azukds/tubular/pull/481/commits/acade3ed039eb901bc92f37ffc47ca401989bdec

tubular/imputers.py
liamholmes31: yeah fair point, have converted to functions and moved out to a separate file

https://github.com/azukds/tubular/commit/acade3ed039eb901bc92f37ffc47ca401989bdec

tubular/imputers.py
liamholmes31: this one doesnt work for dumping to json unfortunately, as the type is too complex, so have reverted here https://github.com/azukds/tubular/pull/481/commits/97719cd6440c1c338500c724bf6e6ba6511fe470

tubular/_stats.py
```diff
@@ -0,0 +1,183 @@
...
+    Parameters
+    ----------
+
+    column: list[str]
+        column to find median for
+
+    weights_column: str
+        name of weights column
+
+    initial_column_expr: nw.Expr
```
SadieRobertson, line 35: Q: Could this also require a list of expressions as in the mean function?

tubular/_stats.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Returns
+    ----------
+    mean_value_exprs: dict[str, nw.Expr]
+        dict of format col: expression for calculating means
+
+    """
+
+    # if a more complex starting expression for c or weights has been passed,
+    # (e.g. we may be worthing with a version of c that has been mapped)
```
SadieRobertson: N: typo, working

tubular/_stats.py
```diff
@@ -0,0 +1,183 @@
+from typing import Optional
```
SadieRobertson, line 1: Q: I'm wondering if maybe some sort of base expression class which contains the desired structure, including initial columns and weights expressions, could aid understanding

tubular/_stats.py
liamholmes31: fixed here https://github.com/azukds/tubular/pull/481/commits/258ba496b459977ba9edd8a8dcedd9590385fc21

tubular/_stats.py
liamholmes31: for the median case, it actually can't as has to be done one column at a time (due to data sorting) - but this did highlight that some of these type hints were wrong so have updated and added bear type 

https://github.com/azukds/tubular/pull/481/commits/ed5b5c75ec678d166ee39d42ffb3e8f96f0c1a40

and fixed lints/errors introduced here https://github.com/azukds/tubular/pull/481/commits/3e603353ca25df996365c88a8f9feb9e509b0858

tubular/imputers.py
liamholmes31: added more commentary this during second review https://github.com/azukds/tubular/commit/f5367a1ee59defdffe743c220369276fa846eec0

tubular/imputers.py
liamholmes31: have spun up separate milestone/set of issues re lazyframes https://github.com/azukds/tubular/milestone/7

tubular/_stats.py
SadieRobertson: Additional comments in docstrings have aided clarity functions. Comments on this issue address further discussion which took place around this point https://github.com/azukds/tubular/issues/473

PR #483, opened 2025-08-29
Feature/opt group rare levels fit

liamholmes31:
moved to expressions first approach in .fit method, to optimise code

### DIFF
--- a/tubular/nominal.py
+++ b/tubular/nominal.py
@@ -489,12 +489,11 @@ class GroupRareLevelsTransformer(BaseTransformer, WeightColumnMixin):
 
         weights_column = self.weights_column
         if self.weights_column is None:
-            X = WeightColumnMixin._create_dummy_weights_column(
+            X, weights_column = WeightColumnMixin._create_dummy_weights_column(
                 X,
                 backend=native_backend.__name__,
                 return_native=False,
             )
-            weights_column = "dummy_weights_column"
 
         for c in self.columns:
             cols_w_sums = X.group_by(c, drop_null_keys=False).agg(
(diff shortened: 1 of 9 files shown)


tubular/nominal.py
```diff
@@ -369,29 +369,21 @@ def _check_for_nulls(self, X: DataFrame) -> None:
         ...    )
 
         >>>  # non erroring example
-        >>> test_df=pl.DataFrame({'a': ['x', 'y'], 'b': ['w', 'z']})
```
SadieRobertson, line 372: N: It's a little confusing that the dict is called df here and in the second example below

tubular/nominal.py
```diff
@@ -341,7 +341,7 @@ def _check_str_like_columns(self, schema: nw.Schema) -> None:
             raise TypeError(msg)
 
     @beartype
-    def _check_for_nulls(self, X: DataFrame) -> None:
+    def _check_for_nulls(self, present_levels: dict[str, list[Any]]) -> None:
```
SadieRobertson, line 344: S: I think it would also be useful to have a unit test or example that shows this ValueError is raised when the transform method is called. At the moment we don't check that present_levels will be generated correctly from the df passed into transform

tubular/nominal.py
liamholmes31: whoops, changed here https://github.com/azukds/tubular/pull/483/commits/695e7cab5b7c0a874e390cfd2f9152bbe3ec7d01

tubular/nominal.py
liamholmes31: added here :) https://github.com/azukds/tubular/pull/483/commits/695e7cab5b7c0a874e390cfd2f9152bbe3ec7d01

(transform was actually giving an unhandled error for nulls, so this was a good suggestion!)

Issue #149, opened 2023-12-13
Refactor MeanResponseTransformer


davidhopkinson26:
MeanResponseTransformer class is very long and needs breaking down for maintainability and comprehensibility.

This ticket needs expansion but copying across a comment from #135 

"Would be nice not to have the super.transform and self.map_imputation_values twice, and feels confusing that in once case we call the self.check_mappable_rows class method, and in the other we ignore this and call a slightly different but similarly named method."

Potentially we should consider breaking this transformer up into a couple of classes.  Needs more thought.

PR #486, opened 2025-09-04
Feature/MRE refactor

liamholmes31:
moved MeanResponseTransformer fit to expression first approach

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,7 @@ Each individual change should have a link to the pull request after the descript
 Changed
 ^^^^^^^
 - feat: first pass refactor of MeanResponseTransformer `#149 <https://github.com/azukds/tubular/issues/149>`_
+- turned on PT ruff ruleset `#114 <https://github.com/azukds/tubular/issues/114>`_
 
 2.6.0 (19/12/2025)
 ------------------
--- a/tests/nominal/test_MeanResponseTransformer.py
+++ b/tests/nominal/test_MeanResponseTransformer.py
@@ -90,7 +90,7 @@ def create_MeanResponseTransformer_test_df_unseen_levels(library="pandas"):
     return df.to_native()
 
 
-@pytest.fixture()
+@pytest.fixture
 def learnt_mapping_dict():
     full_dict = {}
 
@@ -113,7 +113,7 @@ def learnt_mapping_dict():
     return full_dict
 
 
-@pytest.fixture()
+@pytest.fixture
 def learnt_unseen_levels_encoding_dict_mean():
     return {
         "b": (1.0 + 2.0 + 3.0 + 4.0 + 5.0 + 6.0) / 6,
@@ -123,7 +123,7 @@ def learnt_unseen_levels_encoding_dict_mean():
     }
 
 
-@pytest.fixture()
+@pytest.fixture
 def learnt_unseen_levels_encoding_dict_median():
     return {
         "b": 3.0,
@@ -133,7 +133,7 @@ def learnt_unseen_levels_encoding_dict_median():
     }
 
 
-@pytest.fixture()
+@pytest.fixture
 def learnt_unseen_levels_encoding_dict_highest():
     return {
         "b": 6.0,
@@ -143,7 +143,7 @@ def learnt_unseen_levels_encoding_dict_highest():
     }
 
 
-@pytest.fixture()
+@pytest.fixture
 def learnt_unseen_levels_encoding_dict_lowest():
     return {
         "b": 1.0,
@@ -153,7 +153,7 @@ def learnt_unseen_levels_encoding_dict_lowest():
     }
 
 
-@pytest.fixture()
+@pytest.fixture
 def learnt_unseen_levels_encoding_dict_arbitrary():
     return {
         "b": 22.0,
@@ -536,7 +536,7 @@ class TestFit(GenericFitTests, WeightColumnFitMixinTests, DummyWeightColumnMixin
         )
 
     @pytest.mark.parametrize("library", ["pandas", "polars"])
-    @pytest.mark.parametrize("prior", (1, 3, 5, 7, 9, 11, 100))
+    @pytest.mark.parametrize("prior", [1, 3, 5, 7, 9, 11, 100])
     def test_prior_logic(self, prior, library):
         "Test that for prior>0 encodings are closer to global mean than for prior=0."
         df = create_MeanResponseTransformer_test_df(library=library)
@@ -589,7 +589,7 @@ class TestFit(GenericFitTests, WeightColumnFitMixinTests, DummyWeightColumnMixin
     @pytest.mark.parametrize("library", ["pandas", "polars"])
     @pytest.mark.parametrize(
         ("low_weight", "high_weight"),
-        ((1, 2), (2, 3), (3, 4), (10, 20)),
+        [(1, 2), (2, 3), (3, 4), (10, 20)],
     )
     def test_prior_logic_for_weights(self, low_weight, high_weight, library):
         "Test that for fixed prior a group with lower weight is moved closer to the global mean than one with higher weight."
@@ -1250,7 +1250,7 @@ class TestTransform(GenericTransformTests):
     @pytest.mark.parametrize("from_json", [True, False])
     @pytest.mark.parametrize("library", ["pandas", "polars"])
     @pytest.mark.parametrize(
-        "prior, level, target, unseen_level_handling",
+        ("prior", "level", "target", "unseen_level_handling"),
         [
             (5, "all", "c", "mean"),
             (100, ["a", "b"], "c", "min"),
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -2,6 +2,9 @@
 
 name: Python package build and test
 
+permissions:
+  contents: read
+
 on:
   push:
   pull_request:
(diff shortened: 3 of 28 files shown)


tubular/nominal.py
```diff
@@ -879,125 +880,130 @@ def get_feature_names_out(self) -> list[str]:
...
+                    self.encoded_columns_to_columns[encoded_column]
+                ]
+                + self.prior
+            ).cast(
+                getattr(nw, self.return_type),
+            )
+            for encoded_column in self.encoded_columns
+        }
+
+    def _setup_fit_multi_level(
```
liamholmes31, line 1124: haven't made efforts to refactor more than necessary here, but think through this work it's starting to become clearer what the [refactor](https://github.com/azukds/tubular/issues/149) of this (very overcomplicated) class would look like, possibly with:
- a MultiResponseLevelEncodingClass (here you get outputs for column a over levels x,y,z like a_x, a_y, a_z)
    - though unsure if anyone even uses this atm 
- a BinaryResponseLevelEncodingClass (here the output for column a is just a, as the average is just over the target 0/1 so can be condensed down to one encoding)
- an ABC class that contains the shared logic

liamholmes31: Other distinct bits of functionality are:
- the prior encoding
    - this is a bit complicated, but actually integrates quite well into the rest of the code as the no-prior case is just equivalent to prior=0, so there are no branches introduced 
- unseen level handling
    - this is less seamless and does introduce branching for with/without unseen level handling
    - does anyone use the unseen_levels=10 hardcoded type handling? maybe this can just be dropped
    - then there would still be 5 branches introduced by this, for mean/median/min/max/none.. this is still a lot
    - would it be better if this transformer mapped unseen levels to None generally and it was left to users to include an imputation step afterwards?

tests/nominal/test_MeanResponseTransformer.py
```diff
@@ -210,18 +210,24 @@ def test_output1(self, library):
 
         x.fit(X=df, y=y)
 
-        expected1 = (1 + 3 * 2.5) / (1 + 3)
+        # note, previously global mean was learnt during the fit
+        # call above and used inside _prior_regularisation
+        # now it is derived as an expression inside
+        # _prior_regularisation
+        expected1 = (1 + 3 * (5 / 3)) / (1 + 3)
```
liamholmes31: this seems to have been incorrect before, as the weight below sums to 3 and the y values sum to 5

CHANGELOG.rst
```diff
@@ -43,7 +43,6 @@ Changed
 
 - feat: optimisation changes to fit methods in imputers file
 - feat: optimisation changes to transform method of NullIndicator
-- feat: optimisation changes to MeanResponseTransformer
```
liamholmes31, line 120: pretty sure this wasn't in the last version

tubular/nominal.py
```diff
@@ -1212,91 +1235,126 @@ def fit(self, X: FrameT, y: nw.Series) -> FrameT:
...
 
-            self.column_to_encoded_columns = {
-                column: [
-                    *self.column_to_encoded_columns[column],
-                    mapping_columns_for_this_level[column],
-                ]
-                for column in self.columns
-            }
+        # need to materialise these here in order to use .over with pandas
+        X_y = X_y.with_columns(**all_response_exprs)
```
cjmwills, line 1313: Q: does materialising here cost us much performance? If so and if it's only needed to calculate weighted means, could be worth considering if that functionality is used enough to justify.

cjmwills: And a wider / more important question, do we know the performance improvement before / after this PR?

tubular/nominal.py
```diff
@@ -1212,91 +1235,126 @@ def fit(self, X: FrameT, y: nw.Series) -> FrameT:
...
+                backend=backend.__name__,
                 return_native=False,
             )
 
         WeightColumnMixin.check_weights_column(self, X, weights_column)
 
-        if (response_null_count := y.is_null().sum()) > 0:
+        y_vals = y.unique().to_list()
+
+        response_null_count = sum(pd.isna(val) for val in y_vals)
```
cjmwills: Q: is there a reason we couldn't just use pl.is_nan? Do pl/pd calculate missings differently?

tubular/nominal.py
```diff
@@ -1060,123 +1060,146 @@ def get_feature_names_out(self) -> list[str]:
...
+                        self.encoded_columns_to_response_columns[encoded_column]
+                    ]
+                    * nw.lit(self.prior)
+                )
+            )
+            / (
+                weight_sum_over_groups_exprs[
+                    self.encoded_columns_to_columns[encoded_column]
+                ]
+                + self.prior
```
cjmwills: S: should we wrap this in nw.lit for consistency with line 1113?

tests/nominal/test_MeanResponseTransformer.py
cjmwills: does this mean the previous function was calculting it wrong? Or just that the test was wrong?

tests/nominal/test_MeanResponseTransformer.py
```diff
@@ -215,13 +221,28 @@ def test_output1(self, library):
...
+        results = group_means_and_weights_df.select(**output_exprs)
+
+        output = {
+            "a": results.item(0, 0),
+            "b": results.item(1, 0),
+        }
+
         assert output == expected, (
             f"output of _prior_regularisation not as expected, expected {expected} but got {output}"
         )
```
cjmwills: Q: the test is working as expected, but I'm a bit confused by the structure of it. We fit MeanResponseTransformer with column a and target y, but then later on we have output for columns a and b that use a different target column. Is the fitting at the beginning just to access `_prior_regularisation`? I was similarly confused with the others in TestPriorRegularisation

tubular/nominal.py
cjmwills: Agree with the direction of this, I did find the current set up tricky to follow. My instinct is the functionality for any MultiResponseLevelEncodingClass isn't currently used much, so would be worth considering with any refactor. Even if we don't go as far as dropping it, we could refactor in a way that prioritised the BinaryResponseLevelEncodingClass given the importance of that.

liamholmes31: good point! we could go with binary first, and put the usual deprecation warning on this one to invite people to create an issue if they do use the multi level stuff

tubular/nominal.py
liamholmes31: thanks for the prompt on this, I had been checking on small/dummy data frames but this nudged me to put a more realistic 200k rows through and found that I had actually made things worse.. 

I think I overgeneralised a bit between fit/transform which are optimised a bit differently for small/large data, e.g. for a small example dataset like {'a': ['a', 'b', 'c']} it isn't worth materialising a grouped frame as it doesn't reduce the size at all - but for like {'a': ['a', 'b', 'c']*100000} it's worth materialising as it reduces the size for the end calculations by loads.

So basically have now reverted to be pretty similar to the old method, and this is now more of a refactor PR. I think it's still worth completing as it will help pave the way for a bigger refactor further down the line, but I'll update the PR title/description/changelog

polars old  0.010043323040008545
pandas old  0.06000413656234741

polars new  0.009704241752624512
pandas new  0.05942017316818237

tests/nominal/test_MeanResponseTransformer.py
liamholmes31: I think it was just wrong - I've now refactored these test cases to just be part of TestFit, the old tests were quite tied to implementation and also had lots of room for errors to creep in!

https://github.com/azukds/tubular/pull/486/commits/b1751acf2e7be5b1c380fe067e9b5b1a62f3e6dd

tubular/nominal.py
liamholmes31: re the materialisation, it doesn't seem to be a bottleneck when I put through snakeviz. Two bits that do standout are:

- the unit weights handling, have spun up an issue here https://github.com/azukds/tubular/issues/666
- the unseen_level fitting, which I think we'll address in the refactor

tubular/nominal.py
liamholmes31: good idea! https://github.com/azukds/tubular/pull/486/commits/c11deedcc8ef246739078464c53aad89d5efb222

tubular/nominal.py
liamholmes31: good idea https://github.com/azukds/tubular/pull/486/commits/5c558a37eb4ead7618e4e75a20ffe1fd8be6d68b

tests/nominal/test_MeanResponseTransformer.py
liamholmes31: agree these were quite messy, I've rewritten now to just run as part of TestFit which has simplified a lot I think - see what you think https://github.com/azukds/tubular/commit/b1751acf2e7be5b1c380fe067e9b5b1a62f3e6dd

cjmwills: That's clearer thanks!

tubular/nominal.py
cjmwills: Thanks, good thinking. The unit weights handling could be a good one to tackle early as (my understanding is) this transformer is most commonly run with no weights.

cjmwills (approved):
Looks good. Agree it's worth getting this in even if the outcome differed from original intention (optimisation vs refactor).

PR #493, opened 2025-09-17
Feature/beartype new column name

liamholmes31:
converted NewColumnNameMixin and associated transformers to beartype

Removed BaseDateTwoColumnTransformer as this is now fully covered by type hints

https://github.com/azukds/tubular/issues/560

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,6 +34,7 @@ Each individual change should have a link to the pull request after the descript
 - feat: removed NewColumnNameMixin and converted associated transformers/tests to beartype
 - feat: removed TwoColumnMixin and BaseDateTwoColumnTransformer and replaced with beartype
 - feat: added `get_feature_names_out` method to all transformers, which gives back list of modified/created features. This follows the sklearn API.
+- feat: added `to_json` method for BaseGenericDateTransformer `#540 <https://github.com/azukds/tubular/issues/540>_`
 
 2.0.0 (16/10/2025)
 ------------------
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -1134,26 +1134,6 @@ class ColumnsCheckTests:
     Note this deliberately avoids starting with "Tests" so that the tests are not run on import.
     """
 
-    @pytest.mark.parametrize("non_list", [1, True, {"a": 1}, None, "True"])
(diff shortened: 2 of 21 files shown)


tubular/dates.py
```diff
@@ -355,12 +359,13 @@ class attribute, indicates whether transform requires fit to be run first
 
     jsonable = False
 
+    @beartype
     def __init__(
         self,
-        columns: list[str],
-        new_column_name: str | None = None,
+        columns: Union[list[str], str],
+        new_column_name: str,
```
cjmwills, line 392: Q: was it intentional to no longer allow this to be None?

tests/base_tests.py
```diff
@@ -73,7 +73,7 @@ def test_columns_empty_list_error(
         args = minimal_attribute_dict[self.transformer_name].copy()
         args["columns"] = []
 
-        with pytest.raises(ValueError):
+        with pytest.raises(BeartypeCallHintParamViolation):
```
cjmwills, line 76: Q: just to confirm, is this why we've been able to remove all the `test_new_column_name_type_error` from dates module?

CHANGELOG.rst
```diff
@@ -30,7 +30,9 @@ Each individual change should have a link to the pull request after the descript
 2.1.0 (unreleased)
 ------------------
 
-- feat: - feat: converted DropOriginalMixin and associated transformers to beartype
+- feat: converted DropOriginalMixin and associated transformers to beartype
```
cjmwills, line 33: Q: was this in a previous PR?

tubular/dates.py
liamholmes31: yeah that's right, looking through the file this never gets passed down to the base class as None, so made the change

tests/base_tests.py
liamholmes31: yeah the `test_new_column_name_type_error` had been overloaded for a few classes which were already beartyped, as they had different error behaviour to the mixin which was not yet beartyped - but now the they've all been brought in line the inherited test works for all

CHANGELOG.rst
liamholmes31: yes, this one here https://github.com/azukds/tubular/pull/430

PR #496, opened 2025-09-23
added demo to/from json functionality for imputers file

liamholmes31:
- added to/from json methods for base and imputer classes
- added `jsonable` class attr to mark which transformer do/don't have this functionality
- setup transform tests to do an addition run with json dumped/loaded versions of the transformer and verify no behaviour changes

(there are lots of files touched here, but mostly just adds an attribute!

### DIFF
--- a/tubular/imputers.py
+++ b/tubular/imputers.py
@@ -93,6 +93,14 @@ class BaseImputer(BaseTransformer):
         {'tubular_version': ..., 'classname': 'MeanImputer', 'init': {'columns': ['a', 'b'], 'copy': False, 'verbose': False, 'return_native': True, 'weights_column': None}, 'fit': {'impute_values_': {'a': 1.0, 'b': 2.0}}}
 
         """
+        if not self.jsonable:
+            msg = (
+                "This transformer has not yet had to/from json functionality developed"
+            )
+            raise RuntimeError(
+                msg,
+            )
+
         self.check_is_fitted("impute_values_")
 
         json_dict = super().to_json()
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -1340,6 +1340,10 @@ class ToFromJsonTests:
 
         df = minimal_dataframe_lookup[self.transformer_name]
 
+        # skip polars test if not narwhalified
+        if not transformer.polars_compatible and isinstance(df, pl.DataFrame):
+            return
+
         # skip test if transformer is jsonable
         if transformer.jsonable:
             return


tubular/base.py
```diff
@@ -107,6 +114,70 @@ def __init__(
         self.copy = copy
         self.return_native = return_native
 
+    def to_json(self) -> dict[str, dict[str, Any]]:
```
cjmwills, line 131: Q/S: given every transformer will inherit from this method, is it worth ensuring we reduce as much potential duplication as possible? When I asked copilot it's suggestion was having an `_extra_json_init_keys` and `_extra_json_fit_keys` method which can be overridden in child classes. They would be empty dicts in this base class, with something like `**self._extra_json_init_keys()` in the creation of the json. Then all child classes would override with attributes specific to that transformer e.g. `if hasattr(self, "weights_column"): keys["weights_column"] = getattr(self, weights_column", None)`. What are your thoughts on that?

tubular/imputers.py
```diff
@@ -47,8 +47,60 @@ class attribute, indicates whether transformer has been converted to polars/pand
 
     polars_compatible = True
 
+    jsonable = True
+
     FITS = False
 
+    def to_json(self) -> dict[str, dict[str, Any]]:
```
cjmwills, line 68: Q: is it intentional that there isn't an equivalent `from_json` method in this class? Is it because no specific behaviour is required that isn't provided in the the inherited method?

tests/utils.py
```diff
@@ -132,3 +133,31 @@ def dataframe_init_dispatch(
         "The library parameter should be either 'pandas' or 'polars'."
     )
     raise ValueError(library_error_message)
+
+
+def _handle_from_json(
```
cjmwills, line 138: P: neat way to structure this and apply to all tests!

tubular/base.py
liamholmes31: ah OK, to check i'm following is the suggestion that:
- only BaseTransformer has `to_json`
- all transformers implement the 2 extra methods, and to_json always calls these

hmmm, I suppose it's nice to have just one defined 'to_json' method, but some things putting me off:
- these two extra methods feel a bit messier/less readable than just having a layered to_json to me, also duplication is already kept down by having the super().to_json() call in the logic
- some transformers don't share the BaseTransformer init API, e.g. MappingTransformer(mappings=...), so these will need to redefine to_json anyway

what do you think?

tubular/imputers.py
liamholmes31: yeah this is correct :) have now added a to/from json example to the doctests in the imputers file, so this inherited behaviour will be clearer to a reader https://github.com/azukds/tubular/pull/496/commits/ea2e1b2862bf4bb06ece2fced9638ddccfbd8c2b

tubular/base.py
cjmwills: that's correct yep. I'll paste copilot's dummy code for how it would look below for clarity. Fair point that introducing these new methods wouldn't necessarily make it clearer - you'd still be defining custom logic but just in a different place essentially. I _think_ the difference in API's could be handled with this set up, given you'd just be adding to a dict that's then passed to `json_dict`? I'm happy to take your steer on this - just worth discussing given the other transformers will follow this logic.

```
class BaseTransformer(BaseEstimator, TransformerMixin):
    # ...existing code...

    def _extra_json_init_keys(self) -> dict[str, Any]:
        """Override in subclasses to add extra keys to 'init' dict."""
        return {}

    def _extra_json_fit_keys(self) -> dict[str, Any]:
        """Override in subclasses to add extra keys to 'fit' dict."""
        return {}

    def to_json(self) -> dict[str, dict[str, Any]]:
        """dump transformer to json dict"""
        json_dict = {
            "tubular_version": self._version,
            "init": {
                "columns": self.columns,
                "copy": self.copy,
                "verbose": self.verbose,
                "return_native": self.return_native,
                **self._extra_json_init_keys(),
            },
            "fit": {
                **self._extra_json_fit_keys(),
            },
        }
        return json_dict

    @classmethod
    def from_json(cls, json: dict[str, Any]) -> BaseTransformer:
        instance = cls(**json["init"])
        for attr, value in json["fit"].items():
            setattr(instance, attr, value)
        return instance
# ...existing code...
```

```
class BaseImputer(BaseTransformer):
    # ...existing code...

    def _extra_json_init_keys(self) -> dict[str, Any]:
        keys = {}
        if hasattr(self, "weights_column"):
            keys["weights_column"] = getattr(self, "weights_column", None)
        if hasattr(self, "impute_value"):
            keys["impute_value"] = getattr(self, "impute_value", None)
        return keys

    def _extra_json_fit_keys(self) -> dict[str, Any]:
        self.check_is_fitted("impute_values_")
        return {"impute_values_": self.impute_values_}
# ...existing code...
```

liamholmes31:
> > So my thinking on this is, to make our lives easier I would say the only behaviour we want to guarantee for a transfomer that's output by from_json is .transform. I think this is reasonable because the general workflow will be init->fit->to_json->from_json->transform. So something like transformer.from_json(..).fit() is an unusual usage.
> > So we could do something like:
> > 
> > * add an attribute which gets set in init like `from_json=False`
> > * in from_json we then set `from_json=True`
> > * we setup fit to just error out if `from_json=True`, and we can test for this behaviour
> > 
> > what do you think?
> 
> Agree with the usage, and approach sounds sensible to me! For my understanding, is the intention of this to fit a pipeline of several transformers as usual, but then export the pipeline using `to_json` rather than as, say, `.pkl`?

nice, i'll have a go at implementing that!

and yes that's right! we can add some code to handle that saving/reading to tubular later, but it should be quite simple, like:
```
json_pipeline={}
for name, step in pipeline.named_steps.items():
    json_pipeline[name]=step.to_json()
```

tubular/base.py
liamholmes31: yeah okay i'm with you - I think part of what puts me off is this doesn't follow the inheritance pattern we use for other methods in the repo, e.g. for fit each class just holds a fit method and makes use of super.fit - we don't have `_extra_fit_logic` etc. I think technically both would work so if you feel strongly can work this in, but it does feel slightly  'off' to me at first glance

cjmwills: No don't feel strongly, happy to keep as is!

liamholmes31:
> Thanks the implementation of `block_from_json` looks great. One last one before closing this off, all the classes that inherit from `BaseTransformer` can now run `to_json` even if the class attribute `jsonable = False`. Is that likely to cause any problems?
> 
> (Realising it's difficult to keep track of the conversation as comments - won't do that next time!)

thanks!! 

good point, have now added handling to error if these methods are called for a transformer where jsonable=False https://github.com/azukds/tubular/pull/496/commits/78c5e5cbdc5541517f6475aa78b49bc5fa8b4930

(all good :) )

PR #497, opened 2025-09-23
453 uncap polars version

alphanumericmale:
upgraded polars version to latest 1.33.0, which caused one test to fail - group rare levels transformer.

Investigations showed that the intended behaviour of the transformer was unaffected however, the levels still appeared when calling  nw.from_native(output_df)[column].cat.get_categories().to_list().

The cause of the issue seems to be how new polars deals with categorical data. Polars implements a category mapping for a series of that datatype which is not overwritten, after removal of categories. 

The Solution was to convert the categorical col into Enum which solved the issue, but required some adjustments to other tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "pandas>=1.5.0",
     "scikit-learn>=1.2.0",
     "narwhals >= 1.31.0",
-    "polars >= 1.33.0",
+    "polars >= 1.9.0",
     "beartype >= 0.19.0",
     "typing-extensions>=4.5.0",
 ]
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -19,7 +19,7 @@ dependencies = [
     "pandas>=1.5.0",
     "scikit-learn>=1.2.0",
     "narwhals >= 1.31.0",
-    "polars < 1.32.0",
+    "polars >= 1.33.0",
```
liamholmes31: probably better  to restore the earlier version and be less  restrictive: "polars >= 1.9.0"

bayesandconfused: gotcha!

bayesandconfused: fixed here: https://github.com/azukds/tubular/pull/497/commits/bd0a0da99d9dd07194b56450e1f3e1cedbee35b2

caused no changes to compiled requirements-dev.txt and tests still passed

Issue #504, opened 2025-09-25
doctests for mapping module


liamholmes31:
What?
 
- Add doctest examples for undeprecated classes/methods in mapping module
- Delete old example notebooks for this module

Why?
have maintained usage examples for our transformers

How?
see example PR https://github.com/azukds/tubular/pull/491

PR #515, opened 2025-09-29
added doctests to mapping file and deleted old notebooks

liamholmes31:
https://github.com/azukds/tubular/issues/504

Left out doctests for BaseMappingTransformMixin as:

- some of the methods output expressions, which seem to abstract to include in a docstring
- the transform method relies on the mixin being inherited by a mapping class, so the doctest felt messy


tubular/mapping.py
```diff
@@ -211,6 +248,9 @@ def _create_mapping_conditions_and_outcomes(
         Returns
         -------
         Tuple[nw.Expr, nw.Expr]: prepared pair of mapping condition/outcome
+
+        # currently not including doctests for this, as need to look into most meaningful
```
Ritunjai-Sharma, line 252: Have we created a separate issue/ticket for adding doctests for methods that output expressions and rely on the mixin being inherited by the class? Just so that we don't forget about this!

liamholmes31: good point, added here :) https://github.com/azukds/tubular/issues/518

Ritunjai-Sharma (approved):

PR #519, opened 2025-09-30
added doctests to dates file, deleted old example notebooks

liamholmes31:
https://github.com/azukds/tubular/issues/502

left out doctests for BaseDateTwoColumnTransformer as this will soon be replaced with beartype


tubular/dates.py
```diff
@@ -180,6 +208,34 @@ def transform(
         X : pd/pl.DataFrame
             Validated data
 
+        Example:
+        --------
+        >>> import polars as pl
```
SadieRobertson, line 213: Q: do we need import polars in each example? I'm wondering if doctest will have already imported it for this module's doctest as it's included in the previous example

tubular/dates.py
```diff
@@ -408,6 +511,34 @@ def transform(self, X: DataFrame) -> DataFrame:
...
             Data containing self.columns
 
+        Example:
+        --------
+        >>> import polars as pl
+        >>> import datetime
+
+        >>> transformer=DateDifferenceTransformer(
+        ... columns=["a",  "b"],
+        ... new_column_name='bla',
```
SadieRobertson: S: Could be helpful to use a more descriptive new column name here e.g. a_b_difference_years

tubular/dates.py
```diff
@@ -876,6 +1085,22 @@ def _process_provided_mappings(
         Returns
         -------
         None
+
+        Example:
```
SadieRobertson, line 1089: Q: Do we want to provide examples for private methods? Or should this method be made public?

tests/dates/test_DateTimeInfoExtractor.py
```diff
@@ -48,9 +48,10 @@ class TestInit(
     def setup_class(cls):
         cls.transformer_name = "DatetimeInfoExtractor"
 
-    @pytest.mark.parametrize("incorrect_type_include", [2, 3.0, "invalid", "dayofweek"])
-    def test_error_when_include_not_list(self, incorrect_type_include):
-        """Test that an exception is raised when value include variable is not a list."""
+    @pytest.mark.parametrize("incorrect_type_include", [2, 3.0, "invalid"])
```
SadieRobertson: Q: Should we also test if an invalid include option is contained within a list as part of the parameterise, e.g. ["dayofweek", 2.0]?

SadieRobertson (commented):
Have left a few comments but examples look good overall!

tubular/dates.py
liamholmes31: yeah fair question - I think they're still valuable for new developers coming to the code/reviewing, but can see that it's mixed messages for users with the private label. Think I lean to leaving in, but what do you think?

tubular/dates.py
liamholmes31: yeah it does seem to fail without this - not 100% on how doctest works but guess it must start a new session for each one?

tubular/dates.py
liamholmes31: updated https://github.com/azukds/tubular/pull/519/commits/ee74a345cc769ee5343ace0e46e705b7a781f6fd

tests/dates/test_DateTimeInfoExtractor.py
liamholmes31: added here :) https://github.com/azukds/tubular/pull/519/commits/ee74a345cc769ee5343ace0e46e705b7a781f6fd

tubular/dates.py
SadieRobertson: I wonder if it could be that polars isn't imported at the top of this file?

tubular/dates.py
SadieRobertson: Yeah I think it is ok to leave it in, the private label should be clear enough

PR #520, opened 2025-10-01
added doctests for nominal module

liamholmes31:
https://github.com/azukds/tubular/issues/506

added example usage doctests and removed old/stale notebooks for nominal file


tubular/nominal.py
```diff
@@ -67,6 +74,19 @@ def check_mappable_rows(
...
 
+        Example:
+        --------
+        >>> import polars as pl
+
+        >>> transformer = BaseNominalTransformer(
+        ... columns='a',
+        ...    )
+
+        >>> transformer.mappings={'a': {'a': 0, 'b': 1}}
```
danielarifmurphy: N: Nothing wrong here but think the double usage of `a` and `b` here could cause some minor confusion - would prefer `x`, `y` unless this is intentional

tubular/nominal.py
```diff
@@ -264,6 +316,21 @@ def _check_str_like_columns(self, schema: nw.Schema) -> None:
...
+        >>> transformer = GroupRareLevelsTransformer(
+        ... columns='a',
+        ... cut_off_percent=0.02,
+        ... rare_level_name='rare_level',
+        ...    )
+
+        >>> test_df=pl.DataFrame({'a': ['a','b'], 'b': ['c','d']})
+        >>> schema=nw.from_native(test_df).schema
+
+        >>> transformer._check_str_like_columns(schema)
```
danielarifmurphy, line 315: S: Would it potentially be more informative to given an example of an exception being raised?

tubular/nominal.py
```diff
@@ -429,6 +525,32 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+
+        >>> transformer = GroupRareLevelsTransformer(
+        ... columns='a',
+        ... cut_off_percent=0.5,
+        ... rare_level_name='rare_level',
+        ...    )
+
+        >>> test_df=pl.DataFrame({'a': ['a', 'a', 'b'], 'b': ['c', 'd', 'e']})
+
+        >>> _=transformer.fit(test_df)
```
danielarifmurphy, line 539: Q: for my benefit - do we have to assign here? do we need to avoid premature returns with doctest?

tubular/nominal.py
```diff
@@ -664,6 +794,8 @@ def _prior_regularisation(
         -------
         regularised : nw.Series
             Series of regularised encoding values
+
+        # not adding doctests yet as this method will change in an upcoming PR
```
danielarifmurphy: N: worth adding `TODO:`? I find this visually easier to notice personally

tubular/nominal.py
```diff
@@ -973,6 +1123,30 @@ def transform(self, X: DataFrame) -> DataFrame:
         X : pd/pl.DataFrame
             Transformed input X with levels mapped accoriding to mappings dict.
 
+        Example:
+        --------
+        >>> import polars as pl
+
+        >>> transformer=MeanResponseTransformer(
+        ... columns='a',
+        ... prior=1,
```
danielarifmurphy, line 1153: S: having this non zero weighting toward the global mean may mean this example is a little confusing in isolation? would be good to get an example of this `prior` behaviour separately

danielarifmurphy: ah actually ignore me - can see example will be added when the method changes!

tubular/nominal.py
```diff
@@ -1261,6 +1463,24 @@ def _warn_missing_levels(
...
+
+        >>> test_df=pl.DataFrame({'a': ['a', 'b'], 'b': [1,2]})
+
+        >>> _ = transformer.fit(test_df)
+
+        >>> transformer._warn_missing_levels(
+        ... present_levels=['a', 'b'],
+        ... c='a',
+        ... missing_levels={}
+        ... )
```
danielarifmurphy, line 1511: Q: out of interest are we able to `ruff format` doctest examples or equivalent?

tubular/nominal.py
```diff
@@ -1314,6 +1549,28 @@ def transform(
...
+        >>> _ = transformer.fit(test_df)
+
+        >>> transformer.transform(test_df)
+        shape: (2, 4)
+        ┌─────┬─────┬───────┬───────┐
+        │ a   ┆ b   ┆ a_a   ┆ a_b   │
+        │ --- ┆ --- ┆ ---   ┆ ---   │
+        │ str ┆ i64 ┆ bool  ┆ bool  │
+        ╞═════╪═════╪═══════╪═══════╡
+        │ a   ┆ 1   ┆ true  ┆ false │
```
danielarifmurphy: didn't actually realise we returned explicit bools!

tubular/nominal.py
danielarifmurphy, line 1: N: I notice the typing throughout this script is inconsistent - `Union` vs `|`, `Optional` vs `| None`, `ListOfStrs` vs `list[str]` etc - not a particularly value add task though so feel free to ignore this!

tubular/nominal.py
liamholmes31: yeah doctest validates every line, so by assigning am making this line not return anything and so pass (there may be a better way but learning doctest as we go..)

tubular/nominal.py
liamholmes31: I think we brought this behaviour in along with the polars support, as does seem better for this transformer!

tubular/nominal.py
liamholmes31: it looks like ruff currently doesn't do this meaningfully (the D lints will apply standard docstring formatting but not treated as code)

have spun up an issue to look into  https://github.com/azukds/tubular/issues/524

tubular/nominal.py
liamholmes31: there are other issues open around beartyping, but happy to squeeze in https://github.com/azukds/tubular/pull/520/commits/42ae2de3785020da67026ec3ae7d4b55e367eefd

tubular/nominal.py
liamholmes31: added exception examples https://github.com/azukds/tubular/pull/520/commits/941da5ec7330a506f15c36ba897c057228dfa460

had a go at adding one for the warn_missing_levels function, but seems doctest doesn't play well with testing warnings

tubular/nominal.py
liamholmes31: added https://github.com/azukds/tubular/pull/520/commits/94e24203adabb6386f4c872c50816e838a0f4f01

tubular/nominal.py
liamholmes31: added another example anyway! https://github.com/azukds/tubular/pull/520/commits/42ae2de3785020da67026ec3ae7d4b55e367eefd#diff-0ff13ed51eb89d895189e53aaadc09cae59088d56bf3514753d75c7311a37ff5

tubular/nominal.py
liamholmes31: fair point, updated throughout here! https://github.com/azukds/tubular/pull/520/commits/39af0fe0af34b955f439e5a4538ce56a09b69afa

Issue #535, opened 2025-10-08
make BaseTransformer lazyframe compatible


liamholmes31:
What?
this will involve turning off some of the data-specific checks we currently make, e.g. len(X)>0, but I think these are fairly low value anyway!

Why?
allow use of lazyframes!

liamholmes31:
as part of this will add lazyframe testing for all of the transformers that are already lazyframe compatible

PR #538, opened 2025-10-09
Feature/lazyframes

liamholmes31:
- rolled out lazyframe testing framework to all classes that are already lazyframe compatible

https://github.com/azukds/tubular/issues/535


tests/base_tests.py
```diff
@@ -56,6 +60,22 @@ def test_verbose_non_bool_error(
                 **minimal_attribute_dict[self.transformer_name],
             )
 
+    def test_unexpected_kwarg_error(
```
liamholmes31, line 65: this was under transform tests for some reason, moved here

tests/dates/test_DateTimeInfoExtractor.py
```diff
@@ -747,28 +793,24 @@ def test_custom_mappings_can_be_used(self, library):
...
+            )
             df_expected_row = expected[[i]].to_native()
 
             assert_frame_equal_dispatch(
-                df_transformed_row,
+                _collect_frame(df_transformed_row, polars, lazy),
                 df_expected_row,
             )
 
-    def test_is_serialisable(self, tmp_path):
```
liamholmes31, line 767: this test is inherited from the base test class, so removed

tests/imputers/test_BaseImputer.py
```diff
@@ -372,26 +434,23 @@ def test_nulls_imputed_correctly_weights(
 
         df_transformed = transformer.transform(df)
 
-        # Convert both DataFrames to a common format using Narwhals
```
liamholmes31, line 402: this one wasn't actually testing single row transforms, so rewrote

tests/imputers/test_NullIndicator.py
```diff
@@ -52,40 +53,49 @@ def expected_df_1(self, request):
...
         transformer = NullIndicator(columns=columns)
 
-        df_transformed = transformer.transform(df)
+        polars = isinstance(df, pl.DataFrame)
+
+        if u._check_if_skip_test(transformer, df, lazy):
+            return
 
-        # Convert both DataFrames to a common format using Narwhals
-        df_transformed_common = nw.from_native(df_transformed)
```
liamholmes31, line 77: this one wasn't actually testing single row transforms, so rewrote

tubular/base.py
```diff
@@ -186,17 +177,9 @@ def _combine_X_y(self, X: FrameT, y: nw.Series) -> FrameT:
...
         """
-        if not isinstance(X, (nw.DataFrame, nw.LazyFrame)):
-            msg = f"{self.classname()}: X should be a polars or pandas DataFrame/LazyFrame"
-            raise TypeError(msg)
 
-        if not isinstance(y, nw.Series):
-            msg = f"{self.classname()}: y should be a polars or pandas Series"
-            raise TypeError(msg)
-
-        if X.shape[0] != y.shape[0]:
```
liamholmes31, line 367: think this check is a more valid than the len(X)>0 type checks, open to reviewers thoughts on whether we should still include some handling. Could e.g. return _temporary_response as null and add an 'error_msg' col explaining why

Issue #489, opened 2025-09-08
create DifferenceTransformer to replace TwoColumnOperatorTransformer


liamholmes31:
What?
have seen TwoColumnOperatorTransformer used for calculating:
- sum
- difference (a-b)
- ration (a/b)
from two columns. The sum functionality is now covered by transformers in the aggregation file, but would be useful to create a DifferenceTransformer which could cover ratios/subtraction and allow us to deprecate TwoColumnOperatorTransformer.

Why?
allow us to deprecate TwoColumnOperatorTransformer

How?
transformer would need to accept as args:
- two columns in a list
- a new column name (or default to e.g. a_subtract_b)
- an operation (subtract or divide)

would not have a fit method, and transform would build expressions using the given columns and operation.

PR #544, opened 2025-10-13
Difference Transformer

hariinduu:
Created a new difference transformer to deprecate TwoColumnOperatorTransformer.

#489

### DIFF
--- a/tubular/numeric.py
+++ b/tubular/numeric.py
@@ -30,7 +30,13 @@ from tubular.mixins import (
     NewColumnNameMixin,
     TwoColumnMixin,
 )
-from tubular.types import DataFrame, FloatTypeAnnotated, ListOfTwoStrs
+from tubular.types import (
+    DataFrame,
+    FloatTypeAnnotated,
+    ListOfOneStr,
+    ListOfTwoStrs,
+    PositiveNumber,
+)
 
 if TYPE_CHECKING:
     from narwhals.typing import FrameT, IntoSeriesT
@@ -244,7 +250,10 @@ class OneDKmeansTransformer(BaseNumericTransformer, DropOriginalMixin):
     @beartype
     def __init__(
         self,
-        columns: Union[str, list[str]],
+        columns: Union[
+            str,
+            ListOfOneStr,
+        ],
         new_column_name: str,
         n_init: Union[str, int] = "auto",
         n_clusters: int = 8,
@@ -252,10 +261,6 @@ class OneDKmeansTransformer(BaseNumericTransformer, DropOriginalMixin):
         kmeans_kwargs: Optional[dict[str, object]] = None,
         **kwargs: dict[str, bool],
     ) -> None:
-        if (isinstance(columns, list)) and (len(columns) > 1):
-            msg = f"{self.classname()}: columns arg should be a single str or a list length 1 giving the column to group."
-            raise TypeError(msg)
-
         if kmeans_kwargs is None:
             kmeans_kwargs = {}
 
@@ -263,6 +268,7 @@ class OneDKmeansTransformer(BaseNumericTransformer, DropOriginalMixin):
         self.new_column_name = new_column_name
         self.n_init = n_init
         self.kmeans_kwargs = kmeans_kwargs
+        self.drop_original = drop_original
 
         if isinstance(columns, str):
             self.columns = [columns]
@@ -270,7 +276,33 @@ class OneDKmeansTransformer(BaseNumericTransformer, DropOriginalMixin):
             self.columns = columns
 
         super().__init__(columns=[columns], **kwargs)
-        self.set_drop_original_column(drop_original)
+
+    def get_feature_names_out(self) -> list[str]:
+        """list features modified/created by the transformer
+
+        Returns
+        -------
+        list[str]:
+            list of features modified/created by the transformer
+
+        Examples
+        --------
+
+        >>> transformer = OneDKmeansTransformer(
+        ... columns='a',
+        ... n_clusters=2,
+        ... new_column_name="kmeans_column",
+        ... drop_original=False,
+        ... kmeans_kwargs={"random_state": 42},
+        ...  )
+
+        >>> transformer.get_feature_names_out()
+        ['kmeans_column']
+        """
+
+        return [
+            self.new_column_name,
+        ]
 
     @nw.narwhalify
     def fit(self, X: FrameT, y: IntoSeriesT | None = None) -> OneDKmeansTransformer:
@@ -476,10 +508,11 @@ class LogTransformer(BaseNumericTransformer, DropOriginalMixin):
 
     FITS = False
 
+    @beartype
     def __init__(
         self,
-        columns: str | list[str] | None,
-        base: float | None = None,
+        columns: Optional[Union[str, list[str]]],
+        base: Optional[PositiveNumber] = None,
         add_1: bool = False,
         drop_original: bool = True,
         suffix: str = "log",
@@ -487,24 +520,7 @@ class LogTransformer(BaseNumericTransformer, DropOriginalMixin):
     ) -> None:
         super().__init__(columns=columns, **kwargs)
 
-        if not isinstance(add_1, bool):
-            add_1_error_msg = f"{self.classname()}: add_1 should be bool"
-            raise TypeError(add_1_error_msg)
-
-        if not isinstance(suffix, str):
-            suffix_error_msg = f"{self.classname()}: suffix should be str"
-            raise TypeError(suffix_error_msg)
-
-        if base is not None:
-            if not isinstance(base, (int, float)):
-                msg = f"{self.classname()}: base should be numeric or None"
-                raise ValueError(msg)
-            if not base > 0:
-                msg = f"{self.classname()}: base should be strictly positive"
-                raise ValueError(msg)
-
-        self.set_drop_original_column(drop_original)
-
+        self.drop_original = drop_original
         self.base = base
         self.add_1 = add_1
         self.suffix = suffix
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,12 @@ We use the tags:
 
 Each individual change should have a link to the pull request after the description of the change.
 
+2.1.0 (unreleased)
+------------------
+
+- feat: - feat: converted DropOriginalMixin and associated transformers to beartype
+- feat: added `get_feature_names_out` method to all transformers, which gives back list of modified/created features. This follows the sklearn API.
+
 2.0.0 (16/10/2025)
 ------------------
 
@@ -34,7 +40,6 @@ Each individual change should have a link to the pull request after the descript
 - feat: added AggregateColumnsOverRowTransformer `#385 <https://github.com/azukds/tubular/issues/385>_`
 - chore: renamed AggregateRowOverColumnsTransformer to AggregateRowsOverColumnTransformer
 - feat: optimisation changes to GroupRareLevelsTransformer fit method
-- placeholder
 - feat: optimisation changes to DatetimeSinusoidCalculator, added 'return_native_override' argument to DatetimeSinusoidCalculator, reduced with_columns being called many times. `<#465 <https://github.com/azukds/tubular/issues/465>_`
 - chore: turned on doctest
 - chore: deprecated DataFrameMethodTransformer
@@ -55,7 +60,6 @@ Each individual change should have a link to the pull request after the descript
 - chore: doctests for dates module `#502 <https://github.com/azukds/tubular/issues/502>_`
 - chore: doctests for nominal module `#506 <https://github.com/azukds/tubular/issues/506>_`
 - chore: doctests for imputers module `#503 <https://github.com/azukds/tubular/issues/503>_`
-
 
 1.4.8 (03/09/25)
 ----------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -165,3 +165,6 @@ fail_under = 80
 
 # Additional Options (AddOpts) for pytest
 addopts = '--doctest-modules'
+filterwarnings = [
+    "ignore::DeprecationWarning",
+]
--- a/tests/base/test_DataFrameMethodTransformer.py
+++ b/tests/base/test_DataFrameMethodTransformer.py
@@ -1,9 +1,8 @@
-import re
-
 import numpy as np
 import pandas as pd
 import pytest
 import test_aide as ta
+from beartype.roar import BeartypeCallHintParamViolation
 
 import tests.test_data as d
 from tests.base_tests import (
@@ -25,8 +24,7 @@ class DataFrameMethodTransformerInitTests(ColumnStrListInitTests):
         """Test an exception is raised if pd_method_kwargs not a dict"""
 
         with pytest.raises(
-            TypeError,
-            match=f"DataFrameMethodTransformer: pd_method_kwargs should be a dict but got type {type(not_dictionary)}",
+            BeartypeCallHintParamViolation,
         ):
             DataFrameMethodTransformer(
                 new_column_names="a",
@@ -45,14 +43,11 @@ class DataFrameMethodTransformerInitTests(ColumnStrListInitTests):
         }
 
         with pytest.raises(
-            TypeError,
-            match=re.escape(
-                f"DataFrameMethodTransformer: unexpected type ({type(not_string)}) for pd_method_kwargs key in position {1}, must be str",
-            ),
+            BeartypeCallHintParamViolation,
         ):
             DataFrameMethodTransformer(
                 new_column_names="a",
-                pd_method_name="b",
+                pd_method_name="max",
                 columns=["b", "c"],
                 pd_method_kwargs=pd_method_kwargs,
             )
@@ -62,10 +57,7 @@ class DataFrameMethodTransformerInitTests(ColumnStrListInitTests):
         """Test an exception is raised if pd_method_name is not a string"""
 
         with pytest.raises(
-            TypeError,
-            match=re.escape(
-                f"DataFrameMethodTransformer: unexpected type ({type(not_string)}) for pd_method_name, expecting str",
-            ),
+            BeartypeCallHintParamViolation,
         ):
             DataFrameMethodTransformer(
                 new_column_names="a",
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -142,8 +142,7 @@ class DropOriginalInitMixinTests:
         args["drop_original"] = drop_orginal_column
 
         with pytest.raises(
-            TypeError,
-            match=f"{self.transformer_name}: drop_original should be bool",
+            BeartypeCallHintParamViolation,
         ):
             uninitialized_transformers[self.transformer_name](**args)
 
@@ -1376,9 +1375,52 @@ class ToFromJsonTests:
             transformer.from_json({})
 
 
+class GetFeatureNamesOutTests:
+    """
+    Tests for the BaseTransformer.get_feature_names_out method.
+    Note this deliberately avoids starting with "Tests" so that the tests are not run on import.
+    """
+
+    @pytest.mark.parametrize(
+        "minimal_dataframe_lookup",
+        ["pandas", "polars"],
+        indirect=["minimal_dataframe_lookup"],
+    )
+    def test_get_feature_names_out_matches_new_features(
+        self,
+        minimal_dataframe_lookup,
+        initialized_transformers,
+    ):
+        """Test that the expected newly created features (if any) are indeed contained
+        in the output df"""
+
+        df = minimal_dataframe_lookup[self.transformer_name]
+
+        x = initialized_transformers[self.transformer_name]
+
+        # skip polars test if not narwhalified
+        if not x.polars_compatible and isinstance(df, pl.DataFrame):
+            return
+
+        if x.FITS:
+            x.fit(df, df["a"])
+
+        output = x.transform(df)
+
+        output_columns = set(output.columns)
+
+        expected_new_columns = set(x.get_feature_names_out())
+
+        # are expected columns in the data
+        assert expected_new_columns.intersection(output_columns), (
+            f"{x.classname()}: get_feature_names_out does not agree with output of .transform, expected {expected_new_columns} but got {output_columns}"
+        )
+
+
 class OtherBaseBehaviourTests(
     ColumnsCheckTests,
     CombineXYTests,
+    GetFeatureNamesOutTests,
     ToFromJsonTests,
 ):
     """
--- a/tests/comparison/test_EqualityChecker.py
+++ b/tests/comparison/test_EqualityChecker.py
@@ -1,5 +1,7 @@
+import numpy as np
 import pytest
 import test_aide as ta
+from beartype.roar import BeartypeCallHintParamViolation
 
 import tests.test_data as d
 from tests.base_tests import (
@@ -23,6 +25,27 @@ class TestInit(
     @classmethod
     def setup_class(cls):
         cls.transformer_name = "EqualityChecker"
+
+    # overload until we beartype the new_column_name mixin
+    @pytest.mark.parametrize(
+        "new_column_type",
+        [1, True, {"a": 1}, [1, 2], np.inf, np.nan],
+    )
+    def test_new_column_name_type_error(
+        self,
+        new_column_type,
+        minimal_attribute_dict,
+        uninitialized_transformers,
+    ):
+        """Test an error is raised if any type other than str passed to new_column_name"""
+
+        args = minimal_attribute_dict[self.transformer_name].copy()
+        args["new_column_name"] = new_column_type
+
+        with pytest.raises(
+            BeartypeCallHintParamViolation,
+        ):
+            uninitialized_transformers[self.transformer_name](**args)
 
 
 class TestFit(GenericFitTests):
--- a/tests/dates/test_BaseDatetimeTransformer.py
+++ b/tests/dates/test_BaseDatetimeTransformer.py
@@ -5,6 +5,7 @@ import numpy as np
 import pandas as pd
 import polars as pl
 import pytest
+from beartype.roar import BeartypeCallHintParamViolation
 
 from tests.base_tests import (
     ColumnStrListInitTests,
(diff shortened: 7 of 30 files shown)


tubular/numeric.py
```diff
@@ -1230,3 +1245,148 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+                          new_column_name='c', drop_original=False)
+    """
+    
+    polars_compatible = True
+    FITS = False
+    
+    @beartype
+    def __init__(
+        self,
+        columns: list[str],
```
liamholmes31: S: if you add something like the below to the types file and use to typehint, you can drop the TwoColumnMixin (this will be removed soon) 
```

ListOfTwoStrs = Annotated[
    list[str],
    Is[lambda list_arg: len(list_arg) == 2],
]
```
(taken from this open PR https://github.com/azukds/tubular/pull/430/files#diff-b8ba16edcde8126c16f5612a1e482c7ed028c818f9a65131b84021b160c68247)

tubular/numeric.py
```diff
@@ -1230,3 +1245,148 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        operation: ListOfDifferenceOperations,
+        drop_original: bool = False,
+        new_column_name: Optional[str] = None,
+        verbose: bool = False,
```
liamholmes31: S: don't need to include verbose explicitly, can just do `**kwargs: Optional[bool]`

tubular/numeric.py
```diff
@@ -1230,3 +1245,148 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+            Whether to drop the original columns after transformation.
+        new_column_name : Optional[str], default=None
+            Name of the column to store the result. If None, a default name will be generated.
+        verbose : bool, default=False
+            Whether to print verbose output during transformation.
+        """
+        super().__init__(columns=columns, verbose=verbose)
+        
+        # Validate columns
+        if len(columns) != 2:
```
liamholmes31: S: can remove this after the beartype suggestion above

tubular/numeric.py
```diff
@@ -1230,3 +1245,148 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        operation : ListOfDifferenceOperations
+            Operation to perform between columns. Must be either 'subtract' or 'divide'.
+        drop_original : bool, default=False
+            Whether to drop the original columns after transformation.
+        new_column_name : Optional[str], default=None
+            Name of the column to store the result. If None, a default name will be generated.
+        verbose : bool, default=False
+            Whether to print verbose output during transformation.
+        """
+        super().__init__(columns=columns, verbose=verbose)
```
liamholmes31: S: would pass **kwargs instead of verbose

tubular/numeric.py
```diff
@@ -1230,3 +1245,148 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        
+        # Validate columns
+        if len(columns) != 2:
+            msg = f"{self.classname()}: columns must be a list of exactly two column names"
+            raise ValueError(msg)
+            
+        self.columns = columns
```
liamholmes31: S: super.init sets columns, so don't need to do this here

Also, suggest not storing self.column1 and self.column2 and just accessing as columns[0]/columns[1] in logic. This is because we're in the process of adding to_json functionality, and don't want to clutter that with extra attrs if we can avoid!

tubular/numeric.py
```diff
@@ -1230,3 +1245,148 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        self.column1 = columns[0]
+        self.column2 = columns[1]
+        
+            
+        self.operation = operation
+        self.drop_original = drop_original
+        
+        # Set new_column_name or generate a default one
+        if new_column_name is None:
```
liamholmes31: S: maybe just drop new column name and always use the defaults - I've started to think where possible it's best to just fix the output names and let a user rename later themselves if they want to!

tubular/numeric.py
```diff
@@ -1230,3 +1245,148 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        
+        
+        # Create the expression based on operation
+        if self.operation == 'subtract':
+            expr = nw.col(self.column1) - nw.col(self.column2)
+        else:  # divide
+            expr = nw.col(self.column1) / nw.col(self.column2)
+            
+        # Add the new column
+        X = X.with_columns(**{self.new_column_name: expr})
```
liamholmes31: S: better as `X=X.with_columns(expr.alias(self.new_column_name))`

tubular/numeric.py
```diff
@@ -1230,3 +1245,148 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        │ 300   ┆ 200  ┆ 100    │
+        └───────┴──────┴────────┘
+        """
+        X = _convert_dataframe_to_narwhals(X)
+
+        X = super().transform(X, return_native_override=False)
+        
+        
+        # Create the expression based on operation
+        if self.operation == 'subtract':
```
liamholmes31: S: can do inline like 
```
expr= (nw.col(self.column1) - nw.col(self.column2)) if self.operation=='subtract' else  nw.col(self.column1) / nw.col(self.column2)
```

tubular/numeric.py
```diff
@@ -1230,3 +1245,148 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+            expr = nw.col(self.column1) - nw.col(self.column2)
+        else:  # divide
+            expr = nw.col(self.column1) / nw.col(self.column2)
```
liamholmes31: S: actually might need to rewrite this like:
```
nw.when(nw.col(self.column2)!=0).then(nw.col(self.column1) / nw.col(self.column2)).otherwise(None)
```

tubular/numeric.py
hariinduu: Done

tubular/numeric.py
hariinduu: modified

tubular/numeric.py
hariinduu: Added ListOfTwoStrs

tubular/numeric.py
hariinduu: Removed

tubular/numeric.py
hariinduu: modified

tubular/numeric.py
hariinduu: Modified expression

tubular/numeric.py
hariinduu: Dropped new column name

tubular/numeric.py
hariinduu: Modified

tubular/numeric.py
liamholmes31: thanks! remember to:

- update the type in the docstring
- remove the TwoColumnMixin inheritance

liamholmes31:
for each new class, could you also define a function like:
```
def get_feature_names_out(self)->list[str]:

    return [self.new_column_name]
```

want to start doing this as standard (based off of your pipeline work!)

tubular/numeric.py
```diff
@@ -1230,3 +1245,229 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+    
+    def get_feature_names_out(self) -> list[str]:
+        """Get the names of the output features.
+
+        Returns
+        -------
+        list[str]
+            List containing the name of the new column created by the transformation.
+        """
+        return [self.new_column_name]
```
liamholmes31: S: can change to "self.columns[0]}_divided_by_{self.columns[1]" (and similar change for the other one)

tubular/numeric.py
```diff
@@ -29,12 +29,27 @@
...
+from tubular.types import DataFrame, ListOfTwoStrs
+from enum import Enum
+from beartype.vale import Is
+from beartype.typing import Annotated
 
 if TYPE_CHECKING:
     from narwhals.typing import FrameT, IntoSeriesT
 
 
+class DifferenceOperationOptions(str, Enum):
```
liamholmes31: S: can now delete these two type classes

tubular/numeric.py
```diff
@@ -1230,3 +1245,229 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+    """
+
+    polars_compatible = True
+    FITS = False
+
+    @beartype
+    def __init__(
+        self,
+        columns: ListOfTwoStrs,
+        drop_original: bool = False,
```
liamholmes31: S: might be useful to have a return_dtype arg here that can be either 'Float32' or 'Float64' and then cast to this at the end in transform - I would have it default to 'Float32', and a user can decide if they need more precision than  this

tubular/numeric.py
```diff
@@ -1230,3 +1245,229 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        │ 100   ┆ 80   ┆ 1.25           │
+        │ 200   ┆ 150  ┆ 1.3333         │
+        │ 300   ┆ 200  ┆ 1.5            │
+        └───────┴──────┴────────────────┘
+        """
+        X = _convert_dataframe_to_narwhals(X)
+        X = super().transform(X, return_native_override=False)
+
+        # Create the division expression
+        expr = nw.when(nw.col(self.columns[1]) != 0).then(nw.col(self.columns[0]) / nw.col(self.columns[1])).otherwise(None)
```
liamholmes31: the cast would go on the end of here `.otherwise(None).cast(getattr(nw, self.return_dtype))`

tubular/numeric.py
liamholmes31: (this is less necessary on DifferenceTransformer, as this one will preserve type - whereas RatioTransformer will often convert from e.g. int to float, so it seems appropriate to be able to control which type of float)

tubular/numeric.py
```diff
@@ -1230,3 +1245,229 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+
+    This transformer allows performing subtraction between two columns in a DataFrame
+    and stores the result in a new column.
+
+    Attributes
+    ----------
+    columns : ListOfTwoStrs
+        List of exactly two column names to operate on. The first column is the minuend,
+        and the second column is the subtrahend.
+    drop_original : bool
```
liamholmes31: S: could you drop this argument as well for these? Have been thinking about getting rid of this mixin too, and dropping here means we don't need to do anything extra for the to/from json functionality to turn on :)

tubular/numeric.py
```diff
@@ -1230,3 +1245,229 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+    -------
+    >>> transformer = DifferenceTransformer(
+    ... columns=['a', 'b'],
+    ... drop_original=False,
+    ... )
+    DifferenceTransformer(columns=['a', 'b'], drop_original=False)
+    """
+    
+    polars_compatible = True
+    FITS = False
```
liamholmes31, line 1361: S: could you include a class attribute `jsonable=True` (will work once drop_original is gone)

tubular/numeric.py
```diff
@@ -1230,3 +1245,229 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+class DifferenceTransformer(BaseNumericTransformer):
+    """Transformer that performs subtraction operation between two columns.
+
+    This transformer allows performing subtraction between two columns in a DataFrame
+    and stores the result in a new column.
+
+    Attributes
+    ----------
+    columns : ListOfTwoStrs
+        List of exactly two column names to operate on. The first column is the minuend,
```
liamholmes31:
```suggestion
        List of exactly two column names to operate on. The second column is subtracted from the first.
```

tubular/numeric.py
```diff
@@ -1230,3 +1245,229 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        columns: ListOfTwoStrs,
+        drop_original: bool = False,
+        **kwargs: Optional[bool],
+    ) -> None:
+        """Initialize the DifferenceTransformer.
+        
+        Parameters
+        ----------
+        columns : ListOfTwoStrs
+            List of exactly two column names to operate on. The first column is the minuend/numerator,
```
liamholmes31:
```suggestion
            List of exactly two column names to operate on. The second column is subtracted from the first.
```

tubular/numeric.py
hariinduu: Modified

tubular/numeric.py
hariinduu: Removed both classes

tubular/numeric.py
hariinduu: Added a return_dtype argument

tubular/numeric.py
hariinduu: Done

tubular/numeric.py
hariinduu: Dropped this

tubular/numeric.py
hariinduu: Added

tubular/numeric.py
liamholmes31: looks like this line hasn't changed btw

tubular/numeric.py
```diff
@@ -1320,3 +1334,252 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        --------
+        >>> ratio_transformer = RatioTransformer(columns=['a', 'b'], return_dtype='Float32')
+        >>> ratio_transformer.to_json()
+        {'tubular_version': ..., 'classname': 'RatioTransformer', 'init': {'columns': ['a', 'b'], 'copy': False, 'verbose': False, 'return_native': True, 'return_dtype': 'Float32'}, 'fit': {}}
+        """
+        if not self.jsonable:
+            error_message = "This transformer does not support JSON serialization."
+            raise RuntimeError(error_message)
+
+        json_dict = {
```
liamholmes31: can delete lines 1488-1498, this is what super().to_json does :)

tubular/numeric.py
```diff
@@ -1320,3 +1334,252 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        dict[str, dict[str, Any]]:
+            JSON representation of the transformer, including init parameters.
+
+        Examples
+        --------
+        >>> ratio_transformer = RatioTransformer(columns=['a', 'b'], return_dtype='Float32')
+        >>> ratio_transformer.to_json()
+        {'tubular_version': ..., 'classname': 'RatioTransformer', 'init': {'columns': ['a', 'b'], 'copy': False, 'verbose': False, 'return_native': True, 'return_dtype': 'Float32'}, 'fit': {}}
+        """
+        if not self.jsonable:
```
liamholmes31: S: can remove this, as these transformers have been written to be jsonable

tubular/numeric.py
```diff
@@ -29,12 +32,23 @@
...
     TwoColumnMixin,
 )
-from tubular.types import DataFrame
+from tubular.types import DataFrame, ListOfTwoStrs
 
 if TYPE_CHECKING:
     from narwhals.typing import FrameT, IntoSeriesT
 
 
+class ReturnDtypeOptions(Enum):
```
liamholmes31: S:  I would actually move this to types.py and name `FloatTypeOptions`, think it's one that will be reused

tubular/numeric.py
```diff
@@ -1320,3 +1334,252 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        X = _convert_dataframe_to_narwhals(X)
+
+        X = super().transform(X, return_native_override=False)
+
+        # Create the subtraction expression
+        expr = nw.col(self.columns[0]) - nw.col(self.columns[1])
+
+        X = X.with_columns(expr.alias(self.new_column_name))
+
+        # Add the new column
```
liamholmes31: typo, repeated line here

tubular/numeric.py
```diff
@@ -1320,3 +1334,252 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+    ['a', 'b']
+    >>> transformer.return_dtype
+    'Float32'
+    """
+
+    polars_compatible = True
+    FITS = False
+    jsonable = True
+
+    def to_json(self) -> dict[str, dict[str, Any]]:
```
liamholmes31, line 1469: from utils, can you import 'block_from_json' and add as a decorator here

tests/test_data.py
```diff
@@ -671,3 +671,21 @@ def expected_date_diff_df_2(library="pandas"):
     }
 
     return u.dataframe_init_dispatch(df_dict, library=library)
+
+
+def create_difference_test_df(library="pandas"):
```
liamholmes31: S: would just put these in the two test files, they're not going to be reused so makes it easier to read if they're in the same location as the tests

tests/numeric/test_RatioTransformer.py
```diff
@@ -0,0 +1,242 @@
...
+
+class TestRatioTransformerInit(BaseNumericTransformerInitTests):
+    """Tests for init method in RatioTransformer."""
+
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "RatioTransformer"
+
+    @pytest.mark.parametrize("columns", (["a"], ["a", "b", "c"], None, "a"))
+    def test_columns_arg_errors(
```
liamholmes31: S: can you call this 'test_errors_if_not_two_columns' so that it's clearer :)

tests/numeric/test_RatioTransformer.py
```diff
@@ -0,0 +1,242 @@
...
+from tests import utils as u
+from tests.numeric.test_BaseNumericTransformer import (
+    BaseNumericTransformerInitTests,
+    BaseNumericTransformerTransformTests,
+)
+from tests.test_data import (
+    create_ratio_test_df,  # Ensure this function generates the correct test data
+)
+
+# from tests.utils import _handle_from_json, assert_frame_equal_dispatch, dataframe_init_dispatch
```
liamholmes31: N: can delete commented code here

tests/numeric/test_RatioTransformer.py
```diff
@@ -0,0 +1,242 @@
...
+    ):
+        args = minimal_attribute_dict[self.transformer_name].copy()
+        args["return_dtype"] = return_dtype
+        with pytest.raises(BeartypeCallHintParamViolation):
+            uninitialized_transformers[self.transformer_name](**args)
+
+
+class TestRatioTransformerTransform(BaseNumericTransformerTransformTests):
+    """Tests for transform method in RatioTransformer."""
+
```
liamholmes31, line 59: S: so all of the tests in the transform section need another parametrize:
```
@pytest.mark.parametrize('from_json', [True, False'])
```

and then inside the test, on the line before .transform is called need to add:
```
transformer=_handle_from_json(transformer, from_json)
```

and the same will apply to DifferenceTransformer tests

tests/numeric/test_RatioTransformer.py
```diff
@@ -0,0 +1,242 @@
...
+            nw.col("a_divided_by_b").cast(getattr(nw, return_dtype)),
+        )
+        transformed_df = transformed_df.select(expected_df.columns)
+        expected_df = expected_df.to_native()
+        transformed_df = transformed_df.to_native()
+
+        u.assert_frame_equal_dispatch(transformed_df, expected_df)
+
+
+class TestRatioTransformerSerialization:
```
liamholmes31: S: this test is inherited from base_tests so can delete :)

tests/numeric/test_RatioTransformer.py
```diff
@@ -0,0 +1,242 @@
...
+        transformed_df = nw.from_native(transformed_df)
+        expected_df = expected_df.with_columns(
+            nw.col("a_divided_by_b").cast(getattr(nw, return_dtype)),
+        )
+        transformed_df = transformed_df.select(expected_df.columns)
+        expected_df = expected_df.to_native()
+        transformed_df = transformed_df.to_native()
+
+        u.assert_frame_equal_dispatch(transformed_df, expected_df)
+
```
liamholmes31, line 184: S: can you add a test where column2 is 0 to check this behaviour (both single row and multi-row, can possibly just work into existing tests)

tests/numeric/test_RatioTransformer.py
```diff
@@ -0,0 +1,242 @@
...
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize("return_dtype", ["Float32", "Float64"])
+    def test_single_row(
```
liamholmes31, line 119: S: can you parametrize this one to test a few different single rows, including some with null input values as this is where weird behaviour often creeps in

tests/numeric/test_DifferenceTransformer.py
```diff
@@ -0,0 +1,149 @@
...
+
+class TestDifferenceTransformerInit(BaseNumericTransformerInitTests):
+    """Tests for init method in DifferenceTransformer."""
+
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "DifferenceTransformer"
+
+    @pytest.mark.parametrize("columns", (["a"], ["a", "b", "c"], None, "a"))
+    def test_columns_arg_errors(
```
liamholmes31: S: same comment on renaming

tests/numeric/test_DifferenceTransformer.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class TestDifferenceTransformerTransform(BaseNumericTransformerTransformTests):
+    """Tests for transform method in DifferenceTransformer."""
+
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "DifferenceTransformer"
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
```
liamholmes31, line 52: S: this is inherited so can delete

tests/numeric/test_DifferenceTransformer.py
```diff
@@ -0,0 +1,149 @@
...
+        expected_data = {
+            "a": [100, 200, 300],
+            "b": [80, 150, 200],
+            "a_minus_b": [20, 50, 100],
+        }
+        expected_df = u.dataframe_init_dispatch(expected_data, library)
+
+        u.assert_frame_equal_dispatch(transformed_df, expected_df)
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
```
liamholmes31, line 81: S: same comment on parametrising to test a few different single rows

tubular/numeric.py
hariinduu: Sorry, just changed

tubular/numeric.py
hariinduu: Deleted

tubular/numeric.py
hariinduu: Removed

tubular/numeric.py
hariinduu: Moved to types.py

tubular/numeric.py
hariinduu: Deleted

tubular/numeric.py
hariinduu: Done

tests/test_data.py
hariinduu: Makes sense. Moved into test files

tests/numeric/test_RatioTransformer.py
hariinduu: Renamed

tests/numeric/test_RatioTransformer.py
hariinduu: deleted

tests/numeric/test_RatioTransformer.py
hariinduu: Added

tests/numeric/test_RatioTransformer.py
hariinduu: deleted

tests/numeric/test_RatioTransformer.py
hariinduu: added into the existing tests

tests/numeric/test_RatioTransformer.py
hariinduu: done

tests/numeric/test_DifferenceTransformer.py
hariinduu: renamed

tests/numeric/test_DifferenceTransformer.py
hariinduu: done

tests/numeric/test_RatioTransformer.py
```diff
@@ -0,0 +1,209 @@
...
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize("return_dtype", ["Float32", "Float64"])
+    @pytest.mark.parametrize("from_json", [True, False])
+    @pytest.mark.parametrize(
+        "a_values, b_values, expected_division",
+        [
+            ([100], [80], [1.25]),
+            ([0], [80], [0]),
+            ([100], [0], [None]),
```
liamholmes31, line 114: Can you add two more cases:
```
([None], [10], [None]),
([10], [None], [None]),
```

tests/numeric/test_DifferenceTransformer.py
```diff
@@ -0,0 +1,157 @@
...
+    @pytest.mark.parametrize("from_json", [True, False])
+    @pytest.mark.parametrize(
+        "a_values, b_values, expected_value",
+        [
+            ([100], [80], 20),
+            ([0], [80], [-80]),
+            ([100], [0], [100]),
```
liamholmes31, line 88: same comment here :)

tubular/numeric.py
```diff
@@ -1320,3 +1321,235 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+        columns: ListOfTwoStrs,
+        **kwargs: Optional[bool],
+    ) -> None:
+        """Initialize the DifferenceTransformer.
+
+        Parameters
+        ----------
+        columns : ListOfTwoStrs
+            List of exactly two column names to operate on. The second column is subtracted from the first.
+            and the second column is the subtrahend/denominator.
```
liamholmes31: can  delete this line  of docstring

liamholmes31: (1361)

tubular/numeric.py
```diff
@@ -1320,3 +1321,235 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
...
+    """Transformer that performs subtraction operation between two columns.
+
+    This transformer allows performing subtraction between two columns in a DataFrame
+    and stores the result in a new column.
+
+    Attributes
+    ----------
+    columns : ListOfTwoStrs
+        List of exactly two column names to operate on. The second column is subtracted from the first.
+        and the second column is the subtrahend.
```
liamholmes31: can delete line 1336

tests/numeric/test_DifferenceTransformer.py
```diff
@@ -0,0 +1,157 @@
...
+        transformed_df = transformer.transform(single_row_df)
+
+        # Expected output for a single-row DataFrame
+        expected_data = {
+            "a": a_values,
+            "b": b_values,
+            "a_minus_b": expected_value,
+        }
+        expected_df = u.dataframe_init_dispatch(expected_data, library)
+
```
liamholmes31, line 143: will need to add a cast here so it works with null rows:
# ensure null rows are numeric type (not Unknown type)
expected_df=nw.from_native(expected_df).with_columns(nw.col(a'').cast(nw.Float64), nw.col('b').cast(nw.Float64))

tubular/numeric.py
hariinduu: Deleted

tubular/numeric.py
hariinduu: deleted

Issue #513, opened 2025-09-26
include functionality to filter down pipelines


liamholmes31:
What?
Generally there will be differences between the pipeline that goes into feature selection and the one that goes into the final model, but would be nice to programmatically convert feature selection pipelines into final model pipelines without just explicitly redefining.

Some features in tubular would enable this:
- easy to access transformer inputs (we have this with the columns attr)
- easy to access transformer outputs - this is less consistent in tubular, suggest we could mirror the sklearn API and include a `get_feature_names_out` method as standard
- transformer specific methods for dropping unneeded input args, e.g. a `remove_unused_features` method

We could then, given a pipeline, reverse it and step through where for each transformer:
- we compare the outputs to a list of needed columns (this will be initialised as the model features)
- where there is overlap we add the relevant inputs to needed columns
- where there is not overlap, we use the new drop unneeded inputs args method to edit the transformer
- we then continue to the next step

Why?
enable one defined pipeline to service many different model iterations

How?
This will involve lots of smaller issues, would want to split into two tasks per transformer API (transformers  with a shared API will be able to share methods):
- add `get_feature_names` method  
- add `remove_unused_features` method

PR #545, opened 2025-10-14
added 'get_feature_names_out' method to all transformer classes

liamholmes31:
Added 'get_feature_names_out' method (matches sklearn API) which lists columns edited/created by the transformer. Encountered a few cases where not being able to get this easily was a pain, also is a first step for the below:
https://github.com/azukds/tubular/issues/513

as this one is quite simple, felt like the doctests were sufficient, but can add more cases on request!

### DIFF
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -1376,7 +1376,7 @@ class ToFromJsonTests:
             transformer.from_json({})
 
 
-class TestGetFeatureNamesOut:
+class GetFeatureNamesOutTests:
     """
     Tests for the BaseTransformer.get_feature_names_out method.
     Note this deliberately avoids starting with "Tests" so that the tests are not run on import.
@@ -1421,7 +1421,7 @@ class TestGetFeatureNamesOut:
 class OtherBaseBehaviourTests(
     ColumnsCheckTests,
     CombineXYTests,
-    TestGetFeatureNamesOut,
+    GetFeatureNamesOutTests,
     ToFromJsonTests,
 ):
     """


pyproject.toml
```diff
@@ -165,3 +165,6 @@ fail_under = 80
 
 # Additional Options (AddOpts) for pytest
 addopts = '--doctest-modules'
+filterwarnings = [
```
liamholmes31, line 168: added this as warnings were cluttering the terminal since we deprecated lots of transformers

tubular/aggregations.py
```diff
@@ -230,6 +230,29 @@ def __init__(
...
+        ... columns='a',
+        ... aggregations=['min', 'max'],
+        ... key='b',
+        ... )
+
+        >>> transformer.get_feature_names_out()
+        ['a_min', 'a_max']
+        """
+
+        return [f"{col}_{agg}" for col in self.columns for agg in self.aggregations]
```
SadieRobertson, line 294: S: Could we add a test that this function output matches the added column names added when transform is applied to a dataframe? This will ensure this function does not get out of sync with any changes to transform

tubular/nominal.py
```diff
@@ -1354,6 +1424,61 @@ def __init__(
...
+
+        >>> test_df=pl.DataFrame({'a': ['cat', 'dog', 'rat']})
+
+        >>> _ = transformer.fit(test_df)
+
+        >>> transformer.get_feature_names_out()
+        ['a_cat', 'a_dog', 'a_rat']
+        """
+
+        # if wanted leves is not provided, this function
```
SadieRobertson: N: typo, but should be wanted values?

tubular/numeric.py
```diff
@@ -251,6 +251,33 @@ def __init__(
...
+        ... drop_original=False,
+        ... kmeans_kwargs={"random_state": 42},
+        ...  )
+
+        >>> transformer.get_feature_names_out()
+        ['kmeans_column']
+        """
+
+        return [
+            self.new_column_name,
```
SadieRobertson, line 298: S: A couple of transformers use new_column_name, could pass it to the kwargs and have the base class return the new column name if present? Would avoid some repitition, but it looks like it is only a couple of transformers. What do you think?

tubular/aggregations.py
SadieRobertson: I think this test is mostly required for the transformers that have updated the get_feature_names_out() method in this PR. However it could be a useful general check for all of the transform methods, but that would be adding a lot of tests!

tubular/numeric.py
liamholmes31: have actually been thinking about just removing new_column_name from the repo generally, so can we spin this out into discussion on the below?

 https://github.com/azukds/tubular/issues/559

tubular/nominal.py
liamholmes31: thanks, updated here https://github.com/azukds/tubular/pull/545/commits/a49a5e8689333cccee8546dadefe927d37f4c4e2

tubular/aggregations.py
liamholmes31: managed to do this without toooo many extra tests (in the base test class so it gets inherited around)

I did end up splitting over a few commits, so may be easiest to just look at file diffs again?

tests/base_tests.py
```diff
@@ -1376,9 +1376,52 @@ def test_from_json_blocked_for_non_jsonable_transformer(
             transformer.from_json({})
 
 
+class TestGetFeatureNamesOut:
```
SadieRobertson: N: GetFeatureNamesOutTests would fit the naming used throughout this script and make docstring less confusing

tests/nominal/test_GroupRareLevelsTransformer.py
```diff
@@ -403,4 +403,4 @@ class TestOtherBaseBehaviour(OtherBaseBehaviourTests):
 
     @classmethod
     def setup_class(cls):
-        cls.transformer_name = "BaseNominalTransformer"
```
SadieRobertson, line 406: Q: Was this a bug before the update?

tubular/aggregations.py
SadieRobertson: Looks good!

tubular/numeric.py
SadieRobertson: Have added a comment

tests/base_tests.py
liamholmes31: thanks, updated here https://github.com/azukds/tubular/pull/545/commits/d84909982125e9b92eb8a48d0a5e6245f5be0732

tests/nominal/test_GroupRareLevelsTransformer.py
liamholmes31: yes this was, it was just looking and and running the basenominal class tests again before!

Issue #547, opened 2025-10-14
turn on DOC ruff lints


liamholmes31:
What?
this lint set ensures that docstrings describes all possible returns/error messages

Why?
keep docstrings thorough

How?
add "DOC" to the list of ruff rules in pyproject.toml and fix resulting errors - this one may take a while, so might want to split out file by file. Could do this by adding other files to the per-files-ignores list

Issue #115, opened 2023-07-11
apply ruff ruleset D  pydocstyle


davidhopkinson26:
https://beta.ruff.rs/docs/rules/#pydocstyle-d

make code compliant with rule set and then add to ruleset in .ruff.toml.  If appropriate identify specific rules or parts of code to exclude.

Found 314 errors.

ClaireF57:
May I be assigned this task ?  @Sarah-TaylorKnight  @davidhopkinson26

PR #567, opened 2025-10-20
Feature/ruff docs

liamholmes31:
began turning on D/DOC lints (these have many errors so going for a file by file approach)


tubular/base.py
```diff
@@ -465,14 +516,10 @@ def columns_check(self, X: DataFrame) -> None:
         >>> df=pl.DataFrame({'a': [1,2], 'b': [3,4]})
 
         >>> transformer.columns_check(df)
-        """
 
+        """
         X = _convert_dataframe_to_narwhals(X)
 
-        if not isinstance(self.columns, list):
```
liamholmes31, line 472: turned this off and removed test rather than adding the error to docstring, as the check is on an attr which we control it's kinda pointless imo

PR #569, opened 2025-10-22
added to_json method for one hot encoder

liamholmes31:
added to_json method and integrated into tests

https://github.com/azukds/tubular/issues/550

### DIFF
--- a/tests/nominal/test_OneHotEncodingTransformer.py
+++ b/tests/nominal/test_OneHotEncodingTransformer.py
@@ -375,9 +375,9 @@ class TestTransform(
         transformer.fit(df_train)
         transformer2.fit(df_train)
 
-        transformer.transform(df_test)
-
         transformer = _handle_from_json(transformer, from_json)
+
+        transformer.transform(df_test)
 
         assert transformer2.categories_ == transformer.categories_, (
             f"categories_ modified during transform, pre transform had {transformer2.categories_} but post transform has {transformer.categories_}"


tubular/nominal.py
```diff
@@ -1819,6 +1875,7 @@ def _warn_missing_levels(
 
         return missing_levels
 
+    @block_from_json
```
Ritunjai-Sharma: I know we wanted to block fitting a transformer reconstructed from JSON (hence the @block_from_json on fit method). Just wanted to double check if we indeed want to block get_feature_names_out from working on the transformer reconstructed from JSON as well?

tests/nominal/test_OneHotEncodingTransformer.py
```diff
@@ -366,16 +377,19 @@ def test_categories_not_modified(library):
 
         transformer.transform(df_test)
 
+        transformer = _handle_from_json(transformer, from_json)
```
Ritunjai-Sharma, line 378: _handle_from_json check should be before the transformer.transform()?

tubular/nominal.py
liamholmes31: ah thanks didn't mean to add that! https://github.com/azukds/tubular/pull/569/commits/1a8f1272f1cad2d5f754144ea53e9acbf6ffbf65

tests/nominal/test_OneHotEncodingTransformer.py
liamholmes31: whoops, thanks RJ https://github.com/azukds/tubular/pull/569/commits/964d50aacb7967989f7ab80c1cdad28ad23a1405

PR #570, opened 2025-10-22
added to_json method for DateDifferenceTransformer

liamholmes31:
added to_json method and integrated into tests 

https://github.com/azukds/tubular/issues/552


tubular/dates.py
```diff
@@ -587,8 +598,6 @@ def __init__(
             "m",
             "s",
         ] = "D",
-        copy: bool = False,
```
Ritunjai-Sharma, line 544: Q: Why have the copy and verbose parameters been removed here?

tubular/dates.py
```diff
@@ -616,8 +625,6 @@ def __init__(
             columns=columns,
             new_column_name=new_column_name,
             drop_original=drop_original,
-            copy=copy,
```
Ritunjai-Sharma, line 557: Q: I'm assuming that the parameters copy and verbose are being removed here since these were removed from the parent class BaseDateTwoColumnTransformer at some point but weren't removed here?

tests/dates/test_DateDifferenceTransformer.py
```diff
@@ -431,8 +440,11 @@ def test_expected_output_nulls(self, df, expected):
 
         df_transformed = x.transform(df)
 
+        x = _handle_from_json(x, from_json)
```
Ritunjai-Sharma, line 419: Shouldn't `x = _handle_from_json(x, from_json)` come before `df_transformed = x.transform(df)` since we want to test if converting and loading from json transforms the dataframe similar to freshly initialized transformer?

tests/dates/test_DateDifferenceTransformer.py
```diff
@@ -457,6 +469,8 @@ def test_expected_output_nulls2(self, df, expected):
...
```
Ritunjai-Sharma, line 448: Similar to above comment, Shouldn't `x = _handle_from_json(x, from_json)` come before `df_transformed = x.transform(df)` since we want to test if converting and loading from json transforms the dataframe similar to freshly initialized transformer?

tubular/dates.py
liamholmes31: removed these as the general approach is to bundle these into **kwargs - they are rarely used so they clutter up the args list a bit otherwise!

tubular/dates.py
liamholmes31: same comment around bundling into **kwargs

tests/dates/test_DateDifferenceTransformer.py
liamholmes31: ah good catch, moved here https://github.com/azukds/tubular/pull/570/commits/68777eb6ca70af6727beac82c67a11a6b23a4c19

tests/dates/test_DateDifferenceTransformer.py
liamholmes31: ah good catch, moved here https://github.com/azukds/tubular/commit/68777eb6ca70af6727beac82c67a11a6b23a4c19

PR #572, opened 2025-10-23
added to_json method for DatetimeInfoExtractor

liamholmes31:
- added to_json method and integrated into tests
- moved non-argument attrs created in init to transform, as otherwise does not work
- removed duplicated is_serialisable test (inherited from base tests)
- removed use of MappingTransformer, as was not actually adding anything

https://github.com/azukds/tubular/issues/555

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,7 @@ Changed
 - chore: set `jsonable` class attribute to True for BaseCappingTransformer  `#540 <https://github.com/azukds/tubular/pull/594>`_
 - feat: switched pre-commit for `prek <https://github.com/j178/prek>_` `#634 <https://github.com/azukds/tubular/issues/634>`_
 - feat: added `to_json` method for DatetimeInfoExtractor `#555 <https://github.com/azukds/tubular/issues/555>`_
+- bugfix: fixed unhandled error for nulls on GroupRareLevelsTransformer.fit
 - feat: added `to_json` method for OneHotEncodingTransformer `#550 <https://github.com/azukds/tubular/issues/550>`_
 - turned on ruff PLR rule `#509 <https://github.com/azukds/tubular/issues/509>`_
 - feat: added auto generated feature table to readme, to track transform support for e.g. polars/lazyframes/json `#595 <https://github.com/azukds/tubular/issues/595>`_
--- a/tests/nominal/test_GroupRareLevelsTransformer.py
+++ b/tests/nominal/test_GroupRareLevelsTransformer.py
@@ -161,6 +161,22 @@ class TestFit(GenericFitTests, WeightColumnFitMixinTests, DummyWeightColumnMixin
             f"non_rare_levels attribute not fit as expected, expected {expected} but got {actual}"
         )
 
+    @staticmethod
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_nulls_error(library):
+        """Test that checks error is raised if transform is run on column with nulls."""
+        df_dict = {"a": ["a", None]}
+        df = dataframe_init_dispatch(dataframe_dict=df_dict, library=library)
+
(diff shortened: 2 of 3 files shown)


tubular/dates.py
```diff
@@ -1223,7 +1227,7 @@ def __init__(
         include: Optional[Union[DatetimeInfoOptionList, DatetimeInfoOptionStr]] = None,
         datetime_mappings: Optional[dict[DatetimeInfoOptionStr, dict[int, str]]] = None,
         drop_original: Optional[bool] = False,
-        **kwargs: bool,
+        **kwargs: Union[str, bool],
```
Ritunjai-Sharma, line 1230: Why was str included here?

tubular/dates.py
```diff
@@ -1278,32 +1290,44 @@ def __init__(
...
+        {'tubular_version': ..., 'classname': 'DatetimeInfoExtractor', 'init': {'columns': ['a'], 'copy': False, 'verbose': False, 'return_native': True, 'new_column_name': 'dummy', 'drop_original': False, 'include': ['timeofday', 'timeofmonth', 'timeofyear', 'dayofweek'], 'datetime_mappings': {}}, 'fit': {}}
+
+        """
+        json_dict = super().to_json()
+
+        json_dict["init"].update(
+            {
+                "include": self.include,
+                "datetime_mappings": self.datetime_mappings,
+                "drop_original": self.drop_original,
```
Ritunjai-Sharma: Wouldn't drop_original be added as part of super().to_json()?

tubular/dates.py
```diff
@@ -1358,26 +1382,15 @@ def _process_provided_mappings(
...
 
-        for include_option in self.include:
             # check provided mappings fit required format
-            if (
-                set(self.final_datetime_mappings[include_option].keys())
-                != self.RANGE_TO_MAP[include_option]
-            ):
-                msg = f"{self.classname()}: {include_option.value} mapping dictionary should contain mapping for all values between {min(self.RANGE_TO_MAP[include_option])}-{max(self.RANGE_TO_MAP[include_option])}. {self.RANGE_TO_MAP[include_option] - set(self.final_datetime_mappings[include_option].keys())} are missing"
+            if set(datetime_mappings[key].keys()) != self.RANGE_TO_MAP[key]:
+                msg = f"{self.classname()}: {key} mapping dictionary should contain mapping for all values between {min(self.RANGE_TO_MAP[key])}-{max(self.RANGE_TO_MAP[key])}. {self.RANGE_TO_MAP[key] - set(datetime_mappings[key].keys())} are missing"
```
Ritunjai-Sharma: Looks like this is out of the above for loop. So, only the last mapping key would get validated. Should be an easy fix by moving the range check inside the loop.

tubular/dates.py
liamholmes31: ah this is because of new_column_name, even though it isn't used it is set in the base class, so it gets returned by to_json and fed into from_json, so init has to be able to handle it being passed

tubular/dates.py
liamholmes31: good catch https://github.com/azukds/tubular/pull/572/commits/2a7f0a5327936f537139bf167732709b1970bc63

tubular/dates.py
liamholmes31: ah thanks!! https://github.com/azukds/tubular/pull/572/commits/4c65113e05b36fd31083fce686fd56a370fed873

utils/feature_table.py
```diff
@@ -2,12 +2,13 @@
 
 import pathlib
 import re
+import sys
 
 import pandas as pd
 
-from tubular.base import FEATURE_REGISTRY
+sys.path.append(str(pathlib.Path(__file__).resolve().parents[1]))
```
liamholmes31, line 9: this is from another PR, but the import mysteriously started failing (it passed when the other PR went in), so have just done a quick fix here

Issue #542, opened 2025-10-10
add to_json method for SetValueTransformer


liamholmes31:
What?
want to make it so that transformers can be dumped to/built from json - to do this in this case just need to add a to_json method which:

- calls super.to_json()
- adds the additional init attrs (`value`) to this dict
- Then need to add test cases for transforms with to/from json-ed transformers - this can be done fairly easily with parametrise and following the example in this PR https://github.com/azukds/tubular/pull/496

Will need to:
- set the jsonable class attr to True
- add the `block_from_json` decorator to any instance methods that are not needed for transform
- set up and transform tests for this class to do an additional run for json'd transformers, following the PR example

Why?
Provide a more readable and env independent way of storing pipelines/transformers

liamholmes31:
this PR may be a useful example to follow https://github.com/azukds/tubular/pull/566

PR #581, opened 2025-11-02
Issue #542 - added 'to_json' method for 'SetValueTransformer'

KatheeO:


tubular/misc.py
```diff
@@ -74,8 +76,31 @@ def __init__(
...
+
+
+        """
+        json_dict = super().to_json()
+
+        json_dict["init"]["value"] = self.value
+
+        return json_dict
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
```
liamholmes31: for the type hint, could you import DataFrame from tubular.types and use this here :) is more flexible to include all of pd/pl/nw DataFrames

tubular/misc.py
```diff
@@ -74,8 +76,31 @@ def __init__(
...
+        -------
+        dict[str, dict[str, Any]]:
+            jsonified transformer. Nested dict containing levels for attributes
+            set at init and fit.
+
+        Examples
+        --------
+        >>> transformer = SetValueTransformer(columns='a', value=1)
+        >>> transformer.to_json()
+        {'tubular_version': 'dev', 'classname': 'SetValueTransformer', 'init': {'columns': ['a'], 'copy': False, 'verbose': False, 'return_native': True, 'value': 1}, 'fit': {}}
```
liamholmes31: could you change `'dev'` to `...` here? tubular version will be detected differently depending on whether the code is run locally or in CI, so this is more flexible and will pass in both contexts :)
```suggestion
        {'tubular_version': ..., 'classname': 'SetValueTransformer', 'init': {'columns': ['a'], 'copy': False, 'verbose': False, 'return_native': True, 'value': 1}, 'fit': {}}
```

tubular/misc.py
```diff
@@ -112,10 +137,14 @@ def transform(self, X: FrameT) -> FrameT:
         └─────┴─────┘
 
         """
-        X = nw.from_native(super().transform(X))
+        X = super().transform(X, return_native_override=False)
+
+        X = _convert_dataframe_to_narwhals(X)
+
+        for c in self.columns:
```
liamholmes31: thanks for tidying up this method! but could you change these two lines to:
```
X = X.with_columns([nw.lit(self.value).alias(col) for col in self.columns])
```

it's more efficient to build up all the expressions and just call .with_columns once, as each call of this will materialise a new df :)

tests/misc/test_SetValueTransformer.py
```diff
@@ -58,13 +61,17 @@ def setup_class(cls):
...
-    def test_value_set_in_transform(self, library, value):
+    @pytest.mark.parametrize("from_json", [True, False])
+    def test_value_set_in_transform(self, library, value, from_json):
         """Test that transform sets the value as expected."""
 
         df = d.create_df_2(library)
 
         x = SetValueTransformer(columns=["a", "b"], value=value)
 
+        if from_json:
```
liamholmes31: if you import _handle_from_json from tests.utils, you can do:
```
x=_handle_from_json(x, from_json)
```
 which will have the same effect here :)

tests/misc/test_SetValueTransformer.py
```diff
@@ -74,6 +81,28 @@ def test_value_set_in_transform(self, library, value):
             df2=expected,
         )
 
+    @pytest.mark.parametrize("value", ["a", 1, 1.0, None, np.nan])
+    def test_to_json_returns_correct_dict(self, value):
```
liamholmes31: this test would sit better in the OtherBaseBehaviour class :)

tests/misc/test_SetValueTransformer.py
```diff
@@ -74,6 +81,28 @@ def test_value_set_in_transform(self, library, value):
...
+        """Test that to_json is working as expected."""
+        transformer = SetValueTransformer(columns="a", value=value)
+
+        actual = transformer.to_json()
+
+        expected = {
+            "tubular_version": "dev",
```
liamholmes31: this part will fail in the CI context, I would suggest maybe doing something like:
```
assert isinstance(actual["tubular_version"], str), 'expected tubular version to be captured as str in to_json'
del actual["tubular_version"]
```
and then proceeding as you have but without 'tubular_version' in `expected`

we actually haven't been explicitly testing to_json elsewhere, and have been treating it as implicitly tested by the combined to/from_json tests - but happy to keep in your test as an extra check!

KatheeO:
Thank you for the review, the changes you suggested are great and make sense. It will definitely teach me some things :) One thing: I moved the feature changes to a dedicated branch on my fork, since I previously pushed straight to main, so I will make a new PR from that dedicated branch and add a link to this one.

PR #584, opened 2025-11-05
removed SeparatorColumnMixin

liamholmes31:
removed this mixin class - as lots of work has already been done on beartyping, this had no impact on tests/was a very easy change

https://github.com/azukds/tubular/issues/562

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,7 +33,7 @@ Each individual change should have a link to the pull request after the descript
 Changed
 ^^^^^^^
 
-- placeholder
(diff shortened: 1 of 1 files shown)


tubular/nominal.py
```diff
@@ -1468,7 +1467,7 @@ def __init__(
 
         self.wanted_values = wanted_values
         self.drop_original = drop_original
-        self.check_and_set_separator_column(separator)
+        self.separator = separator
```
Boluwatife28, line 1470: Looks good! Just to confirm, the SeparatorColumnMixin isn't needed anymore because seperator is now passed as an argument and type-checked with beartype right?

liamholmes31: yeah exactly :)

PR #589, opened 2025-11-05
Revised: Issue #542 - added 'to_json' method for 'SetValueTransformer'

KatheeO:
Continued work from PR #581 with revisions

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,6 +34,7 @@ Changed
 ^^^^^^^
 
 - feat: added `to_json` method for `SetValueTransformer` `#542 <https://github.com/azukds/tubular/issues/542>`_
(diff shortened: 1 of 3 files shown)


tests/misc/test_SetValueTransformer.py
```diff
@@ -58,13 +61,17 @@ def setup_class(cls):
...
-    def test_value_set_in_transform(self, library, value):
+    @pytest.mark.parametrize("from_json", [True, False])
+    def test_value_set_in_transform(self, library, value, from_json):
         """Test that transform sets the value as expected."""
 
         df = d.create_df_2(library)
 
         x = SetValueTransformer(columns=["a", "b"], value=value)
 
+        if from_json:
```
liamholmes31, line 73: fyi, _handle_from_json doesn't need the if statement, but is harmless so can leave here

PR #591, opened 2025-11-07
Include option for DatetimeInfoExtractor to return unmapped values

hariinduu:
Added an option to DatetimeInfoExtractor to return unmapped values if specified

@488

### DIFF
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -1067,9 +1067,6 @@ class DatetimeInfoExtractor(BaseDatetimeTransformer):
     drop_original: str
         indicates whether to drop provided columns post transform
 
-    return_unmapped_values: bool, default = False
-        If True, returns raw datetime attribute values without mapping.
-
     **kwargs
         Arbitrary keyword arguments passed onto BaseTransformer.init method.
 
@@ -1168,9 +1165,11 @@ class DatetimeInfoExtractor(BaseDatetimeTransformer):
         include: Optional[Union[DatetimeInfoOptionList, DatetimeInfoOptionStr]] = None,
         datetime_mappings: Optional[dict[DatetimeInfoOptionStr, dict[int, str]]] = None,
         drop_original: Optional[bool] = False,
-        return_unmapped_values: Optional[bool] = False,
         **kwargs: dict[str, bool],
     ) -> None:
+        
+        self.final_datetime_mappings = copy.deepcopy(self.DEFAULT_MAPPINGS)
+        
         if include is None:
             include = self.INCLUDE_OPTIONS
 
@@ -1186,7 +1185,6 @@ class DatetimeInfoExtractor(BaseDatetimeTransformer):
 
         self.include = include
         self.datetime_mappings = datetime_mappings
-        self.return_unmapped_values = return_unmapped_values
         self._process_provided_mappings(datetime_mappings=datetime_mappings)
 
         # this is a situation where we know the values our mappings allow,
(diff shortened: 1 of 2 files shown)


tubular/dates.py
```diff
@@ -1325,30 +1334,56 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+
+        for col in self.columns:
+            for include_option in self.include:
+                # Extract raw datetime attribute and give it an alias
+                unmapped_dt_dict[col + "_" + include_option] = getattr(
                     nw.col(col).dt,
                     self.DATETIME_ATTR[include_option],
-                )().replace_strict(
-                    self.mapping_transformer.mappings[col + "_" + include_option],
+                )().alias(col + "_" + include_option)
```
liamholmes31: S: will be helpful to use different aliases for these two or  they will overwrite each other, e.g. `col + "_" + include_option +"_" + unmapped` for the first

tubular/dates.py
```diff
@@ -1325,30 +1334,56 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+            return_dict.update(unmapped_dt_dict)
+        if any(self.final_datetime_mappings[include_option] for include_option in self.include):
+            return_dict.update(mapped_dt_dict)
+
+        # Final casts
+        if return_unmapped_values:
+            # Cast to Float64 to accommodate NaN values
+            return_dict = {
+                col + "_" + include_option: return_dict[col + "_" + include_option].cast(
+                    nw.Float64,
```
liamholmes31: S: also, would cast to nw.Int8 - month<=12, day<=31, hour<=24, weekday<=7 so should all be OK

liamholmes31: looks like this is failing due to nulls, so ignore me and would change back to float!

tubular/dates.py
```diff
@@ -1323,34 +1332,87 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+
         X = super().transform(X, return_native_override=False)
 
-        transform_dict = {
-            col + "_" + include_option: (
-                getattr(
+        print("DataFrame after base transformation:")
+        print(X)
+
+        unmapped_dt_dict = {}
```
liamholmes31: S: could do this in one dict, as the expressions all have different keys anyway

tubular/dates.py
```diff
@@ -1323,34 +1332,87 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+        for col in self.columns:
+            for include_option in self.include:
+                # Extract raw datetime attribute and give it an alias
+                unmapped_alias = col + "_" + include_option + "_unmapped"
+                unmapped_dt_dict[unmapped_alias] = getattr(
                     nw.col(col).dt,
                     self.DATETIME_ATTR[include_option],
-                )().replace_strict(
-                    self.mapping_transformer.mappings[col + "_" + include_option],
+                )().alias(unmapped_alias)
```
liamholmes31: S: could also do the casting on the end here, and remove the extra if/else statement below

tubular/dates.py
hariinduu: Added

tubular/dates.py
```diff
@@ -1323,34 +1332,87 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+
+                # Apply mapping transformation if mappings are provided
+                if self.final_datetime_mappings[include_option]:
+                    mapped_alias = col + "_" + include_option
+                    mapped_dt_dict[mapped_alias] = (
+                        unmapped_dt_dict[unmapped_alias]
+                        .replace_strict(
+                            self.mapping_transformer.mappings[mapped_alias],
+                        )
+                        .alias(mapped_alias)
```
liamholmes31: same comment re casting

tubular/dates.py
hariinduu: Changed back to float

tubular/dates.py
```diff
@@ -1164,6 +1168,7 @@ def __init__(
         include: Optional[Union[DatetimeInfoOptionList, DatetimeInfoOptionStr]] = None,
         datetime_mappings: Optional[dict[DatetimeInfoOptionStr, dict[int, str]]] = None,
         drop_original: Optional[bool] = False,
+        return_unmapped_values: Optional[bool] = False,
```
liamholmes31: S: I think actually to keep things simpler, let's drop this argument and just have the transformer always return both mapped/unmapped values

liamholmes31:
closing, as pivoted approach to https://github.com/azukds/tubular/pull/603

PR #594, opened 2025-11-09
add to_json method to BaseCappingTransformer

svojt:
Added a to_json method to BaseCappingTransformer to support JSON serialization of init and fitted attributes.
Tests ensure that a transformer reconstructed from JSON works as expected.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -36,20 +36,21 @@ Changed
 - feat: added tests for BaseCappingTransformer to verify correct behaviour when using transformers rebuilt from JSON `#540 <https://github.com/azukds/tubular/pull/594>_`
 - chore: set `jsonable` class attribute to True for BaseCappingTransformer  `#540 <https://github.com/azukds/tubular/pull/594>_`
 
+- switched pre-commit for `prek <https://github.com/j178/prek>_` `#634 <https://github.com/azukds/tubular/issues/634>`_
 
 2.4.0 (01/12/2025)
 ------------------
 
 Changed
 ^^^^^^^
-- turned on DOC lints for remaining classes `#547 <https://github.com/azukds/tubular/issues/547>_`
-- turned on D lints for remaining classes `#115 <https://github.com/azukds/tubular/issues/115>_`
+- turned on DOC lints for remaining classes `#547 <https://github.com/azukds/tubular/issues/547>`_
+- turned on D lints for remaining classes `#115 <https://github.com/azukds/tubular/issues/115>`_
 - copied optimisations/improvements from #484 into MedianImputer
-- feat: added `to_json` method for MeanResponseTransformer `#549 <https://github.com/azukds/tubular/issues/549>_`
+- feat: added `to_json` method for MeanResponseTransformer `#549 <https://github.com/azukds/tubular/issues/549>`_
 - feat: added `to_json` method for `DatetimeSinusoidCalculator` `#556 <https://github.com/azukds/tubular/issues/556>`_
 - feat: added `to_json` method for `BetweenDatesTransformer` `#554 <https://github.com/azukds/tubular/issues/554>`_
 - feat: added a new transformer called DatetimeComponentExtractor
-- feat: add a decorator to register transformer classes by name `#588 <https://github.com/azukds/tubular/issues/588>_`
+- feat: add a decorator to register transformer classes by name `#588 <https://github.com/azukds/tubular/issues/588>`_
 - bugfix: MappingTransformer was hitting a recursion error for many mappings, adjusted logic to avoid this
 
 2.3.0 (18/11/2025)
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -12,7 +12,7 @@
 	// "forwardPorts": [],
 
 	// Use 'postCreateCommand' to run commands after the container is created.
-	"postCreateCommand": "pip3 install --user -r requirements-dev.txt; pre-commit install",
+	"postCreateCommand": "pip3 install --user -r requirements-dev.txt; prek install",
 
 	// Configure tool-specific properties.
 	"customizations": {
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -30,15 +30,15 @@ Then install tubular and dependencies for development;
 
      pip install . -r requirements-dev.txt
 
-We use `pre-commit <https://pre-commit.com/>`_ for this project which is configured to check that code is formatted with `black <https://black.readthedocs.io/en/stable/>`_ and passes `ruff <https://beta.ruff.rs/docs/>`_ checks.  For a list of ruff rules follwed by this project check .ruff.toml.
+We use `prek <https://github.com/j178/prek>`_ for this project which is configured to check that code passes `ruff <https://beta.ruff.rs/docs/>`_ checks.  For a list of ruff rules follwed by this project check pyproject.toml.
 
-To configure ``pre-commit`` for your local repository run the following;
+To configure ``prek`` for your local repository run the following;
 
    .. code::
 
-     pre-commit install
+     prek install
 
-If working in a codespace the dev requirements and precommit will be installed automatically in the dev container.
+If working in a codespace the dev requirements and prek will be installed automatically in the dev container.
 
 If you are building the documentation locally you will need the `docs/requirements.txt <https://github.com/azukds/tubular/blob/main/docs/requirements.txt>`_.
 
@@ -70,9 +70,9 @@ General
 Code formatting
 ^^^^^^^^^^^^^^^
 
-We use `black <https://black.readthedocs.io/en/stable/>`_ to format our code and follow `pep8 <https://www.python.org/dev/peps/pep-0008/>`_ conventions. 
+We use `ruff <https://beta.ruff.rs/docs/>`_ to format our code.
 
-As mentioned above we use ``pre-commit`` which streamlines checking that code has been formatted correctly.
+As mentioned above we use ``prek`` which streamlines checking that code has been formatted correctly.
 
 CI
 ^^
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -12,5 +12,5 @@ pytest>=5.4.1
 pytest-mock>=3.5.1
 pyarrow>=17.0.0
 pytest-cov>=2.10.1
-pre-commit==2.15.0
+prek==0.2.19
 ruff==0.2.2
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dev = [
     "pytest-mock>=3.5.1",
     "pyarrow>=17.0.0",
     "pytest-cov<=2.10.1",
-    "pre-commit<=6.1.1", # pytest was unable to run for 6.2.0, 6.2.1
+    "prek",
     "ruff>=0.2.2",
     "auto-walrus==0.3.4",
     ]
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,26 +4,16 @@ auto-walrus==0.3.4
     # via tubular (pyproject.toml)
 beartype==0.21.0
     # via tubular (pyproject.toml)
-cfgv==3.4.0
-    # via pre-commit
 coverage==7.10.7
     # via pytest-cov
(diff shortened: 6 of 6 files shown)


tubular/capping.py
```diff
@@ -585,6 +587,25 @@ def transform(
...
+        data = super().to_json()
+        
+        data["init"].pop("columns", None)
+        data["init"].update({
+            "capping_values": self.capping_values,
+            "quantiles": self.quantiles,
+            "weights_column": self.weights_column,
+        })
+        
+        if hasattr(self, "quantile_capping_values"):
```
liamholmes31: It looks like `quantile_capping_values`/`replactement_values` will always exist after fit, so I would drop the if statements and do:
```
data["fit"].update({'quantile_capping_values': self.quantile_capping_values, '_replacement_values': self._replacement_vaues})
```

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -853,3 +855,217 @@ def test_expected_output(
         assert actual_rounded_1_dp == expected_quantiles, (
             "unexpected weighted quantiles calculated"
         )
+
+
+def _handle_from_json_capping(transformer, from_json=True):
```
liamholmes31: you can import `_handle_from_json`  from tests.utils and use that instead of redefining here :)

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -853,3 +855,217 @@ def test_expected_output(
...
+        [
+            # Test with capping_values only
+            ({"a": [2, 8], "b": [1, 5]}, None, None),
+            ({"a": [None, 10], "b": [0, None]}, None, None),
+            # Test with quantiles only
+            (None, {"a": [0.1, 0.9], "b": [0.05, 0.95]}, None),
+            (None, {"a": [0.2, 0.8]}, "weight"),
+        ],
+    )
+    def test_to_from_json_transform_consistency(
```
liamholmes31: the tests below look good, but are not following the format we're aiming for in the repo. We want to make use of the parametrize decorator to reuse the existing tests, so the goal is to convert the tests in GenericCappingTransformTests to do this. Will show the first test below as an example:

```
# ADDED line
@pytest.mark.parametrize("from_json", [True, False])
@pytest.mark.parametrize("library", ["pandas", "polars"])
def test_non_cap_column_left_untouched(
        self,
        initialized_transformers,
        library,
        from_json, # ADDED line
    ):
        """Test that capping is applied only to specific columns, others remain the same."""
        df = d.create_df_4(library=library)
        expected = self.expected_df_2(library=library)
        transformer = initialized_transformers[self.transformer_name]
        # if transformer is not polars compatible, skip polars test
        if not transformer.polars_compatible and isinstance(df, pl.DataFrame):
            return
  
        transformer=_handle_from_json(transformer, from_json) # ADDED line

        transformer.fit(df)
        df_transformed = transformer.transform(df)
        # exclude transformed columns for this test
        columns_to_test = [
            col for col in df_transformed.columns if col not in transformer.columns
        ]
        assert_frame_equal_dispatch(
            df_transformed[columns_to_test],
            expected[columns_to_test],
        )
        # Check outcomes for single rows
        df = nw.from_native(df)
        expected = nw.from_native(expected)
        for i in range(len(df)):
            df_transformed_row = transformer.transform(df[[i]].to_native())
            df_expected_row = expected[[i]].to_native()
            assert_frame_equal_dispatch(
                df_transformed_row[columns_to_test],
                df_expected_row[columns_to_test],
            )
```

you see above the test is converted with just 3 edits, which is nice!

Would you be able to go through and convert the old tests please, and then would be able to remove these new ones :)

tubular/capping.py
```diff
@@ -213,6 +214,7 @@ def check_capping_values_dict(
                 msg = f"{self.classname()}: both values are None for key {k}"
                 raise ValueError(msg)
 
+    @block_from_json
```
liamholmes31, line 225: Could  you also add this decorator to  prepare_quantiles, check_capping_values_dict, weighted_quantile please :)

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -853,3 +855,217 @@ def test_expected_output(
...
+        
+        # Create test data
+        df_dict = {
+            "a": [1, 3, 5, 7, 9, 11],
+            "b": [0, 2, 4, 6, 8, 10],
+        }
+        if weights_column:
+            df_dict["weight"] = [1, 1, 2, 2, 1, 1]
+        
+        if library == "pandas":
```
liamholmes31: this test can be dropped, but just fyi could have done this in one line with 'dataframe_init_dispatch' from tests.utils like:
`df=dataframe_init_dispatch(dataframe_dict=df_dict, library=library)`

liamholmes31 (commented):
Thanks for submitting your first PR!! :)

The main code looks good, the biggest comments I've left are on getting the testing into the right format - see what you think/let me know if you have any questions!

svojt:
Hi, I updated the code and tests to the correct format based on the comments.

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -443,6 +446,8 @@ def test_non_cap_column_left_untouched(
         if not transformer.polars_compatible and isinstance(df, pl.DataFrame):
             return
 
+        transformer = _handle_from_json(transformer, from_json)
```
liamholmes31: this line needs to occur after fit(+after any additional edits to the transformer), so that everything is captured in the to/from_json process :)

liamholmes31: think this same comment applies to a few tests here

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -15,8 +16,8 @@
     WeightColumnInitMixinTests,
 )
 from tests.utils import assert_frame_equal_dispatch, dataframe_init_dispatch
-from tubular.capping import BaseCappingTransformer
-
+from tubular.capping import BaseCappingTransformer, CappingTransformer, OutOfRangeNullTransformer
```
liamholmes31: Q: are these new imports used?

tubular/capping.py
```diff
@@ -146,6 +147,7 @@ def __init__(
         if capping_values:
             self._replacement_values = copy.deepcopy(self.capping_values)
 
+    @block_from_json
```
liamholmes31: oops, think this one may be causing test issues - can you remove actually? (sorry!)

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -14,7 +14,11 @@
     WeightColumnFitMixinTests,
     WeightColumnInitMixinTests,
 )
-from tests.utils import assert_frame_equal_dispatch, dataframe_init_dispatch
+from tests.utils import (
+    _handle_from_json,
+    assert_frame_equal_dispatch,
+    dataframe_init_dispatch,
```
liamholmes31, line 21: we are now very close, but the tests are failing when they are inherited by the child classes (e.g. CappingTransformer) as these aren't yet jsonable. 

Could you import `_check_if_skip_tests` here, and then in all of the edited tests will need to add the below, just after df and the transformer are defined:

```
if _check_if_skip_test(transformer, df, lazy=False, from_json=from_json):
            return
```

CHANGELOG.rst
```diff
@@ -27,6 +27,17 @@ We use the tags:
 
 Each individual change should have a link to the pull request after the description of the change.
 
+2.5.0 (unreleased)
+------------------
+
+Added
+^^^^^
+
+- feat: added `to_json` method to BaseCappingTransformer to enable dumping transformer state, including init attributes (`capping_values`, `quantiles`, `weights_column`) and fitted attributes (`quantile_capping_values`, `_replacement_values`) `#540 <https://github.com/azukds/tubular/pull/594>_`
```
liamholmes31, line 35: thanks, but this should go under the latest unreleased version (2.4.0) :)

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -474,15 +481,18 @@ def test_non_cap_column_left_untouched(
...
         self,
         fit_value,
         initialized_transformers,
         library,
+        from_json,
     ):
         """Test that the replacements from fit are not changed in transform."""
 
         transformer = initialized_transformers[self.transformer_name]
+        transformer = _handle_from_json(transformer, from_json)
```
liamholmes31: this needs to happen after the .fit call :)

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -712,6 +724,7 @@ def test_fixed_attributes_unchanged_from_transform(
         transformer.fit(df)
 
         transformer2 = uninitialized_transformers[self.transformer_name](**args)
+        transformer2 = _handle_from_json(transformer2, from_json)
```
liamholmes31: this needs to be moved below the fit call, so that the json process will capture everything

tests/capping/test_BaseCappingTransformer.py
liamholmes31: resolving as covered in new comments

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -735,11 +748,13 @@ def expected_df_1(self, library="pandas"):
         return dataframe_init_dispatch(dataframe_dict=df_dict, library=library)
```
liamholmes31, line 787: a few other tests need to have the `from_json` stuff added here:

- test_replacement_values_dict_empty_error
- test_replacement_values_not_fit_error
-  test_capping_values_empty_error
-  test_quantile_capping_values_empty_error
- test_quantile_capping_values_not_fit_error
- test_non_numeric_column_error

(essentially we need to convert all of the tests in this class)

liamholmes31 (commented):
We are getting close to done now, just dealing with some remaining CI errors and I did also spot some unconverted tests which I must have missed before!

Thanks for your work on this :)

CHANGELOG.rst
```diff
@@ -47,6 +47,12 @@ Changed
 - feat: added `to_json` method for `BetweenDatesTransformer` `#554 <https://github.com/azukds/tubular/issues/554>`_
 - feat: added a new transformer called DatetimeComponentExtractor
 - feat: add a decorator to register transformer classes by name `#588 <https://github.com/azukds/tubular/issues/588>_`
+
+- feat: added `to_json` method to BaseCappingTransformer to enable dumping transformer state, including init attributes (`capping_values`, `quantiles`, `weights_column`) and fitted attributes (`quantile_capping_values`, `_replacement_values`) `#540 <https://github.com/azukds/tubular/pull/594>_`
```
liamholmes31: apologies but we released a new version yesterday afternoon, so this now goes under 2.5.0 after all!

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -551,11 +573,13 @@ def test_non_numeric_column_error(
             transformer.transform(df)
 
     @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize("from_json", [True, False])
```
liamholmes31: ah my bad, you can actually remove the json stuff from this one.

The json stuff depends on fit having been called, and this test purposefully does not call fit, so it doesn't belong here

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -634,11 +675,13 @@ def test_capping_values_empty_error(
...
```
liamholmes31: the same applies here, can remove :)

tubular/capping.py
```diff
@@ -595,6 +599,36 @@ def transform(
...
+
+        Returns
+        -------
+         dict
+        Dictionary containing all necessary attributes to recreate the transformer with
+        `from_json`. Keys include 'init' (initialization parameters) and 'fit' (fitted values).
+
+        """
+        data = super().to_json()
+
```
liamholmes31, line 613: we would normally add a `check_is_fitted` check here, but this transformer is messy and only fits for quantiles (I think we'll split it into ValueCappingTransformer and QuantileCappingTransformer in the future - so can leave out from this one :)

svojt:
Ok, i deleted json from_from that 2 tests and updated changelog.

PR #599, opened 2025-11-12
Feature/optimise base capping transformer fit

kajol-patel:
Issue https://github.com/azukds/tubular/issues/484
- added _weighted_quantile_expr to compute cum_sum(weights) / sum(weights) as an expression
- updated BaseCappingTransformer.weighted_quantile to use the new helper
- cleaned up filtering expressions (not_null_expr, nonzero_weight_expr).

PR #604, opened 2025-11-13
made optimisation change to median imputer, and fixed type hints in s…

liamholmes31:
…tats file

spotted a couple of things that could be improved/fixed here while reviewing https://github.com/azukds/tubular/pull/599

### DIFF
--- a/tubular/_stats.py
+++ b/tubular/_stats.py
@@ -77,7 +77,7 @@ def _get_mean_calculation_expressions(
     weights_column: str
         name of weights column
 
-    initial_columns_exprs: dict[str, nw.Expr]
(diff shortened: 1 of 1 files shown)


tubular/_stats.py
```diff
@@ -67,7 +53,7 @@ def _get_median_calculation_expression(
 def _get_mean_calculation_expressions(
```
Boluwatife28, line 53: N: should we update the docstring to mark as Optional where necessary to match the parameter definition?

liamholmes31: done :) https://github.com/azukds/tubular/pull/604/commits/8569f296cbc2be532a0774562f01fe0bf1e06ba2

PR #605, opened 2025-11-13
Add beartype decorator to NullIndicator and SetValueTransformer.

shameem2k4:


tests/imputers/test_NullIndicator.py
```diff
@@ -23,6 +24,13 @@ class TestInit(ColumnStrListInitTests):
     def setup_class(cls):
         cls.transformer_name = "NullIndicator"
 
+    @pytest.mark.parametrize("columns", [[], None])
```
liamholmes31: I think this one isn't actually necessary, as it should be inherited from ColumnStrListInitTests :)

tubular/misc.py
```diff
@@ -69,7 +77,7 @@ def __init__(
         value : various
             Value to set.
 
-        **kwargs: dict[str, Any]
+        **kwargs: dict[str, bool]
```
liamholmes31: because of the way beartype evaluates **kwargs, I think the type hint here should just be `bool`

tests/misc/test_SetValueTransformer.py
```diff
@@ -44,6 +45,13 @@ class TestInit(ColumnStrListInitTests):
     def setup_class(cls):
         cls.transformer_name = "SetValueTransformer"
 
+    @pytest.mark.parametrize("columns", [[], None])
```
liamholmes31: same comment on this one being not needed - can you change the test to be for the 'value' arg instead?

tests/imputers/test_NullIndicator.py
shameem2k4: Removed.

tubular/misc.py
shameem2k4: Changed the type hint for kwargs in doc to bool.

tests/misc/test_SetValueTransformer.py
shameem2k4: Changed the test to validate 'value' arg instead of column.

liamholmes31: thanks! could you rename the test to e.g. `test_value_arg_type_error`, and can delete the parametrise that works with 'columns' on line 49

liamholmes31 (approved):

PR #607, opened 2025-11-13
fixed bug with null handling on GroupRareLevelsTransformer.fit

liamholmes31:
transformer was hitting unhandled error in .fit from trying to call `sorted` on set containing nulls

### DIFF
--- a/utils/feature_table.py
+++ b/utils/feature_table.py
@@ -2,8 +2,11 @@
 
 import pathlib
 import re
+import sys
 
 import pandas as pd
+
+sys.path.append(str(pathlib.Path(__file__).resolve().parents[1]))
 
 from tubular.base import FEATURE_REGISTRY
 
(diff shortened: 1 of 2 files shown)


tubular/nominal.py
```diff
@@ -530,12 +530,13 @@ def fit(
 
         self._check_str_like_columns(schema)
 
-        present_levels = {
-            c: sorted(set(X.get_column(c).unique())) for c in self.columns
```
SadieRobertson, line 534: Q: was the null error just coming from trying to sort?

SadieRobertson: yes! Happy to approve :)

SadieRobertson (approved):

utils/feature_table.py
```diff
@@ -2,9 +2,12 @@
 
 import pathlib
 import re
+import sys
 
 import pandas as pd
```
liamholmes31, line 8: this import has randomly broken since we merged the feature table one in, but this fixes

PR #603, opened 2025-11-13
Datetimecomponentextractor transformer

hariinduu:
New transformer to extract date time information

@488

Issue #559, opened 2025-10-16
look at removing new_column_name argument from transformers?


liamholmes31:
What?
consider removing this argument from transformers, think it may make more sense to just return fixed names and leave it to users to rename on the other side if they wish

Why?
simplify repo, cut down on number of arguments going into transformers

How?
could remove and allow this to be absorbed in **kwargs, and raise a deprecated warning when it's provided

SadieRobertson:
In [this PR comment](https://github.com/azukds/tubular/pull/545#discussion_r2433201184) it was suggested that we could remove duplicated overwrites of the 'get_feature_names_out' method where new_column_name arg is used. However, these updates to the method would not be needed once this issue is resolved.

Issue #557, opened 2025-10-16
Remove DropOriginalMixin


liamholmes31:
What?
This mixin class doesn't add useful functionality and just clutters the repo, a user can easily filter the dataframe output on their end if they choose to. I think we should just remove.

Why?
tidy repo

How?
delete DropOriginalMixin and delete usages around the repo. To avoid an API breaking change, suggest we just leave the argument init methods and add a deprecated warning.

PR #612, opened 2025-11-14
#556 - added 'to_json' functionality for DatetimeSinusoidCalculator

KatheeO:
When modifying the signature of the method, I followed advice from #570 comments, using **kwargs instead of explicitly defining 'copy' and 'return_native'.

Parameter new_column_name is needed by the parent classes, so I found it necessary to also include it in method signature, while setting the default to the original "dummy" string.

Also updated the docstring with the description of this parameter and added entries for the other explicitly defined parameters - verbose and drop_original - too, as they seem to be included in other classes that define them, but were missing here.

### DIFF
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -1469,10 +1469,6 @@ class DatetimeSinusoidCalculator(BaseDatetimeTransformer):
     drop_original: bool, default = False
         Indicates whether to drop original columns.
 
-    new_column_name : str, default = "dummy"
-         This transformer generates its own column names and does not use this parameter.
-         A placeholder is therefore set for API compatibility and consistency with other datetime classes.
-
     **kwargs
          Arbitrary keyword arguments passed onto BaseTransformer.init method.
          Restricted to only boolean values, e.g. `copy`, `return_native`.
@@ -1539,16 +1535,20 @@ class DatetimeSinusoidCalculator(BaseDatetimeTransformer):
             dict[str, DatetimeSinusoidUnitsOptionStr],
         ],
         period: Union[NumberNotBool, dict[str, NumberNotBool]] = 2 * np.pi,
-        verbose: bool = False,
         drop_original: bool = False,
-        new_column_name: str = "dummy",
-        **kwargs: bool,
+        **kwargs: Union[bool, str],
     ) -> None:
+        if "new_column_name" in kwargs:
+            warnings.warn(
+                f"{self.classname()}: new_column_name arg is unused by this transformer",
+                stacklevel=2,
+            )
+            kwargs.pop("new_column_name", None)
+
         super().__init__(
             columns=columns,
             drop_original=drop_original,
-            new_column_name=new_column_name,
-            verbose=verbose,
+            new_column_name="dummy",
             **kwargs,
         )
 
@@ -1616,9 +1616,13 @@ class DatetimeSinusoidCalculator(BaseDatetimeTransformer):
         """
         json_dict = super().to_json()
 
-        json_dict["init"]["method"] = self.method
-        json_dict["init"]["units"] = self.units
-        json_dict["init"]["period"] = self.period
+        json_dict["init"].update(
+            {
+                "method": self.method,
+                "units": self.units,
+                "period": self.period,
+            }
+        )
 
         return json_dict
 


tubular/dates.py
```diff
@@ -1526,12 +1541,15 @@ def __init__(
...
         verbose: bool = False,
         drop_original: bool = False,
+        new_column_name: str = "dummy",
+        **kwargs: bool,
     ) -> None:
         super().__init__(
             columns=columns,
             drop_original=drop_original,
-            new_column_name="dummy",
+            new_column_name=new_column_name,
```
liamholmes31: so this is a bit of an odd case, where the base class is set up to expect new_column_name but the transformer itself doesn't actually use it. So it can be handled in two ways:

1. Don't have new_column_name as an argument for this  class, and just pass 'dummy' to the base class (this is the existing setup)
2. have new_column_name as an argument that doesn't do anything, which is what you've added here.

In the long term, I think we're going to drop the [new_column_name functionality](https://github.com/azukds/tubular/issues/559), but this will be a breaking change so we're holding off on it for now - but given that it's likely to be removed in the future, I would stick with approach 1 and revert this part, so that we don't create more API changes down the line

tubular/dates.py
```diff
@@ -1526,12 +1541,15 @@ def __init__(
         period: Union[NumberNotBool, dict[str, NumberNotBool]] = 2 * np.pi,
         verbose: bool = False,
```
liamholmes31: verbose is one that can be absorbed in kwargs, so could delete this argument

tubular/dates.py
```diff
@@ -1575,6 +1593,35 @@ def get_feature_names_out(self) -> list[str]:
...
+        ...     method='sin',
+        ...     units='month',
+        ... )
+        >>> transformer.to_json()
+        {'tubular_version': ..., 'classname': 'DatetimeSinusoidCalculator', 'init': {'columns': ['a'], 'copy': False, 'verbose': False, 'return_native': True, 'new_column_name': 'dummy', 'drop_original': False, 'method': ['sin'], 'units': 'month', 'period': 6.283185307179586}, 'fit': {}}
+
+        """
+        json_dict = super().to_json()
+
+        json_dict["init"]["method"] = self.method
```
liamholmes31: N:  think it's neater to use json_dict["init"].update({...})

tubular/dates.py
liamholmes31: unless I'm missing the reason why it had to be included as an argument here? :)

tubular/dates.py
liamholmes31: we often use the following annotations on PR comments btw:
P: praise
S: suggestion
N: nitpick
B: bug/blocker

tubular/dates.py
KatheeO: Ah, yes, there certainly is a reason for its inclusion here, I'll try to explain step by step.

When DatetimeSinusoidCalculator's (DSC) _to_json_ method calls _super().to_json()_, it inherits the base json keys from its parent, in this case it's BaseGenericDateTransformer (BGDT). BGDT adds the key "_new_column_name_" (and others) to its json, and then DSC adds its own keys on top of that. To my knowledge, a child cannot delete keys/parameters that are added by the parent, it can only override the values of the existing keys.
This becomes a problem when function _from_json_ is called, which tries to reconstruct the object by unpacking the keys as keyword arguments (_cls(**json["init"])_). One of those keys is gonna be "_new_column_name_", but DSC's init does not have that parameter explicitly defined, and brings about an error when _from_json_ is called.

Skipping forward, there are three keys that BGDT adds to its json that DSC's signature doesn't explicitly define, them being "_new_column_name_", "_copy_", and "_return_native_". We need to somehow catch and handle these keys, otherwise we will be getting this error:
> "E   TypeError: DatetimeSinusoidCalculator.__init__() got an unexpected keyword argument 'copy'". (same for return_native and new_column_name)

We have a couple options then:

- Modifying the parent class's init signature. :) Precisely what you said, a breaking change.
- Explicitly defining all three keys. As noted by you in #570, having arguments of little importance/rarely overriden, such as copy, explicitly defined clutters the args list a bit.
- So we can bundle them into **kwargs. If the "_new_column_name_" is only passed in the _super().__init___ call (the way it was before my edit), **kwargs will catch this keyword argument and raise an error if _**kwargs: bool_. The error raised:
    > E   beartype.roar.BeartypeCallHintParamViolation: Method tubular.dates.DatetimeSinusoidCalculator.__init__() parameter kwargs='new_column_name' violates type hint <class 'bool'>, as str 'new_column_name' not instance of bool.

    Bundling booleans into **kwargs is okay, but we still need to handle the "_new_column_name_", which is a string. Setting the type of kwargs from "**kwargs: bool" to "**kwargs: Any" may result in unexpected behaviour in some cases, and is generally not advised when strict type checking is used (hello beartype :)). 

- In my opinion, the best option is catching the cluttery booleans into **kwargs, and explicitly define "_new_column_name_" in the __init__signature of DSC. This will make the to_json/from_json work properly and minimize unexpected behaviour.



You probably didn't need such a long and detailed reasoning, but I felt the need to write it as such for my own understanding of the chain :)

Also, to address your note about wanting to hear about my usage of tubular, is there a place I can contact you, so we don't clutter the comments here?

tubular/dates.py
KatheeO: Will do! What about _drop_original_? That one can, too, be absorbed in **kwargs, do we bundle that too, or leave it explicitly defined? I see it is left explicitly defined in all the other transformers, so I assume we leave it be?

tubular/dates.py
liamholmes31: ah OK i'm with you thanks :) I think my suggestion would be to bundle into kwargs by changing the type hint to `Union[bool, str]` - this way we don't confuse users by exposing an argument which isn't actually used.

As mentioned, `new_column_name` is a bit messy generally and we will probably remove it in the near future!

liamholmes31: re the other point, feel free to drop me an email on liam.holmes@allianz.co.uk and we can discuss/arrange a chat!

tubular/dates.py
liamholmes31: ah yes you're right this one could also be absorbed! I would leave it out for now, as it does affect the output of the transformer it makes sense to expose it to users to me. 

(it is another one which we plan to remove https://github.com/azukds/tubular/issues/557)

tubular/dates.py
liamholmes31: this [pr](https://github.com/azukds/tubular/pull/603/files#diff-8fa6d11bc8bef9164e04fd99506197093b739e06a643eba7ae9687b144dcd3b2) has been battling the same issues, I think you'll also need to copy from here and add a section like below to init:

```
if "new_column_name" in kwargs:
            warnings.warn(
                f"{self.classname()}: new_column_name arg is unused by this transformer",
                stacklevel=2,
            )
            kwargs.pop("new_column_name", None)
```

otherwise new_column_name ends up being passed twice in the to/from_json process

tubular/dates.py
```diff
@@ -1539,16 +1535,20 @@ def __init__(
...
-        **kwargs: bool,
+        **kwargs: Union[bool, str],
     ) -> None:
+        if "new_column_name" in kwargs:
+            warnings.warn(
+                f"{self.classname()}: new_column_name arg is unused by this transformer",
+                stacklevel=2,
+            )
+            kwargs.pop("new_column_name", None)
+
```
KatheeO, line 1547: Just as a note here, we could also do just kwargs.pop("new_column_name", "dummy") instead of the whole block, but informing users that they input a value that is not used by the transformer is indeed better for now.

liamholmes31 (approved):

liamholmes31:
another one down, thanks @KatheeO !!

PR #614, opened 2025-11-17
Add @register decorator that adds class object to the dictionary.

shameem2k4:


tubular/_registry.py
```diff
@@ -0,0 +1,13 @@
+CLASS_REGISTRY = {}
+
+
+def register(cls: type) -> type:
```
liamholmes31: S: would typehint `BaseTransformer` here

tubular/_registry.py
```diff
@@ -0,0 +1,13 @@
...
+    """Add class object to registry(dict).
```
liamholmes31: maybe better to word as 'Add transformer to registry dict.'

tubular/_registry.py
```diff
@@ -0,0 +1,13 @@
...
+
+    Returns
+    -------
+    cls - class object.
```
liamholmes31: would put 'cls - transformer'

tubular/aggregations.py
```diff
@@ -219,6 +220,7 @@ def transform(
         return _return_narwhals_or_native_dataframe(X, return_native=return_native)
```
liamholmes31, line 222: sorry realise I said the opposite on friday - but on second thoughts could you include the base classes please? realised this will allow us to remove [`get_all_classes`](https://github.com/azukds/tubular/blob/main/tests/conftest.py) in a future issue/PR :)

Issue #595, opened 2025-11-11
add a table to the docs that lists per transformer functionality


liamholmes31:
What?
add a table to the docs that lists per transformer functionality

Why?
make it easier for users to identify if a given transformer is polars/lazyframe/json compatible

How?
we could do something like the below in base transformer:
```
class_registry={}

class BaseTransformer:

    def __init_subclass__(cls):
         class_registry[cls.__name__]={'polars_compatible': cls.polars_compatible, ...}
                           
```

we could then add a script that would convert this info into a markdown table and set this up to be run as part of docs build/readme

```
pkg=importlib.import_module('tubular')
registry=getattr(pkg, 'class_registry')
# sort rows/columns
df=pd.DataFrame.from_dict(registry, orient='index').sort_index().sort_index(axis=1)
# replace bools with unicode tick/cross
df=df.replace({True:  "\u2705", False: "\u274C"})
md_table=df.to_markdown(tablefmt='github')
```

PR #615, opened 2025-11-17
Feature/feature table

liamholmes31:
- Added FEATURE_REGISTRY dict maintained by BaseTransformer to collate info on per-transformer functionality
- Added script that will process this info into table and embed in README
- Added pre-commit/CI steps which will automate this

https://github.com/azukds/tubular/issues/595

### DIFF
--- a/tubular/mapping.py
+++ b/tubular/mapping.py
@@ -340,7 +340,7 @@ class BaseMappingTransformMixin(BaseTransformer):
         schema = X.schema
         mapping_exprs = {
             col: nw.col(col).cast(nw.String)
-            if schema[col] in [nw.Categorical, nw.Enum]
+            if schema[col] in {nw.Categorical, nw.Enum}
             else nw.col(col)
             for col in self.mappings
         }
@@ -858,7 +858,7 @@ class BaseCrossColumnNumericTransformer(BaseCrossColumnMappingTransformer):
 
         for j in mappings.values():
             for k in j.values():
-                if type(k) not in [int, float]:
+                if type(k) not in {int, float}:
                     msg = f"{self.classname()}: mapping values must be numeric"
                     raise TypeError(msg)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,13 +33,14 @@ Each individual change should have a link to the pull request after the descript
 Changed
 ^^^^^^^
 - turned on beartype for WeightsColumnMixin and associated transformers `#561 <https://github.com/azukds/tubular/issues/561>`_
-- feat: added auto generated feature table to readme, to track transform support for e.g. polars/lazyframes/json `#595 <https://github.com/azukds/tubular/issues/595>`_
-- feat: added FEATURE_REGISTRY dict maintained by BaseTransformer, which records per-transformer functionality info
-- feat: added `deprecated` class attr to deprecated classes, useful to filter them out of e.g. feature table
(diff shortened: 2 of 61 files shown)


README.md
```diff
@@ -55,6 +55,38 @@ transformer.transform(test_df)
 # └─────┴─────┴─────┘
 ```
 
+We are currently in the process of rolling out support for some new features:
+- to/from json methods for our transformers to allow json storage
+- polars lazyframe support
+
+You can track our progress here:
+<!-- AUTO-GENERATED feature table -->
```
liamholmes31, line 63: this looks odd here, but if you go to the branch itself and look at the README you can see it will be formatted correctly!

tubular/mapping.py
```diff
@@ -1113,6 +1136,8 @@ class attribute, indicates whether transformer works with lazyframes
 
     jsonable = False
 
+    deprecated = True
```
SadieRobertson, line 1044: S: This seems to have helped us to identify lots of the depreciated transformers which didn't have the appropriate label! Do you think it could also be useful to have a list of the depreciated transformers generated in the readme?

utils/feature_table.py
```diff
@@ -0,0 +1,54 @@
...
+
+    """
+    df = pd.DataFrame.from_dict(FEATURE_REGISTRY, orient="index").sort_index()
+    # replace bools with unicode tick/cross
+    df = df.replace({True: ":heavy_check_mark:", False: ":x:"})
+
+    return df.to_markdown(tablefmt="github").strip()
+
+
+def insert_table_to_readme(table: str, readme_text: str) -> None:
```
SadieRobertson, line 46: S: Could add a test for these functions?

utils/feature_table.py
```diff
@@ -0,0 +1,54 @@
+"""Script to run in pre-commit/CI in order to keep the README feature table up to date."""
```
SadieRobertson: S: Could be more appropriate to more this into a utils function or docs folder or something similar?

SadieRobertson (changes requested):

tubular/mapping.py
liamholmes31: this is actually a new label I added in order to help filter them out from the table :) 

I did consider including it as a column, but decided maybe deprecated transformers are an odd thing to advertise to new users in the README? there may be somewhere more suitable in the docs for this info though, could be worth spinning into another issue?

SadieRobertson: I like the new label then! Yeah I agree, not in the table but as a list somewhere else in the docs. I separate issue would be good, also a good opportunity for someone to give updating the CI a go.

liamholmes31: have written that up here https://github.com/azukds/tubular/issues/625

utils/feature_table.py
liamholmes31: added some doctests here, feels like these might be enough but open to feedback on that!

https://github.com/azukds/tubular/pull/615/commits/46723c83fc906864db5bdc3caadee0aef965bf84

liamholmes31: oops, one was failing https://github.com/azukds/tubular/pull/615/commits/062c960223e1d3f2729fa1f64265f731adf3bfe0

utils/feature_table.py
liamholmes31: moved :) https://github.com/azukds/tubular/pull/615/commits/1ff680dfccdc89585dc0c32382cb261b1d71085e#diff-ee49282f461b4c8ad179f79dd5bcdf93124561074c64a771366caf93e99b9320

PR #621, opened 2025-11-19
Feature/doc lints

liamholmes31:
turned on doc lints for remaining files (except nominal which has large open PRs/would cause conflicts)

### DIFF
--- a/tubular/dates.py
+++ b/tubular/dates.py
@@ -94,7 +94,7 @@ class BaseGenericDateTransformer(
 
         Parameters
         ----------
-        columns : List[str]
+        columns : Union[list[str], str]
             List of 2 columns. First column will be subtracted from second.
 
         new_column_name : str
@@ -391,7 +391,7 @@ class BaseDatetimeTransformer(BaseGenericDateTransformer):
 
         Parameters
         ----------
-        columns : List[str]
+        columns : Union[list[str], str]
             List of 2 columns. First column will be subtracted from second.
 
         new_column_name : str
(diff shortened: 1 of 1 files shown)


tubular/dates.py
```diff
@@ -108,14 +90,34 @@ def __init__(
         drop_original: bool = False,
         **kwargs: Optional[bool],
     ) -> None:
+        """Initialise class instance.
+
+        Parameters
+        ----------
+        columns : List[str]
```
Boluwatife28: Q: Just to verify, should the columns parameter also accept a string, or is it strictly a list of strings?

tubular/dates.py
```diff
@@ -403,6 +387,23 @@ def __init__(
...
```
Boluwatife28: same as the previous comment

tubular/dates.py
```diff
@@ -473,6 +473,8 @@ def transform(
 
 
 class DateDifferenceUnitsOptions(str, Enum):
+    """options for return units in DateDifferenceTransformer."""
```
Boluwatife28: N: We should start with a capital letter to be consistent

tubular/dates.py
liamholmes31: good catch! https://github.com/azukds/tubular/pull/621/commits/792d77b27125bf943191fce83d3ea666215a5f7f

tubular/dates.py
liamholmes31: good catch! https://github.com/azukds/tubular/pull/621/commits/792d77b27125bf943191fce83d3ea666215a5f7f

tubular/dates.py
liamholmes31: good point :) https://github.com/azukds/tubular/pull/621/commits/792d77b27125bf943191fce83d3ea666215a5f7f

Issue #509, opened 2025-09-25
turn on PLR ruff ruleset


liamholmes31:
What?
turn on PLR ruleset, contains some nice refactor suggestions https://docs.astral.sh/ruff/rules/#refactor-plr

Why?
ramp up CI standards

How?
add "PLR" to list of ruff rules in [tool.ruff.lint] section of pyproject.toml, then run ruff format . and ruff check --fix . fix any remaining errors

PR #622, opened 2025-11-19
Feature/ruff plr rules

liamholmes31:
added additional ruff ruleset

https://github.com/azukds/tubular/issues/509

### DIFF
--- a/tubular/numeric.py
+++ b/tubular/numeric.py
@@ -1619,7 +1619,7 @@ class PCATransformer(BaseNumericTransformer):
                 run randomized SVD by the method of Halko et al.
             .. sklearn versionadded:: 0.18.0
 
-        random_state : int, RandomState instance or None, default=None
+        random_state : int or None, default=None
             Used when the 'arpack' or 'randomized' solvers are used. Pass an int
             for reproducible results across multiple function calls.
             .. sklearn versionadded:: 0.18.0


pyproject.toml
```diff
@@ -81,19 +83,26 @@ select = [
     "DTZ",
     "E", # pycodestyle
     "EM",
+    "EXE",
```
liamholmes31, line 94: I added a few extra rules here, because they had no effect I thought we may as well turn on!

tests/capping/test_BaseCappingTransformer.py
```diff
@@ -431,10 +432,6 @@ def test_non_cap_column_left_untouched(
         if _check_if_skip_test(transformer, df, lazy=False, from_json=from_json):
             return
 
-        # if transformer is not polars compatible, skip polars test
```
Ritunjai-Sharma, line 434: Assuming this has been removed since BaseCappingTransformer is polars compatible now?

tests/dates/test_BetweenDatesTransformer.py
```diff
@@ -412,7 +421,6 @@ def test_mismatched_datetypes_error(
         self,
         columns,
         datetime_col,
-        date_col,
```
Ritunjai-Sharma, line 415: why has this been removed?

tests/capping/test_BaseCappingTransformer.py
liamholmes31: this is now covered by `_check_if_skip_test` above

didn't strictly need to be removed here, but came across it during some debugging so took out

tests/dates/test_BetweenDatesTransformer.py
liamholmes31: was hitting an unused arg error of some kind I believe, cus it doesn't actually do anything in the test itself

tubular/numeric.py
```diff
@@ -1570,14 +1563,19 @@ class attribute, indicates whether transformer works with lazyframes
...
-        svd_solver: str = "auto",
-        random_state: int | np.random.RandomState = None,
+        columns: Optional[Union[str, list[str]]],
+        n_components: Union[
+            StrictlyPositiveInt,
+            FloatBetweenZeroOne,
+            Literal["mle"],
+        ] = 2,
+        svd_solver: Literal["auto", "full", "arpack", "randomized"] = "auto",
+        random_state: Optional[int] = None,
```
Ritunjai-Sharma, line 1576: Looks like random_state will no longer accept np.random.RandomStateRandomState? If so, worth updating this in the random_state part of docstring below

liamholmes31: ah yeah, I did this unintentionally but i'm happy with it as a simplification - also this one is currently deprecated anyway so no biggie imo, updated here https://github.com/azukds/tubular/pull/622/commits/25dcbb9d2d8458e18364d8f1409084707af010a9

PR #627, opened 2025-11-25
#554 - added to_json for BetweenDatesTransformer

KatheeO:
#554

### DIFF
--- a/tests/dates/test_BetweenDatesTransformer.py
+++ b/tests/dates/test_BetweenDatesTransformer.py
@@ -319,7 +319,8 @@ class TestTransform(
 
         assert_frame_equal_dispatch(expected, df_transformed)
 
-    def test_warning_message(self):
+    @pytest.mark.parametrize("from_json", [True, False])
+    def test_warning_message(self, from_json):
         """Test a warning is generated if not all the values in column_upper are greater than or equal to column_lower."""
         x = BetweenDatesTransformer(
             columns=["a", "b", "c"],
@@ -327,6 +328,8 @@ class TestTransform(
             lower_inclusive=True,
             upper_inclusive=True,
         )
+
+        x = _handle_from_json(x, from_json)
 
         df = d.create_is_between_dates_df_2()
         df = nw.from_native(df)
@@ -404,6 +407,7 @@ class TestTransform(
         ("library"),
         ["pandas", "polars"],
     )
+    @pytest.mark.parametrize("from_json", [True, False])
     def test_mismatched_datetypes_error(
         self,
         columns,
@@ -411,6 +415,7 @@ class TestTransform(
         date_col,
         uninitialized_transformers,
         library,
+        from_json,
     ):
         "Test that transform raises an error if one column is a date and one is datetime"
 
@@ -418,6 +423,8 @@ class TestTransform(
             columns=columns,
             new_column_name="c",
         )
+
+        transformer = _handle_from_json(transformer, from_json)
 
         df = create_date_diff_different_dtypes(library=library)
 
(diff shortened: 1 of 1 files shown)


tests/dates/test_BetweenDatesTransformer.py
```diff
@@ -475,14 +497,13 @@ def test_bad_timezones_error(
...
-        msg = "a type should be in ['Datetime', 'Date'] but got Unknown. Note, Datetime columns should have time_unit in ['us', 'ns', 'ms'] and time_zones from zoneinfo.available_timezones()"
-
         with pytest.raises(
             TypeError,
         ) as exc_info:
             transformer.transform(df)
 
-        assert msg in str(exc_info.value)
+        assert "type should be in ['Datetime', 'Date']" in str(exc_info.value)
+        assert "time_zones from zoneinfo.available_timezones()" in str(exc_info.value)
```
KatheeO: This specific test (test_bad_timezones_error) was failing, it was putting column 'b' in msg, even though 'b' is correct. Here is an excerpt with the error message:

```
tests/dates/test_BetweenDatesTransformer.py:444 (TestTransform.test_bad_timezones_error[localtime-pandas])
("a type should be in ['Datetime', 'Date'] but got Unknown. Note, Datetime "
 "columns should have time_unit in ['us', 'ns', 'ms'] and time_zones from "
 'zoneinfo.available_timezones()') != ("BetweenDatesTransformer: b type should be in ['Datetime', 'Date'] but got "
 "Unknown. Note, Datetime columns should have time_unit in ['us', 'ns', 'ms'] "
 'and time_zones from zoneinfo.available_timezones()')
```

So, I modified it to check for the key parts of the error message rather than a hardcoded column name and all tests are passing.

KatheeO: But of course it is registering b as Unknown either way, I don't know why. When I print out exec_info, the first test for value "Factory" correctly registers 'a' as the incorrect value. However, the second test - for parameter value "localtime", registers b as the incorrect value.

liamholmes31: I have actually seen similar errors when working on my mac, so wonder if this is down to some weird OS difference. I just opened this PR in a codespace and reverted the change, and the test does pass there - this could be one to spin out into a separate issue to investigate, if the reverted change passes CI I would be happy to approve for the time being anyway!

tests/dates/test_BetweenDatesTransformer.py
```diff
@@ -475,14 +497,13 @@ def test_bad_timezones_error(
...
 
     @pytest.mark.parametrize("library", ["pandas", "polars"])
```
liamholmes31, line 520: looking good :) but could you add the json handling to all of the tests in this class please

liamholmes31 (commented):
thanks KatheeO!!

just left a couple of comments around the error you're seeing and some remaining transform tests :)

tests/dates/test_BetweenDatesTransformer.py
liamholmes31: let me know if you're working on mac to confirm my OS theory, and i'll get the bug issue written up!

KatheeO: Actually I'm on Arch Linux. Might be something with the OS as you mentioned. Should I revert the change back to the original for now?

tests/dates/test_BetweenDatesTransformer.py
KatheeO: Of course, will do :)

tests/dates/test_BetweenDatesTransformer.py
liamholmes31: thanks, will capture that in the issue :) yes please! I think it will pass in CI once reverted (hopefully)

liamholmes31: new issue is here, i'll look into! https://github.com/azukds/tubular/issues/628

PR #629, opened 2025-11-28
bugfix: edits to MappingTransformer to avoid recursion error for many mappings

liamholmes31:
when/then/otherwise approach was failing for many mappings. This approach was also going to hit issues in the move to lazyframes, so this change was due anyway.. have changed the logic to mainly use replace_strict, and just use when/then/otherwise one time (instead of once per mapping before) to separate the rows which can and cannot be mapped.

did some quick profiling, and looks like an improvement:

Old Mapping

	several rows
	pd 0.007901873588562012
	pl 0.0018085432052612305

	single row
	pd 0.004298458099365235
	pl 0.0013586020469665528

new mapping

	several rows
	pd 0.003467137813568115
	pl 0.001332833766937256

	single row
	pd 0.0036037898063659667
	pl 0.0013688111305236816

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,7 +27,7 @@ We use the tags:
 
 Each individual change should have a link to the pull request after the description of the change.
 
-2.5.0 (01/12/2025)
+2.5.0 (unreleased)
 ------------------
 
 Changed


Ritunjai-Sharma (commented):
Looks like we are now casting categorical/enum columns to String before replacement, but using the original mapping dict keys? If these keys are bool, int etc., I think replace_strict will compare strings in the column to non-string keys and won’t match, so it will produce None for rows that should be mapped? Also, later on, looks like fill_null only fills nulls for null-key mappings, not for these cases.

Are we not expecting keys to be anything other than string? If they will only be string, the current setup should be fine as is.

tubular/mapping.py
```diff
@@ -445,60 +329,52 @@ def transform(
...
         X = super().transform(X, return_native_override=False)
 
+        mappable_conditions = {
+            col: nw.col(col).is_in(self.mappings[col]) for col in self.mappings
+        }
+
         # if the column is categorical, narwhals struggles to infer a type
         # during the when/then logic, so we need to record this so that we can
         # tell polars to use string as a common type
         # types are then corrected before returning at the end
```
Ritunjai-Sharma, line 339: I think this comment talks about “recording” and casting outcomes. I think this no longer matches the new implementation, i.e., casting the entire column to String before replacement and then casting back to the requested dtype. Worth updating this comment to reflect that?

liamholmes31: thanks! updated  https://github.com/azukds/tubular/pull/629/commits/effffed8defbc60fc60918a94d8585c278ee1ff7

PR #641, opened 2025-12-08
Feature/add to json support one dk means transformer

kajol-patel:
Issue: [#551](https://github.com/azukds/tubular/issues/551)
- added to_json method for OneDKmeansTransformer
- set the jsonable class attr to True
- added the block_from_json decorator
- setup transform tests to do an addition run with json dumped/loaded versions of the transformer

### DIFF
--- a/tubular/numeric.py
+++ b/tubular/numeric.py
@@ -278,7 +278,7 @@ class OneDKmeansTransformer(BaseNumericTransformer, DropOriginalMixin):
         OneDKmeansTransformer(columns=['a'], kmeans_kwargs={'random_state': 42},
                               n_clusters=2, new_column_name='new')
         >>> x.to_json()
-        {'tubular_version': ..., 'classname': 'OneDKmeansTransformer', 'init': {'columns': ['a'], 'copy': False, 'verbose': False, 'return_native': True, 'new_column_name': 'new', 'n_init': 'auto', 'n_clusters': 2, 'drop_original': False, 'kmeans_kwargs': {'random_state': 42}}, 'fit': {'bins': array([3, 4])}}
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -37,6 +37,7 @@ Changed
 - feat: added lazyframe testing for BaseTransfomer
 - feat: introduced `lazyframe_compatible` class attr to all transformers
 - feat: as part of lazyframe work, transformers no longer error for emptyframes (they just return emptyframes)
+- feat: added 'to_json' method for OneDKmeansTransformer `#551 <https://github.com/azukds/tubular/issues/551>_`
```
liamholmes31: think you might need to pull in main, we're on 2.5.0 now :)

tubular/numeric.py
```diff
@@ -252,10 +252,53 @@ class attribute, indicates whether transformer works with lazyframes
...
+        json_dict["init"].update(
+            {
+                "new_column_name": self.new_column_name,
+                "n_init": self.n_init,
+                "n_clusters": self.n_clusters,
+                "drop_original": self.drop_original,
+                "kmeans_kwargs": self.kmeans_kwargs,
+            },
+        )
+        json_dict["fit"]["bins"] = self.bins
```
liamholmes31, line 296: S: in the fit logic, could we change bins to be stored as a list? either by changing to_numpy to to_list or just casting to a list at the end if that causes issues?

storing simpler datatypes in json is better if we can help it! and I think as this would just change how an attr is stored it shouldn't cause other issues, but let me know if it does..

CHANGELOG.rst
kajol-patel: Done!

tubular/numeric.py
kajol-patel: Done [here](https://github.com/azukds/tubular/pull/641/commits/310037a38d29b63562d28829758294241d6726d1)

Issue #642, opened 2025-12-10
bug: need to update the feature table CI stage to run only on the result of a PR


liamholmes31:
What?
The feature table CI stage currently runs on the incoming changes, but does not take into account other changes in main. This means if e.g. a new feature is added in main but not included in the PR, CI will pass for the PR but then build will fail in main.

Why?
fix misleading CI

How?
update CI steps to run on RESULT of pr merge instead of just PR diffs

PR #643, opened 2025-12-10
updated CI to run on result of merge

liamholmes31:
https://github.com/azukds/tubular/issues/642

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -22,12 +22,10 @@ jobs:
 
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


utils/feature_table.py
```diff
@@ -2,13 +2,13 @@
 
 import pathlib
 import re
+import sys
 
 import pandas as pd
 
+sys.path.append(str(pathlib.Path(__file__).resolve().parents[1]))
 from tubular.base import FEATURE_REGISTRY
```
liamholmes31, line 10: I'm not sure why this import broke, as it worked when the feature table PR went in.. I've just done a quick fix

PR #645, opened 2025-12-10
turned on jsonable for cappingtransformer

liamholmes31:
turned on jsonable for CappingTransformer https://github.com/azukds/tubular/issues/637

### DIFF
--- a/tubular/capping.py
+++ b/tubular/capping.py
@@ -617,12 +617,16 @@ class BaseCappingTransformer(BaseNumericTransformer, WeightColumnMixin):
             },
         )
 
-        data["fit"].update(
-            {
-                "quantile_capping_values": self.quantile_capping_values,
-                "_replacement_values": self._replacement_values,
-            },
-        )
+        # transformer only fits for quantiles setting
+        if self.quantiles is not None:
+            self.check_is_fitted(["quantile_capping_values", "_replacement_values"])
+
+            data["fit"].update(
+                {
+                    "quantile_capping_values": self.quantile_capping_values,
+                    "_replacement_values": self._replacement_values,
+                },
+            )
 
         return data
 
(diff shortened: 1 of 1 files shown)


tubular/capping.py
```diff
@@ -699,7 +699,7 @@ class attribute, indicates whether transformer works with lazyframes
 
     FITS = True
 
-    jsonable = False
+    jsonable = True
```
Ritunjai-Sharma, line 715: Do we not need to add the parameterized _handle_from_json test for capping transformer?

CHANGELOG.rst
```diff
@@ -43,6 +43,7 @@ Changed
 - feat: added auto generated feature table to readme, to track transform support for e.g. polars/lazyframes/json `#595 <https://github.com/azukds/tubular/issues/595>`_
 - feat: added FEATURE_REGISTRY dict maintained by BaseTransformer, which records per-transformer functionality info
 - feat: added `deprecated` class attr to deprecated classes, useful to filter them out of e.g. feature table
+- feat: turned on jsonable for  CappingTransformer `#637 <https://github.com/azukds/tubular/issues/637>`_
```
Ritunjai-Sharma, line 46: Just for my own clarity, we will be adding the to_json method and updating the docstring for CappingTransformer with example of how to use to_json in a separate PR?

liamholmes31: so the to_json method was added for BaseCappingTransformer [here](https://github.com/azukds/tubular/pull/594), and just gets inherited

I forgot the doctest, thanks! and adding did expose a slight bug with to_json so addressed both here https://github.com/azukds/tubular/pull/645/commits/25f2bcf3fee55955fd8bb032be94600ac4a24fb4

tubular/capping.py
liamholmes31: so if you look in tests/capping/test_CappingTransformer, there actually aren't any specific CappingTransformer transform tests - they all get inherited from the BaseCappingTransformer tests (which have already been setup for json), so there's nothing more to change in this one

CHANGELOG.rst
Ritunjai-Sharma: Good to see doctests helping us catch bugs :)

Issue #633, opened 2025-12-02
add CopyColumnTransformer to misc file


liamholmes31:
What?
it can sometimes be useful to have a couple of versions of a given column to work with, given that some tubular transformers overwrite the initial column. For example a user might have a datetime column that they want to transform into year values with ToDatetimeTransformer, but also to still work with the datetime strs elsewhere.

it would be very simple to add a CopyColumnsTransformer to the misc file, with logic like `nw.col(col).alias(f"{col}_copy")` for each expression.

Why?
make it easier to work around transformers that overwrite the input columns (admittedly in future we may want to just change this overwriting behaviour, but this is a quick/easy way to provide a workaround)

How?
Create new transformer as described above - will want the transformer to be written in narwhals and to incorporate our new json/lazyframe functionality

PR #647, opened 2025-12-11
added RenameColumnsTransformer to misc module

liamholmes31:
set up simple transformer for renaming (or copying columns) https://github.com/azukds/tubular/issues/633

a bit niche, but have seen cases where it is/could be used so worth just bringing in I think

### DIFF
--- a/tests/misc/test_RenameColumnsTransformer.py
+++ b/tests/misc/test_RenameColumnsTransformer.py
@@ -63,12 +63,17 @@ class TestTransform(GenericTransformTests):
         "library",
         ["pandas", "polars"],
     )
+    @pytest.mark.parametrize(
+        "empty",
+        [True, False],
+    )
     def test_output(
         self,
         library,
         from_json,
         lazy,
         drop_original,
+        empty,
     ):
         "test outputs for RenameColumnsTransformer.transform"
 
@@ -78,7 +83,10 @@ class TestTransform(GenericTransformTests):
             drop_original=drop_original,
         )
 
-        df_dict = {"a": [1, None, 3], "b": ["x", "y", None]}
+        df_dict = {
+            "a": [1, None, 3] if not empty else [],
+            "b": ["x", "y", None] if not empty else [],
+        }
 
         expected_df_dict = {
             "new_a": df_dict["a"],


tests/pipeline/test_pipeline_dump_and_load_json.py
```diff
@@ -13,8 +13,7 @@ class FakeTransformer:
 class TestPipelineDumpAndLoadJson:
     """Tests for dump_pipeline_to_json() and load_pipeline_from_json()."""
 
-    @staticmethod
-    def test_dump_pipeline_then_load_pipeline():
+    def test_dump_pipeline_then_load_pipeline(self):
```
liamholmes31, line 16: just cleaning up the leftover staticmethods spotted here, a previous PR removed them from the rest of the test suite

tubular/misc.py
```diff
@@ -165,6 +166,235 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+
+        Raises
+        ------
+            ValueError: if provided columns are not keys of
+                provided new_column_names
+
+        """
+        super().__init__(columns=columns, **kwargs)
+
+        msg = f"{self.classname()}: all provided columns must appear as keys in new_column_names"  # noqa: E501
```
hariinduu, line 271: Q: What could happen if you provide an existing column name as a new_column name?

tests/misc/test_RenameColumnsTransformer.py
```diff
@@ -0,0 +1,133 @@
...
+
+class TestFit(GenericFitTests):
+    """Generic tests for RenameColumnsTransformer.fit()"""
+
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "RenameColumnsTransformer"
+
+
+class TestTransform(GenericTransformTests):
```
hariinduu, line 43: S: Should we add a test with an empty data frame to see whether it still renaming the column?

tests/misc/test_RenameColumnsTransformer.py
hariinduu, line 1: Q: Also thinking whether we should test special scenarios such as  case sensitivity  (For example a -> A ) , special characters in new_column names, python keywords in new_column names and overlapping it with indexes

tubular/misc.py
liamholmes31: good question, added handling here https://github.com/azukds/tubular/pull/647/changes/ff1cc05e7f8c31a3b72f2562fba9e698b058948c

tests/misc/test_RenameColumnsTransformer.py
liamholmes31: updated the test case here https://github.com/azukds/tubular/pull/647/changes/ff1cc05e7f8c31a3b72f2562fba9e698b058948c

tests/misc/test_RenameColumnsTransformer.py
liamholmes31: added here https://github.com/azukds/tubular/pull/647/changes/853adadce20aa94c49eb74887d6657d1abbb36db

PR #650, opened 2025-12-12
Added two new transformers

hariinduu:
Added two new transformers called WhenThenOtherwiseTransformer and CompareTwoColumnsTransformer

### DIFF
--- a/tubular/comparison.py
+++ b/tubular/comparison.py
@@ -93,6 +93,8 @@ class WhenThenOtherwiseTransformer(BaseTransformer):
     │ 30  ┆ 30  ┆ true          ┆ 30         │
     └─────┴─────┴───────────────┴────────────┘
 
+    ```
+
     """
 
     polars_compatible = True
@@ -142,11 +144,14 @@ class WhenThenOtherwiseTransformer(BaseTransformer):
 
         Examples
         --------
+        ```pycon
         >>> from pprint import pprint
         >>> transformer = WhenThenOtherwiseTransformer(
-        ...     columns=["a", "b"],when_column="condition_col",then_column="update_col"
+        ...     columns=["a", "b"],
+        ...     when_column="condition_col",
+        ...     then_column="update_col",  # noqa: E501
         ... )
-        >>> pprint(transformer.to_json(),sort_dicts=True)
+        >>> pprint(transformer.to_json(), sort_dicts=True)
         {'classname': 'WhenThenOtherwiseTransformer',
          'fit': {},
          'init': {'columns': ['a', 'b'],
@@ -155,7 +160,9 @@ class WhenThenOtherwiseTransformer(BaseTransformer):
                   'then_column': 'update_col',
                   'verbose': False,
                   'when_column': 'condition_col'},
-         'tubular_version': 'dev'}
+         'tubular_version': ...}
+
+        ```
 
         """
         json_dict = super().to_json()
@@ -221,6 +228,8 @@ class WhenThenOtherwiseTransformer(BaseTransformer):
         │ 2   ┆ 5   ┆ false         ┆ 20         │
         │ 30  ┆ 30  ┆ true          ┆ 30         │
         └─────┴─────┴───────────────┴────────────┘
+
+        ```
 
         """
         X = _convert_dataframe_to_narwhals(X)
@@ -345,6 +354,7 @@ class CompareTwoColumnsTransformer(BaseTransformer):
 
         Examples
         --------
+        ```pycon
         >>> transformer = CompareTwoColumnsTransformer(
         ...     columns=["a", "b"],
         ...     condition=ConditionEnum.GREATER_THAN.value,
@@ -359,7 +369,9 @@ class CompareTwoColumnsTransformer(BaseTransformer):
                   'copy': False,
                   'return_native': True,
                   'verbose': False},
-         'tubular_version': 'dev'}
+         'tubular_version': ...}
+
+        ```
 
         """
         json_dict = super().to_json()
@@ -389,6 +401,7 @@ class CompareTwoColumnsTransformer(BaseTransformer):
 
         Examples
         --------
+        ```pycon
         >>> import polars as pl
         >>> df = pl.DataFrame({"a": [1, 2, 3], "b": [3, 2, 1]})
         >>> transformer = CompareTwoColumnsTransformer(
@@ -407,6 +420,8 @@ class CompareTwoColumnsTransformer(BaseTransformer):
         │ 2   ┆ 2   ┆ false │
         │ 3   ┆ 1   ┆ true  │
         └─────┴─────┴───────┘
+
+        ```
 
         """
         X = _convert_dataframe_to_narwhals(X)


tubular/comparison.py
```diff
@@ -3,18 +3,296 @@
...
+        **kwargs: dict[str, bool]) -> None:
+        """
+        Initialize the WhenThenOtherwiseTransformer.
+
+        Parameters
+        ----------
+        columns : ListOfMoreThanOneStrings
+            List of columns to be transformed.
+
+        condition_column : bool
```
liamholmes31: N: maybe could name condition_column->when_column and updated_values_column->then_column

tubular/comparison.py
```diff
@@ -3,18 +3,296 @@
...
+        1     2     5         False         20
+        2    30    30          True         30
+
+        ```
+
+        """
+        X = _convert_dataframe_to_narwhals(X)
+        X = super().transform(X, return_native_override=False)
+
+        expr = {}
```
liamholmes31: N: would call this exprs_dict

tubular/comparison.py
```diff
@@ -3,18 +3,296 @@
...
+        for col in self.columns:
+            expr[col] = nw.when(nw.col(self.condition_column)).then(nw.col(self.updated_values_column)).otherwise(nw.col(col))
+        
+        X = X.with_columns(*expr.values())
```
liamholmes31: S: can do `X = X.with_columns(**expr)`

tubular/comparison.py
```diff
@@ -3,18 +3,296 @@
...
+        0    10    10          True         10
+        1     2     5         False         20
+        2    30    30          True         30
+
+        ```
+
+        """
+        X = _convert_dataframe_to_narwhals(X)
+        X = super().transform(X, return_native_override=False)
+
```
liamholmes31, line 237: S: can we include some type checks (using `schema=X.schema`):

- schema[condition_col]==nw.Boolean
- all(schema[col]==schema[update_col])
    - in the future we might make this more relaxed so that e.g. schema[col] could be int32 and schema[update_col] could be int64, but let's start simple

liamholmes31: ^note, the above checks are lazyframe compatible as they work with the schema

tubular/comparison.py
```diff
@@ -3,18 +3,296 @@
...
+    FITS = False
+    jsonable = True
+    lazyframe_compatible = True
+
+    @beartype
+    def __init__(
+        self,
+        columns: ListOfTwoStrs, 
+        condition: str, 
+        outcome: str,
```
liamholmes31: S: can we drop the outcome arg and just call the new column like f'{col1}{operation}{col2}'

tubular/comparison.py
```diff
@@ -3,18 +3,296 @@
...
+    polars_compatible = True
+    FITS = False
+    jsonable = True
+    lazyframe_compatible = True
+
+    @beartype
+    def __init__(
+        self,
+        columns: ListOfTwoStrs, 
+        condition: str,
```
liamholmes31: was imagining this one a bit differently, where condition would be e.g. '>', '<', '==', '!=', ... (best to setup an enum for this

tubular/comparison.py
```diff
@@ -3,18 +3,296 @@
...
+           col1  col2  comparison_result
+        0     1     3                  0
+        1     2     2                  0
+        2     3     1                  1
+        ```
+        """
+        X = _convert_dataframe_to_narwhals(X)
+        X = super().transform(X, return_native_override=False)
+
+        expr = nw.when(nw.expr(self.condition)).then(1).otherwise(0)
```
liamholmes31: with the new condition, the expr could be:
```
import operator
_OPS={
">": operator.gt,
"<": operator.lt,
...
}
...
expr=_OPS[condition](nw.col(columns[0], nw.col(columns[1])
```

^ I haven't worked with the operator module before, see if that works!

tubular/comparison.py
hariinduu: Renamed

tubular/comparison.py
hariinduu: Renamed

tubular/comparison.py
hariinduu: Modified

tubular/comparison.py
hariinduu: Added type checks

tubular/comparison.py
hariinduu: Added inside types file

tubular/comparison.py
hariinduu: Modified

tubular/comparison.py
hariinduu: Added

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+    lazyframe_compatible : bool
+        Indicates whether transformer works with lazyframes.
+
+    Examples
+    --------
+    ```pycon
+    >>> import pandas as pd
+    >>> from tubular.base import BaseTransformer
+
+    >>> df = pd.DataFrame(
```
liamholmes31: N: can you use polars for examples plz, want to steer people towards polars

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+
 
 if TYPE_CHECKING:
     import pandas as pd
 
 
+class WhenThenOtherwiseTransformer(BaseTransformer):
+    """Transformer to apply conditional logic across multiple columns.
+
+    This transformer evaluates specified columns within a DataFrame and applies
```
liamholmes31: would the simplify wording a bit
```suggestion
    This transformer evaluates specified columns against a condition and updates with given values based on the results.
```

tubular/comparison.py
liamholmes31: (same comment for other doctest examples too)

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+    @beartype
+    def transform(
+        self,
+        X: DataFrame,
+    ) -> DataFrame:
+        """Apply conditional logic to transform specified columns.
+
+        Parameters
+        ----------
+        X : pd.DataFrame
```
liamholmes31: can you change the types in the  docstrings to 'DataFrame' to match what's elsewhere

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+        """
+        X = _convert_dataframe_to_narwhals(X)
+        X = super().transform(X, return_native_override=False)
+
+        schema = X.schema
+        if schema[self.when_column] != nw.Boolean:
+            raise TypeError(f"The column '{self.when_column}' must be of type Boolean.")
+
+        then_column_type = schema[self.then_column]
+        for col in self.columns:
```
liamholmes31: N: think it's generally  more efficient to do comprehensions than  loops, so like:
`if any(schema[col]!=then_column_type for col in self.columns):`

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+            if schema[col] != then_column_type:
+                raise TypeError(
+                    f"The column '{col}' must be of the same type as '{self.then_column}'."
+                )
+
+        exprs_dict = {}
+        for col in self.columns:
```
liamholmes31: S: similar here, think it would be better to write as a dict comprehension

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+    --------
+    ```pycon
+    >>> import pandas as pd
+    >>> from tubular.base import BaseTransformer
+
+    >>> df = pd.DataFrame({"col1": [1, 2, 3], "col2": [3, 2, 1]})
+
+    >>> transformer = CompareTwoColumnsTransformer(
+    ...     columns=["col1", "col2"],
+    ...     condition='row["col1"] > row["col2"]',
```
liamholmes31: looks like this might be out of date?

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+
+        """
+        json_dict = super().to_json()
+
+        json_dict["init"]["condition"] = self.condition.name
+
+        return json_dict
+
+    @classmethod
+    def from_json(
```
liamholmes31: Ah I see the problem you were having, this is a good solution - but I would suggest match what's been done elsewhere, e.g. in the dates file:
```
class DatetimeInfoOptions(str, Enum):
    """Options for what is returned by DatetimeInfoExtractor."""

    __slots__ = ()

    TIME_OF_DAY = "timeofday"
    TIME_OF_MONTH = "timeofmonth"
    TIME_OF_YEAR = "timeofyear"
    DAY_OF_WEEK = "dayofweek"


DatetimeInfoOptionStr = Annotated[
    str,
    Is[lambda s: s in DatetimeInfoOptions._value2member_map_],
]
```

^ if you make ConditionEnumStr type and hint with that instead, I think things will work without having to overload from_json

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+        1     2     2                  0
+        2     3     1                  1
+        ```
+
+        """
+        X = _convert_dataframe_to_narwhals(X)
+        X = super().transform(X, return_native_override=False)
+
+        # Map the enum to the operator functions
+        ops_map = {
```
liamholmes31: N: better to define this as a class attr outside the methods, as it doesn't need to be defined in transform

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+            ConditionEnum.GREATER_THAN: operator.gt,
+            ConditionEnum.LESS_THAN: operator.lt,
+            ConditionEnum.EQUAL_TO: operator.eq,
+            ConditionEnum.NOT_EQUAL_TO: operator.ne,
+        }
+
+        expr = (
+            nw.when(
```
liamholmes31: Q: what happens if you write this without the when/then/otherwise? so just like
```
ops_map[self.condition](
                    nw.col(self.columns[0]), nw.col(self.columns[1])
                )
```

tubular/comparison.py
```diff
@@ -2,19 +2,385 @@
...
+        2     3     1                  1
+        ```
+
+        """
+        X = _convert_dataframe_to_narwhals(X)
+        X = super().transform(X, return_native_override=False)
+
+        # Map the enum to the operator functions
+        ops_map = {
+            ConditionEnum.GREATER_THAN: operator.gt,
```
liamholmes31: As these are all numeric operations, perhaps we should restrict to numeric types for now - could you add a check/error for this using schema?

tests/test_data.py
```diff
@@ -671,3 +671,15 @@ def expected_date_diff_df_2(library="pandas"):
     }
 
     return u.dataframe_init_dispatch(df_dict, library=library)
+
+
+
+def create_when_then_test_df(library="pandas"):
```
liamholmes31: S: would put this in test_WhenThenOtherwiseTransformer.py as  it's unlikely to be reused elsewhere

liamholmes31: actually  looks like it's already there, so can just delete here

tests/comparison/test_WhenThenOtherwiseTransformer.py
```diff
@@ -0,0 +1,286 @@
...
+            ([], "condition_col", "update_col"), 
+            (["a"], "condition_col", "update_col"),
+            (None, "condition_col", "update_col"),
+            ("a", "condition_col", "update_col"),
+            (123, "condition_col", "update_col"),
+            (["a", "b"], "condition_col", "update_col"),
+            (["a", "b", "c"], "condition_col", "update_col"),
+        ],
+    )
+    def test_errors_if_invalid_columns(
```
liamholmes31: this test isn't needed, as a similar test is  inherited through ColumnStrListInitTests

tests/comparison/test_WhenThenOtherwiseTransformer.py
```diff
@@ -0,0 +1,286 @@
...
+
+        u.assert_frame_equal_dispatch(
+            u._collect_frame(transformed_df, lazy),
+            expected_df,
+        )
+
+    @pytest.mark.parametrize("lazy", [True, False])
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize("from_json", [True, False])
+    def test_with_nulls(
```
liamholmes31: N: could delete this test by adding some rows with nulls to create_when_then_test_df

tests/comparison/test_WhenThenOtherwiseTransformer.py
```diff
@@ -0,0 +1,286 @@
...
+        args["when_column"] = when_column
+        args["then_column"] = then_column
+        with pytest.raises(BeartypeCallHintParamViolation):
+            uninitialized_transformers[self.transformer_name](**args)
+
+
+
+class TestWhenThenOtherwiseTransformerTransform():
+    """Tests for transform method in WhenThenOtherwiseTransformer."""
+
```
liamholmes31, line 103: Can you also add tests  to check for the type checking errors  in transform

tests/comparison/test_CompareTwoColumnsTransformer.py
```diff
@@ -0,0 +1,301 @@
...
+            ([], ConditionEnum.GREATER_THAN),
+            (["col1"], ConditionEnum.GREATER_THAN),
+            (None, ConditionEnum.GREATER_THAN),
+            ("col1", ConditionEnum.GREATER_THAN),
+            (123, ConditionEnum.GREATER_THAN),
+            (["col1", "col2"], ConditionEnum.GREATER_THAN),
+            (["col1", "col2", "col3"], ConditionEnum.GREATER_THAN),
+        ],
+    )
+    def test_errors_if_invalid_columns(
```
liamholmes31: don't need this one due to  inheritance

tests/comparison/test_CompareTwoColumnsTransformer.py
```diff
@@ -0,0 +1,301 @@
...
+        minimal_attribute_dict,
+        uninitialized_transformers,
+        from_json,
+        lazy,
+    ):
+        """Test transform method performs comparison correctly."""
+        args = copy.deepcopy(minimal_attribute_dict[self.transformer_name])
+        # args["kwargs"] = {}
+
+        print("Debug: Transformer args:", args)
```
liamholmes31: looks like this can be deleted

tubular/comparison.py
liamholmes31: would also make sense to cast to Boolean type

tests/comparison/test_CompareTwoColumnsTransformer.py
```diff
@@ -0,0 +1,301 @@
...
+            )
+            .to_native()
+        )
+
+        u.assert_frame_equal_dispatch(
+            u._collect_frame(transformed_df, lazy),
+            expected_df,
+        )
+
+    @pytest.mark.parametrize("lazy", [True, False])
```
liamholmes31, line 162: would delete this test and add some null rows to the main test df in this file

tests/comparison/test_CompareTwoColumnsTransformer.py
```diff
@@ -0,0 +1,301 @@
...
+    ):
+        args = minimal_attribute_dict[self.transformer_name].copy()
+        args["condition"] = condition
+        with pytest.raises(BeartypeCallHintParamViolation):
+            uninitialized_transformers[self.transformer_name](**args)
+
+
+class TestCompareTwoColumnsTransformerTransform:
+    """Tests for the transform method of CompareTwoColumnsTransformer."""
+
```
liamholmes31, line 49: will need to  include type error tests  once we add those checks

tests/comparison/test_CompareTwoColumnsTransformer.py
```diff
@@ -0,0 +1,301 @@
...
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "CompareTwoColumnsTransformer"
+
+    @pytest.mark.parametrize("lazy", [True, False])
```
liamholmes31, line 106: could  you add another parametrize  and include tests for the different operations? we need to cover them all really

(same applies for  test_single_row)

tubular/comparison.py
hariinduu: Changed in both places

tubular/comparison.py
hariinduu: Changed

tubular/comparison.py
hariinduu: Modified

tubular/comparison.py
hariinduu: Changed here as well

tubular/comparison.py
hariinduu: Ah yes, updated

tubular/comparison.py
hariinduu: Tried something like below. But still I think it treats None as less than any number resulting 1>Null -> 1
To fix that again have to introduce some conditional logic. What do you think?

expr = self.ops_map[self.condition](
            nw.col(self.columns[0]).fill_null(0), nw.col(self.columns[1]).fill_null(0)
        ).cast(nw.Boolean)

tubular/comparison.py
hariinduu: Added

tubular/comparison.py
hariinduu: Moved as a class attribute

tests/comparison/test_WhenThenOtherwiseTransformer.py
hariinduu: Deleted

tests/test_data.py
hariinduu: Deleted

tests/comparison/test_WhenThenOtherwiseTransformer.py
hariinduu: Added three rows covering different cases and deleted

tubular/comparison.py
hariinduu: Tried this way, Still de-serialisation is not working properly as soon as I delete from_json method

tests/comparison/test_CompareTwoColumnsTransformer.py
hariinduu: Deleted

tests/comparison/test_CompareTwoColumnsTransformer.py
hariinduu: Deleted

tests/comparison/test_CompareTwoColumnsTransformer.py
hariinduu: Added two more rows and deleted this

tubular/comparison.py
hariinduu: This is working now

tubular/comparison.py
hariinduu: Modified as below and it is working now

        expr = (
                    nw.when(
                        ~null_filter_expr
                    )
                    .then(self.ops_map[self.condition](
                            nw.col(self.columns[0]), nw.col(self.columns[1])
                        ))
                    .otherwise(None)
                )

        backend=nw.get_native_namespace(X).__name__

        if backend=='polars':

            expr=expr.cast(nw.Boolean)

        outcome_column_name = (
            f"{self.columns[0]}{self.condition.value}{self.columns[1]}"
        )

        X = X.with_columns(expr.alias(outcome_column_name))

        if backend=='pandas':
            X = X.with_columns(
                nw.maybe_convert_dtypes(X[outcome_column_name]).cast(
                    nw.Boolean
                )
            )

tests/comparison/test_CompareTwoColumnsTransformer.py
hariinduu: Added covering all the operations

tests/comparison/test_WhenThenOtherwiseTransformer.py
hariinduu: Added tests covering type checks

tests/comparison/test_CompareTwoColumnsTransformer.py
```diff
@@ -0,0 +1,218 @@
...
+
+class TestCompareTwoColumnsTransformerTransform:
+    """Tests for the transform method of CompareTwoColumnsTransformer."""
+
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "CompareTwoColumnsTransformer"
+
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_transform_raises_error_on_non_numeric_column_type(
```
liamholmes31, line 55: N: would combine this test and the next one using parametrize, they're quite similar

tubular/comparison.py
```diff
@@ -2,17 +2,446 @@
...
+        │ 3   │ 1   │ 1             │
+        └─────┴─────┴───────────────┘
+        ```
+
+        """
+        X = _convert_dataframe_to_narwhals(X)
+        X = super().transform(X, return_native_override=False)
+
+        schema = X.schema
+        numeric_types = {nw.Int64, nw.Float64, nw.Int32, nw.Float32}
```
liamholmes31: S: can reuse the NumericTypes list from tubular.types instead of defining here

tubular/comparison.py
```diff
@@ -2,17 +2,446 @@
...
+        Indicates whether transformer works with lazyframes.
+
+    Examples
+    --------
+    ```pycon
+    >>> import polars as pl
+    >>> df = pl.DataFrame({"a": [1, 2, 3], "b": [3, 2, 1]})
+    >>> transformer = CompareTwoColumnsTransformer(
+    ...     columns=["a", "b"],
+    ...     condition=ConditionEnum.GREATER_THAN,
```
liamholmes31: better to demo the string inputs, as this is how most users will use it
```suggestion
    ...     condition=">",
```

tubular/comparison.py
```diff
@@ -2,17 +2,446 @@
...
+            JSON representation of the transformer, including init parameters.
+
+        Examples
+        --------
+        >>> transformer = WhenThenOtherwiseTransformer(
+        ...     columns=['a', 'b'],
+        ...     when_column='condition_col',
+        ...     then_column='update_col'
+        ... )
+        >>> transformer.to_json()
```
liamholmes31: duplicate
```suggestion
```

tubular/comparison.py
```diff
@@ -2,17 +2,446 @@
...
+        TypeError
+            If the columns are not of a numeric type.
+
+        Examples
+        --------
+        >>> import polars as pl
+        >>> df = pl.DataFrame({"a": [1, 2, 3], "b": [3, 2, 1]})
+        >>> transformer = CompareTwoColumnsTransformer(
+        ...     columns=["a", "b"],
+        ...     condition=ConditionEnum.GREATER_THAN,
```
liamholmes31:
```suggestion
        ...     condition=">",
```

tubular/comparison.py
```diff
@@ -2,17 +2,446 @@
...
+        └─────┴─────┴───────────────┘
+        ```
+
+        """
+        X = _convert_dataframe_to_narwhals(X)
+        X = super().transform(X, return_native_override=False)
+
+        schema = X.schema
+        numeric_types = {nw.Int64, nw.Float64, nw.Int32, nw.Float32}
+        for col in self.columns:
```
liamholmes31: S: better like `if any(schema[col] not in numeric_types) for col in self.columns`

tests/comparison/test_CompareTwoColumnsTransformer.py
hariinduu: Combined

tubular/comparison.py
hariinduu: Modified

tubular/comparison.py
hariinduu: If we keep it that way, we won't be able to see which column is having numeric types in the error. So kept it like this. what do you think?

liamholmes31: good point, would do it in two steps:
```
bad_cols=[col for col in self.columns if schema[col] not in numeric_types)]
if bad_cols:
    message = f"Columns must be of a numeric type, but the following are not: {bad_cols}"
    raise TypeError(message)
```

tubular/comparison.py
```diff
@@ -2,17 +2,445 @@
...
+        self.condition = ConditionEnum(condition)
+
+    def to_json(self) -> dict[str, dict[str, Any]]:
+        """Serialize the transformer to a JSON-compatible dictionary.
+
+        Returns
+        -------
+        dict[str, dict[str, Any]]:
+            JSON representation of the transformer, including init parameters.
+
```
liamholmes31, line 354: S: remember the doctest here

tubular/comparison.py
```diff
@@ -2,17 +2,445 @@
...
+        Examples
+        --------
+        >>> transformer = WhenThenOtherwiseTransformer(
+        ...     columns=['a', 'b'],
+        ...     when_column='condition_col',
+        ...     then_column='update_col'
+        ... )
+        >>> transformer.to_json()
+        {
```
liamholmes31: doctest is erroring because or formatting here - i'm going to tidy this everywhere in a later PR, so would just put the output as it appears in your terminal (will all be in one line)

tubular/comparison.py
```diff
@@ -2,17 +2,445 @@
...
+    ...     }
+    ... )
+    >>> transformer = WhenThenOtherwiseTransformer(
+    ...     columns=["a", "b"], when_column="condition_col", then_column="update_col"
+    ... )
+    >>> transformed_df = transformer.transform(df)
+    >>> print(transformed_df)
+    shape: (3, 4)
+    ┌─────┬─────┬──────────────┬───────────┐
+    │ a   │ b   │ condition_col│ update_col│
```
liamholmes31: some of these doctests are out of date now that we've switched to bool

tubular/comparison.py
hariinduu: Modified this as well as tests

tubular/comparison.py
hariinduu: Modified

tubular/comparison.py
hariinduu: I think doctests are up to date

tubular/comparison.py
hariinduu: added

tubular/comparison.py
```diff
@@ -2,17 +2,457 @@
...
+        >>> pprint(transformer.to_json(),sort_dicts=True)
+        {'classname': 'WhenThenOtherwiseTransformer',
+         'fit': {},
+         'init': {'columns': ['a', 'b'],
+                  'copy': False,
+                  'return_native': True,
+                  'then_column': 'update_col',
+                  'verbose': False,
+                  'when_column': 'condition_col'},
+         'tubular_version': 'dev'}
```
liamholmes31: version is picked up as being different in CI, so can switch to the wildcard '...'
```suggestion
         'tubular_version': ...}
```

tests/comparison/test_CompareTwoColumnsTransformer.py
liamholmes31: can't see a change on this one?

Issue #529, opened 2025-10-08
make MappingTransformer lazyframe compatible


liamholmes31:
What?
this transformer currently inspects the values present in the data, to:
- warn about unused mappings/unmapped values in data, suggest this is one we just drop
- to filter down mapping conditions to only those needed for efficiency

in order to allow lazyframes we need to either:
1) find a way of removing these that is still time efficient for single rows
2) add in a .collect call to allow values to be inspected

I think 1 should be preferred, but may be challenging as a chunk of time has already been spent trying to optimise!

Why?
allow use of lazyframes

How?
One idea to explore would be converting mappings into a lookup table and joining in order to map - I think this has the benefit that the lookup can be converted to lazy, and then filtered on a lazy .select of the present values in the main df, which partially preserves the current behaviour.

PR #652, opened 2025-12-15
made MappingTransformer classes lazyframe compatible

liamholmes31:
had to:
- remove some warnings (think these were generally more of an annoyance anyway..)
- refine the boolean type handling to just single out pandas dfs

https://github.com/azukds/tubular/issues/529

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,8 +33,10 @@ Each individual change should have a link to the pull request after the descript
 Changed
 ^^^^^^^
 - feat: turned on lazyframe_compatible for MappingTransformer and base classes, and setup lazy tests for these `#529 <https://github.com/azukds/tubular/issues/529>`_
+- added `to_json` method for BaseAggregationTransformer `#610 <https://github.com/azukds/tubular/issues/610>_`
 - feat: first pass refactor of MeanResponseTransformer `#149 <https://github.com/azukds/tubular/issues/149>`_
 - turned on PT ruff ruleset `#114 <https://github.com/azukds/tubular/issues/114>`_
+- feat: added two new transformers called WhenThenOtherwiseTransformer and CompareTwoColumnsTransformer
 
 2.6.0 (19/12/2025)
 ------------------
--- a/README.md
+++ b/README.md
@@ -54,20 +54,90 @@ transformer.transform(test_df)
 # │ 20  ┆ 1   ┆ 4   │
 # └─────┴─────┴─────┘
 ```
+Tubular also supports saving/reading transformers and pipelines to/from json format (goodbye .pkls!), which we demo below:
 
-We are currently in the process of rolling out support for some new features:
-- to/from json methods for our transformers to allow json storage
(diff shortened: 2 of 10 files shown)


tubular/mapping.py
```diff
@@ -545,26 +547,6 @@ def transform(
 
         X = BaseTransformer.transform(self, X, return_native_override=False)
 
-        mapped_columns = self.mappings.keys()
```
adamsardar, line 548: P: Completely agree with deleting these checks. They're trying to be helpful, but I think that we should focus on performance over expensive checks.

tests/mapping/test_BaseMappingTransformMixin.py
```diff
@@ -248,10 +252,11 @@ def test_expected_output_boolean_with_nulls(self, library):
             expected = expected.with_columns(nw.maybe_convert_dtypes(expected["a"]))
             expected = expected.to_native()
 
-        assert_frame_equal_dispatch(expected, df_transformed)
+        assert_frame_equal_dispatch(expected, _collect_frame(df_transformed, lazy=lazy))
```
adamsardar, line 255: Q: Am I right that we never test that the frame is lazy? We collect it and then compare against `expected`?

liamholmes31: yeah the return type isn't explicity tested - I was about to say that it's implicitly tested by _collect_frame, but then spotted a typo which meant this wasn't true, but have now fixed here

https://github.com/azukds/tubular/pull/652/commits/4e1b218b28be7e159e1f874da1efa024c16a61f8

i'm keen for us to commit to just always returning lazyframes in the future, as this will make it easier to turn on a static type checker, issue for this here https://github.com/azukds/tubular/issues/668

tests/mapping/test_BaseMappingTransformMixin.py
```diff
@@ -324,60 +328,22 @@ def test_mappings_unchanged(self, mapping, library):
...
-        x_fitted = transformer.fit(df, df["c"])
-
-        with pytest.raises(
-            BeartypeCallHintParamViolation,
-        ):
-            x_fitted.transform(X=non_df)
-
-    @pytest.mark.parametrize("library", ["pandas", "polars"])
-    def test_original_df_not_updated(self, mapping, library):
+    def test_original_df_not_updated(self, mapping, library, lazy):
```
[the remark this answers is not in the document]
adamsardar: Need to find if I can turn this off. Too much text.

tests/mapping/test_BaseMappingTransformMixin.py
adamsardar: Couldn't agree more. LazyFrames are for closers.

adamsardar:
> changlog conflict resolved here [3ec0baf](https://github.com/azukds/tubular/commit/3ec0baff581fc2342ee79ffc8dcfe9dfff5a0c31)

Still a MC on this PR

Issue #528, opened 2025-10-08
make WeightColumnMixin lazyframe compatible


liamholmes31:
What?
Current blockers with this one are:
1) checks for >0
2) checks for nulls
3) checks for inf
4) checks not all 0
5) if 'unit_weights_column' exists, checks is all 1

these all require materialising data, so to allow lazyframe support we need to change this

Why?
Allow lazyframe support

How?
NOTE - leaving for visibility, but this is out of date. See latest comments!
~~Initial thoughts:~~
~~1) we could just add a .collect call to the .select that checks these, but in a pipeline this could become expensive. Also often the same weight column is going through the same checks for every transformer in a pipeline, which already seems excessive.
2) we could just edit the logic to replace bad weight values with None using expressions, and accept the output as being None for these rows~~

~~I think 2 is better, as moves us away from doing lots of unnecessary materialisations~~

liamholmes31:
this will in turn make the capping module lazyframe compatible, and be a big help to getting the imputers and nominal modules there

PR #654, opened 2025-12-17
Feature/lazy weightsmixin

liamholmes31:
Converted WeightColumnMixin to pave way for lazyframe work, slightly heavy one.. changes involved:

- splitting checks performed by this mixin into schema based ones and expression based ones, in two functions
- the expr-based method now returns the expressions, which transformers use to filter out invalid rows without erroring
- if all rows are removed, fit can return nonsense results, so added _check_for_failed_fit methods to error in this case
- added inheritable test for the above error
- added to existing test cases to include cases where some rows are filtered out
- had to convert OrdinalEncodingTransformer.fit to narwhals as it uses the weight mixin

https://github.com/azukds/tubular/issues/528

### DIFF
--- a/tests/mixins/test_WeightColumnMixin.py
+++ b/tests/mixins/test_WeightColumnMixin.py
@@ -184,3 +184,21 @@ class TestGetValidWeightsFilterExpr:
         df = df.filter(filter_expr)
 
         assert_frame_equal_dispatch(df.to_native(), expected_df.to_native())
+
+    @pytest.mark.parametrize("verbose", [True, False])
+    def test_warning(self, verbose, recwarn):
+        "test expected warning is given (depending on verbose)"
+
+        obj = WeightColumnMixin()
+
+        obj.get_valid_weights_filter_expr(weights_column="weights", verbose=verbose)
+
+        if not verbose:
+            assert len(recwarn) == 0
+
+        else:
+            assert len(recwarn) == 1
+            assert (
+                str(recwarn[0].message)
+                == "Weights must be strictly positive, non-null, and finite - rows failing this will be filtered out."
+            )


tubular/mixins.py
ChristinaKatsiafa:
```suggestion
        Returns
        --------
```

ChristinaKatsiafa:
```suggestion
        Parameters
        ----------
```
It looks like our docstring is not consistent through the whole package.
Which docstring style do we use? Google or Numpy Style?
Also should we raise a ticket to make this consistent through the whole package?

ChristinaKatsiafa:
```suggestion
        Raises
```

ChristinaKatsiafa, line 189: S: should we allow the user to set the column name?
So if they already have a column with weights called weights then they can pass this is instead of having to rename it to make this compatible with the transformer.

tubular/mixins.py
```diff
@@ -241,50 +237,38 @@ def check_weights_column(self, X: FrameT, weights_column: str) -> None:
             ValueError:
                 if weights_column is non-numeric
 
-            ValueError:
-                if weights_column is not strictly positive, contains nulls, or is infinite
-
         """
+        X = _convert_dataframe_to_narwhals(X)
```
ChristinaKatsiafa, line 242: Q: Is this even needed here? Mixins will pass extra functionality to all the transformers that they will already convert the dfs to narwhals. Correct?
So my thought are:

- Is this not an extra step that can be avoided?
- If this is removed from here then we can pick errors from the transformers that will not have converted the dataframes to narwhals. Right?

So If those points are correct and I am not missing something, then I assume that removing this from the whole mixins area will save us extra checking (time & memory) and also help identify errors with the transformers.

What do you think?

tubular/mixins.py
```diff
@@ -241,50 +237,38 @@ def check_weights_column(self, X: FrameT, weights_column: str) -> None:
...
+        Returns
+        -------
+            nw.Expr: expression to be used for filtering down to valid weights rows
 
-        # check weight not inf
-        if not all_finite:
-            msg = f"{self.classname()}: weight column must not contain infinite values."
-            raise ValueError(msg)
+        """
+        expr_ge_0 = nw.col(weights_column) > 0
```
ChristinaKatsiafa, line 280: Q: Why are we filtering out the negative weights?

tubular/mixins.py
ChristinaKatsiafa:
```suggestion
```
Q: Is this still needed?

tubular/base.py
ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        y : None or Series, default = None
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

tubular/capping.py
```diff
@@ -220,6 +221,28 @@ def check_capping_values_dict(
...
+        Raises
+        ------
+            ValueError: if quantile_capping_values have fit as None where values were expected
+
+        """
+        for col in self.quantiles:
+            failed_columns = []
+            for i, init_value in enumerate(self.quantiles[col]):
+                fit_value = self.quantile_capping_values[col][i]
+                if (not pd.isna(init_value)) and pd.isna(fit_value):
```
ChristinaKatsiafa: S: Should we move away from pandas here?
For example this line can be converted to something like:
is_missing = lambda x: x is None or x != x
if not is_missing(init_value) and is_missing(fit_value):

tubular/capping.py
ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        y : Series or None. Defaults to None
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa, line 597: N: I know that this is not the scope of this PR but I would expect most of this to be part of the fit and at the end of the fit to store self.exprs = exprs
So the transform will not do any calculation at all apart from retrieving the expressions and applying to them
X = X.with_columns(**exprs)

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

tubular/imputers.py
```diff
@@ -164,6 +165,23 @@ def _generate_imputation_expressions(self, expr: nw.Expr, col: str) -> nw.Expr:
...
+        """Check if fit failed to find needed attrs (if impute_values_ are None).
+
+        Raises
+        ------
+            ValueError: if impute_values_ have come out as None
+
+        """
+        for col in self.columns:
+            failed_columns = []
+            if pd.isna(self.impute_values_[col]):
```
ChristinaKatsiafa:
```suggestion
            if self.impute_values_[col] is None or np.isnan(self.impute_values_[col]):
```
Similarly should we move away from pandas?

tubular/imputers.py
ChristinaKatsiafa, line 383: S: Again this is not in the scope of this PR but we dont we use selectors here?
import narwhals.selectors as nws

obj_columns  = X.select(nws.by_dtype(nw.Object)).columns
cat_columns  = X.select(nws.categorical()).columns
num_columns  = X.select(nws.numeric()).columns
bool_columns = X.select(nws.boolean()).columns
str_columns  = X.select(nws.string()).columns
null_columns = X.select(nws.by_dtype(nw.Unknown)).columns

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        y : Series or None, default = None
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        y : Series or None, default = None
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        y : Series or None, default = None
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        DataFrame:
```

tubular/nominal.py
ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        y : Series or None, default = None
```
This is mainly so we can be consistent throughout the whole package.

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

tubular/nominal.py
```diff
@@ -1189,6 +1192,34 @@ def _setup_fit_multi_level(
...
+
+        Raises
+        ------
+            ValueError: if mapping values or unseen_level_handling_dict values
+            have come out as None unexpectedly
+
+        """
+        for col in self.encoded_columns:
+            failed_columns = []
+            if any(pd.isna(value) for value in self.mappings[col].values()):
```
ChristinaKatsiafa: S: should we move away from pandas please?

tubular/nominal.py
ChristinaKatsiafa:
```suggestion
        X_y : DataFrame
```

ChristinaKatsiafa:
```suggestion
        y : Series
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        DataFrame
```
N: Keep consistent with the rest of the package

tubular/mixins.py
ChristinaKatsiafa:
```suggestion
        X: DataFrame
```

ChristinaKatsiafa:
```suggestion
            DataFrame:
```

ChristinaKatsiafa:
```suggestion
        DataFrame: Transformed dataframe
        
```
S: The format of the returns is not consistent. Can we please pick one and keep this consistent throughout
Maybe raise a PR to sort this out?

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
            DataFrame:
```

tubular/nominal.py
ChristinaKatsiafa: S: should we move away from pandas?

ChristinaKatsiafa: S: similarly here. Should we move away from pandas?

ChristinaKatsiafa: Q: is this still correct? As we use expressions now arent we?

ChristinaKatsiafa: S: Similarly here as well please. Remove pandas?

tubular/nominal.py
```diff
@@ -2424,6 +2485,8 @@ def fit(self, X: pd.DataFrame, y: pd.Series) -> pd.DataFrame:
         self.mappings_from_null = base_mapping_transformer.mappings_from_null
         self.return_dtypes = base_mapping_transformer.return_dtypes
 
+        self._check_for_failed_fit()
+
         return self
 
     def transform(self, X: pd.DataFrame) -> pd.DataFrame:
```
ChristinaKatsiafa:
```suggestion
    @beartype
    def transform(self, X: DataFrame) -> DataFrame:
```

tubular/nominal.py
ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa:
```suggestion
        X : DataFrame
```

ChristinaKatsiafa, line 2279: Q: Why are converting this? I cant see this being used anywhere. Does it?

ChristinaKatsiafa, line 523: Q: Is check_for_failed_fit not needed in all fit? As it looks like it is not aways used.

liamholmes31: Had to convert this one because it uses the weights mixin which now outputs a nw expression, so had to make it narwhals friendly in order to not break

(it is deprecated and we may remove in the next major release, but this was just to keep things moving/working for now)

tubular/mixins.py
liamholmes31: i've just replicated the functionality that was already there for now, so the real answer is 'legacy'

but also, is there a common enough scenario where negative weights would be valid? haven't encountered one myself

tubular/imputers.py
liamholmes31: this was our initial implementation, but we changed when we found the selectors were quite slow, as X.select triggers a materialisation

tubular/mixins.py
ChristinaKatsiafa: I think I have only seen negative weights once in a glm years ago. Tbh is it very rare but as tubular is an open source tool my questions why should we have this restriction as in theory it is a valid weight.

tubular/mixins.py
liamholmes31: we commit to numpy style in the readme, have set one up to check everything is standardised here https://github.com/azukds/tubular/issues/692

liamholmes31: the way it generally happens is that this method is hidden under the hood, and a user would provide weights to the transformer itself (which determines whether or not this method gets called, if weights_column=None)

can see an example below

https://github.com/azukds/tubular/blob/4822cc8f329179ef783aa61944cb4f71a7083ec0/tubular/imputers.py#L929

tubular/mixins.py
liamholmes31: currently, that would break the following one:
`TestCheckWeightsColumn::test_weight_col_non_numeric[pandas] - AttributeError: 'DataFrame' object has no attribute 'schema'`

So would require making the mixins polars only - we could do this, but think it should be a spinout ticket where we consider the benefit vs the cost of introducing a complication (this function in most cases will just do a type check and nothing else, so think the speed implications will be really minimal)

tubular/mixins.py
liamholmes31: hmm, a few thoughts on this:
- in most use cases I've worked on negative weights would indicate a data issue and should be filtered out
- so think we would want to allow users to specify whether they are valid/invalid, which introduces some complexity, which is a downside
- given that no users have actually requested support for negative weights, I think my inclination would be not to pre-empt and leave this out until we do actually get some feature requests for it

tubular/capping.py
liamholmes31: think we're agreed on trying this, and the new issue is here https://github.com/azukds/tubular/issues/691

tubular/nominal.py
liamholmes31: I've just added it to the ones that meet these conditions:
- use weights
- fitting on an empty df gives nonsensical results, e.g. impute_value is None

so for example GroupRareLevelsTransformer meets the first but not the second condition, as finding no rare levels is still a result that makes sense

liamholmes31: thanks, have updated to remove a lot of the extra text, felt like the one liner is enough https://github.com/azukds/tubular/pull/654/changes/3802b07c1a8fd8ee0c42402b1aa1c978e265a161

tubular/nominal.py
liamholmes31: updated here https://github.com/azukds/tubular/pull/654/changes/5ca574bba59ec9e77f1c2f1c3c979ff50b4ae7e9

and here https://github.com/azukds/tubular/pull/654/changes/82f377dcf87a6c4f11babddd981f5f314b5928ff

tubular/nominal.py
liamholmes31: updated here https://github.com/azukds/tubular/pull/654/changes/6d58abc9a3809a2ca9491cc8dae92cbdc7a65750

liamholmes31: updated here https://github.com/azukds/tubular/pull/654/changes/6d58abc9a3809a2ca9491cc8dae92cbdc7a65750

tubular/imputers.py
liamholmes31: updated here https://github.com/azukds/tubular/pull/654/changes/5c4c21323b03829e90a8d0104960aec2c09b00ec

tubular/mixins.py
liamholmes31: removed here https://github.com/azukds/tubular/pull/654/changes/3ea0222e99c77bbe099128376f37e8c445a01778

tubular/capping.py
liamholmes31: updated here https://github.com/azukds/tubular/pull/654/changes/6969eb3a17ee1b78c763b2e871edfcd6de12d4ff

liamholmes31:
it wouldn't let me accept all of the docstring suggestions for some reason, but have tidied up the last ones here

https://github.com/azukds/tubular/pull/654/changes/9b40814ce37d1be6c25af98f3d973d06e227f6e7
https://github.com/azukds/tubular/pull/654/changes/e03e8d89fb01600455e43cd3e5b56e33ea8e9031
https://github.com/azukds/tubular/pull/654/changes/fa3474c51949f7ca068f96cd3ddd7b1df23d3dbd

tubular/mixins.py
```diff
@@ -241,50 +237,38 @@ def check_weights_column(self, X: FrameT, weights_column: str) -> None:
...
-        expr_null = nw.col(weights_column).is_null().sum().alias("null_count")
-        expr_nan = nw.col(weights_column).is_nan().sum().alias("nan_count")
-        expr_finite = (nw.col(weights_column).is_finite()).all().alias("all_finite")
-        expr_sum = nw.col(weights_column).sum().alias("sum")
-
-        checks = X.select(expr_min, expr_null, expr_nan, expr_finite, expr_sum)
-        min_val, null_count, nan_count, all_finite, sum_val = checks.row(0)
+    @staticmethod
+    @beartype
+    def get_valid_weights_filter_expr(weights_column: str) -> nw.Expr:
```
ChristinaKatsiafa: Q: As this is a new behaviour can we please add a warning that if there are weights that are negative, nan, None and infinite will be filtered out? and if this is not what they want to happen then they should provide weights that are strictly positive numbers.

liamholmes31: added that here https://github.com/azukds/tubular/pull/654/changes/07459ff6a90e3d6578244ea2f5d6cd408c820799

I added a verbose arg to control it, as lots of warnings can get quite annoying!

ChristinaKatsiafa: N: Can you please add a small test for this one as well?

liamholmes31: added here :) https://github.com/azukds/tubular/pull/654/changes/c16c7b737e13f93446d2dd1062bb65fb65b8d955

ChristinaKatsiafa (approved):
I am happy with the changes

PR #662, opened 2025-12-23
Add to_json() method for BaseAggregationTransformer

shameem2k4:


tests/aggregations/test_BaseAggregationTransformer.py
```diff
@@ -85,7 +92,7 @@ def test_error_for_bad_col_types(
 
         if _check_if_skip_test(transformer, test_df, lazy):
```
liamholmes31: this will skip the test for the other unconverted agg transformer
```suggestion
        if _check_if_skip_test(transformer, test_df, lazy=lazy, from_json=from_json):
```

PR #671, opened 2026-01-09
Feature/nominal doclints

liamholmes31:
cleaning up the last of the doclint errors

will close


tubular/nominal.py
```diff
@@ -626,7 +642,7 @@ def fit(
 
     @beartype
     def transform(self, X: DataFrame) -> DataFrame:
-        """Grouped rare levels together into a new 'rare' level.
+        """Group rare levels together into a new 'rare' level.
```
Boluwatife28, line 645: can we include a Raises section in the docstring for this transform method? it seems to raise a ValueError when nulls are detected

liamholmes31: I do agree this would  be sensible, but this ruff rule doesn't like it https://docs.astral.sh/ruff/rules/docstring-extraneous-exception/

and it doesn't seem to want to be  turned off! happy to leave as is and can revisit that ruff rule in future?

Boluwatife28: Noted, that makes sense. Happy to approve as is.

Boluwatife28 (dismissed):

JohnStokes228 (dismissed):
test

Boluwatife28 (approved):

PR #683, opened 2026-02-02
Feature/align output medianimputer

shameem2k4:
Below error is thrown for .over() method that is called at the end to broadcast the scalar value sum back to the original length.

NotImplementedError: Only elementary expressions are supported for `.over` in pandas-like backends when `partition_by` is specified.

### DIFF
--- a/tubular/imputers.py
+++ b/tubular/imputers.py
@@ -89,8 +89,8 @@ class BaseImputer(BaseTransformer):
 
         Raises
         ------
-            RuntimeError:
-                if class is not jsonable
+        RuntimeError
+            if class is not jsonable
 
         Examples
         --------
@@ -351,7 +351,8 @@ class ArbitraryImputer(BaseImputer):
 
         Raises
         ------
-            TypeError: if given impute value clashes with types of given columns
+        TypeError
+            if given impute value clashes with types of given columns
 
         Returns
         -------
@@ -482,10 +483,6 @@ class ArbitraryImputer(BaseImputer):
         X : DataFrame
             Data containing columns to impute.
 
-        return_native_override: Optional[bool]
-            option to override return_native attr in transformer, useful when calling parent
-            methods
-
         Returns
         -------
         X : DataFrame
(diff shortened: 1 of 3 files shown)


tubular/imputers.py
```diff
@@ -728,39 +725,41 @@ def fit(self, X: DataFrame, y: Optional[Series] = None) -> MedianImputer:
         # as median depends on data ordering, it is less amenable to writing in
         # pure expression form, so implementation here is still
```
liamholmes31, line 726: getting a little bit complicated here, so just had a play with it on my end, and the below works with simpler changes:
```
@block_from_json
    @beartype
    def fit(self, X: DataFrame, y: Optional[Series] = None) -> MedianImputer:
        """Calculate median values to impute with from X.

        Parameters
        ----------
        X : pd/pl.DataFrame
            Data to "learn" the median values from.

        y : None or pd/pl.Series, default = None
            Not required.

        Returns
        -------
            MedianImputer:
                fitted class instance.

        Examples
        --------
        ```pycon
        >>> import polars as pl
        >>> test_df = pl.DataFrame({"a": [1, None, 2], "b": [3, None, 4]})
        >>> imputer = MedianImputer(columns=["a", "b"])
        >>> imputer = imputer.fit(test_df)
        >>> imputer.transform(test_df)
        shape: (3, 2)
        ┌─────┬─────┐
        │ a   ┆ b   │
        │ --- ┆ --- │
        │ f64 ┆ f64 │
        ╞═════╪═════╡
        │ 1.0 ┆ 3.0 │
        │ 1.5 ┆ 3.5 │
        │ 2.0 ┆ 4.0 │
        └─────┴─────┘

        ```

        """
        y = _convert_series_to_narwhals(y)

        super().fit(X, y)

        self.impute_values_ = {}

        all_null_cols = _get_all_null_columns(X, self.columns)

        if all_null_cols:
            # touch the dict entry for each all null col so that they are recorded
            self.impute_values_.update(
                dict.fromkeys(all_null_cols),
            )

            warnings.warn(
                f"{self.classname()}: The Median of columns {all_null_cols} will be None",
                stacklevel=2,
            )

        not_all_null_columns = sorted(set(self.columns).difference(set(all_null_cols)))

        weights_column = self.weights_column
        if self.weights_column is None:
            X, weights_column = WeightColumnMixin._create_unit_weights_column(
                X,
                backend=nw.get_native_namespace(X).__name__,
                return_native=False,
            )

        WeightColumnMixin.check_weights_column(self, X, weights_column)

        # as median depends on data ordering, it is less amenable to writing in
        # pure expression form, so implementation here is still
        # slightly pandas-like
        # also, the weighted median approach is genuinely different to the unweighted
        # approach, so have left as two separate logic flows
        for c in not_all_null_columns:
            col_not_null_expr = ~nw.col(c).is_null()

            X = X.sort(c)

            col_expr = nw.when(col_not_null_expr).then(nw.col(c)).otherwise(None)
            weight_expr = (
                nw.when(col_not_null_expr)
                .then(nw.col(weights_column))
                .otherwise(None)
            )

            median_expr = _get_median_calculation_expression(
                initial_column_expr=col_expr,
                initial_weights_expr=weight_expr,
            )

            # impute value is weighted median
            self.impute_values_[c] = X.select(median_expr).item(0, 0)

        return self
```

liamholmes31: we'll also need to make a similar change to this:
```
col_expr = nw.when(col_not_null_expr).then(nw.col(c)).otherwise(None)
            weight_expr = (
                nw.when(col_not_null_expr)
                .then(nw.col(weights_column))
                .otherwise(None)
            )

            median_expr = _get_median_calculation_expression(
                initial_column_expr=col_expr,
                initial_weights_expr=weight_expr,
            )
```

in nominal.MeanResponseTransformer and capping.BaseCappingTransformer, where _get_median_expression and _weighted_quantiles_expr are called

PR #687, opened 2026-02-06
Converted ColumnDtypeSetter to narwhals

liamholmes31:
Converted ColumnDtypeSetter to narwhals

https://github.com/azukds/tubular/issues/414

### DIFF
--- a/tubular/misc.py
+++ b/tubular/misc.py
@@ -189,6 +189,7 @@ SimpleCastDtypesStr = Annotated[
 ]
 
 
+@register
 class ColumnDtypeSetter(BaseTransformer):
     """Transformer to set transform columns in a dataframe to a dtype.
 


tubular/misc.py
```diff
@@ -165,13 +165,30 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+    CATEGORICAL = "Categorical"
+
+
+SimpleCastDtypesStr = Annotated[
+    str,
+    Is[lambda s: s in SimpleCastDtypes._value2member_map_],
+]
+
+
 class ColumnDtypeSetter(BaseTransformer):
```
Ritunjai-Sharma, line 193: Q: Do we need to add @register decorator to ColumnDtypeSetter class like SetValueTransformer?

tests/misc/test_ColumnDtypeSetter.py
```diff
@@ -0,0 +1,159 @@
...
+        assert_frame_equal_dispatch(
+            _collect_frame(df_transformed, lazy=lazy), expected_df
+        )
+
+        # Test single row transformation
+        df = nw.from_native(df)
+        expected_df = nw.from_native(expected_df)
+        for i in range(len(df)):
+            df_transformed_row = transformer.transform(
+                _convert_to_lazy(df[[i]].to_native(), lazy)
```
Ritunjai-Sharma, line 130: Q: The code here seems to be using `df[[i]]` to select the i-th row. Looks like the df here is a narwhals DataFrame, and double brackets with a list, i.e., `(df[[…]])` means “select these columns,” not “select this row,” in both pandas and polars (and narwhals follows this)?

liamholmes31: so the docs for getitem say:
- Integers are always interpreted as positions
- Strings are always interpreted as column names.

so this is just looking up a row index. Doing it like df[[i]] rather than df[i] means it returns a DataFrame rather than a Series, as the lookup reference is a list so could contain multiple indices

https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.__getitem__

tubular/misc.py
liamholmes31: ah yeah should have done, thanks https://github.com/azukds/tubular/pull/687/changes/bd1920e444d02a823440926e3bb750bf61513581

tests/misc/test_ColumnDtypeSetter.py
liamholmes31: so the docs for getitem say:

- Integers are always interpreted as positions
- Strings are always interpreted as column names.

so this is just looking up a row index. Doing it like df[[i]] rather than df[i] means it returns a DataFrame rather than a Series, as the lookup reference is a list so could contain multiple indices

https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.__getitem__

Ritunjai-Sharma (approved):

Issue #532, opened 2025-10-08
make OneHotEncoder lazyframe compatible


liamholmes31:
What?
The first issue is from the weights mixin, which is covered below:
https://github.com/azukds/tubular/issues/528

Other issues:
- fussy about nulls, suggest we could just turn this off
- transformer looks up present levels:
    - in fit, this is the primary functionality so can leave in place
    - in transform, we warn about requested levels that are missing and new levels that are present, suggest we just turn this off

Why?
allow lazyframe support

How?
see what

PR #688, opened 2026-02-09
Feature/lazy onehot

liamholmes31:
converted OHE to be lazy-friendly

https://github.com/azukds/tubular/issues/532

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,7 @@ Each individual change should have a link to the pull request after the descript
 Changed
 ^^^^^^^
 - feat: made OneHotEncoder lazyframe compatible `#532 <https://github.com/azukds/tubular/issues/532>`_
+- chore: small refactor of BaseCappingTransformer to remove prepare_quantiles method for readability
 
 2.8.0 (23/02/2026)
 ------------------
--- a/tubular/capping.py
+++ b/tubular/capping.py
@@ -231,8 +231,8 @@ class BaseCappingTransformer(BaseNumericTransformer, WeightColumnMixin):
             ValueError: if quantile_capping_values have fit as None where values were expected
 
         """
+        failed_columns = []
         for col in self.quantiles:
-            failed_columns = []
             for i, init_value in enumerate(self.quantiles[col]):
                 fit_value = self.quantile_capping_values[col][i]
                 if not _is_null(init_value) and _is_null(fit_value):
@@ -249,8 +249,8 @@ class BaseCappingTransformer(BaseNumericTransformer, WeightColumnMixin):
         """Learn capping values from input data X.
 
         Calculates the quantiles to cap at given the quantiles dictionary supplied
-        when initialising the transformer. Saves learnt values in the capping_values
-        attribute.
+        when initialising the transformer. Saves learnt values in the
+        quantile_capping_values and replacement_values attributes.
 
         Parameters
         ----------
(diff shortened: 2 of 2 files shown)


tubular/nominal.py
```diff
@@ -2013,105 +1966,41 @@ def fit(
...
-                if not self.wanted_values
-                else len(self.wanted_values[c])
-            )
+            # if the user has provided a 'wanted_values' as a list of expected dummies,
+            # then there is actually nothing we need to fit on data here
+            if not self.wanted_values:
+                self.categories_[c] = sorted(
+                    category
+                    for category in results_dict[c].unique().to_list()
+                    if not _is_null(category)
```
hariinduu: Q: What if we need to include 'None' as another category? Maybe we can include another parameter at the init stage and pass it here so we can decide whether to ignore, error or include nulls?

tubular/nominal.py
```diff
@@ -2013,105 +1966,41 @@ def fit(
...
+                )
+
+            else:
+                self.categories_[c] = self.wanted_values[c]
+
+            level_count = len(self.categories_[c])
 
             if level_count > self.MAX_LEVELS:
```
hariinduu, line 1999: S: Also we can set a warning if the level_count == 1 which means there is only 1 level in the column

tubular/nominal.py
```diff
@@ -2213,44 +2102,14 @@ def transform(
         X = _convert_dataframe_to_narwhals(X)
         X = BaseTransformer.transform(self, X, return_native_override=False)
 
-        # first handle checks
-        present_levels = {}
-        for c in self.columns:
-            # print warning for unseen levels
-            present_levels[c] = set(X.get_column(c).unique().to_list())
-            unseen_levels = set(present_levels[c]).difference(set(self.categories_[c]))
```
hariinduu, line 2221: Q: Are we handling unseen_levels in the new version?

tests/nominal/test_OneHotEncodingTransformer.py
hariinduu, line 27: Q: Are we testing columns with duplicate column names?

tubular/nominal.py
liamholmes31: I think this seems a bit'extra' to me and like something we can just leave to a user? or wait and see if we get requests for something like this before we action

tubular/nominal.py
liamholmes31: no we've removed this warning as it involves inspecting the data so would be tricky for lazyframes

also we're generally trying to move to doing less 'extra' stuff like this type of warning

tests/nominal/test_OneHotEncodingTransformer.py
liamholmes31: not currently, had a quick read into this:
- polars doesn't allow duplicate column names and would error at df init
- but pandas does allow this

think this needs a bit more thought, so have spun up another issue https://github.com/azukds/tubular/issues/698

tubular/nominal.py
liamholmes31: added a warning explaining that nulls get filtered here https://github.com/azukds/tubular/pull/688/changes/26f9143316fb8c3935ff4369ea2335c610d8cceb

I think generally you would not want to OHE nulls, so adding an argument to control feels like it wouldn't pass a complexity cost/benefit

tubular/nominal.py
hariinduu: Yes sounds good, Ideally we will see those non value adding columns in the EDA and shouldn't be passing them

PR #689, opened 2026-02-11
refactored BaseCappingTransformer to remove prepare_quantiles method,…

liamholmes31:
… for readability

### DIFF
--- a/tubular/capping.py
+++ b/tubular/capping.py
@@ -249,8 +249,8 @@ class BaseCappingTransformer(BaseNumericTransformer, WeightColumnMixin):
         """Learn capping values from input data X.
 
         Calculates the quantiles to cap at given the quantiles dictionary supplied
-        when initialising the transformer. Saves learnt values in the capping_values
-        attribute.
+        when initialising the transformer. Saves learnt values in the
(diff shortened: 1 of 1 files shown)


tubular/capping.py
Ritunjai-Sharma: Q: Not related to work done in this PR but looks like `_check_for_failed_fit` resets failed_columns inside the per-column loop, so failures in earlier columns would be lost and only the last column would effectively be considered? I think  initializing failed_columns before the loop over columns should fix the issue?

Ritunjai-Sharma: Q: The fit docstring says “Saves learnt values in the capping_values attribute,” but the implementation saves them in quantile_capping_values (and sets _replacement_values from that). Worth updating docstring?

liamholmes31: good catch! updated here https://github.com/azukds/tubular/pull/689/changes/8f1d7e4e18a86142136c6c9fc8d369cd2219dd6c

liamholmes31: updated here :) https://github.com/azukds/tubular/pull/689/changes/e2c8c0568d8c4c9937ae1cee87c5be6a4c38e6a8

PR #694, opened 2026-02-23
Feature/lazy grouprare

liamholmes31:
Converted transformer to work with lazyframes:
- needed a couple of tweaks to the unit weights handling to be lazy friendly
- updated null handling, to just leave nulls alone and not error/warn
- changed logic to use groupby/agg approach, as fitting on large data will generally benefit from the size reduction this gives

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
     -   id: blacken-docs
         args: [--skip-errors]
         additional_dependencies:
-        - black==22.12.0
+        - black==26.3.1
 -   repo: local
     hooks:
       - id: generate-feature-table
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,6 +34,8 @@ Changed
 ^^^^^^^
 - feat: made GroupRareLevelsTransformer lazyframe compatible `#531 <https://github.com/azukds/tubular/issues/531>`_
 - feat: edited dummy unit weight handling to be more lazy friendly
+- chore: updated pre commit black version
+- chore: updated pygments version on dependabot recommendation
 
 3.1.0 (20/03/2026)
 ------------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,7 +52,8 @@ dev = [
     "codespell",
     "typos",
     "blacken-docs",
-    "black >= 26.3.1"
+    "black >= 26.3.1",
+    "pygments >= 2.20.0",
     ]
 
 [project.urls]
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -51,8 +51,10 @@ prek==0.3.6
     # via tubular (pyproject.toml)
 pyarrow==23.0.1
     # via tubular (pyproject.toml)
-pygments==2.19.2
-    # via pytest
+pygments==2.20.0
+    # via
+    #   tubular (pyproject.toml)
+    #   pytest
 pytest==9.0.2
     # via
     #   tubular (pyproject.toml)


tubular/nominal.py
```diff
@@ -707,50 +635,49 @@ def transform(self, X: DataFrame) -> DataFrame:
...
 
-        # sort once nulls removed
-        present_levels = {c: sorted(present_levels[c]) for c in self.columns}
+        null_filter_exprs = {}
 
-        if not self.unseen_levels_to_rare:
-            for c in self.columns:
-                unseen_vals = set(present_levels[c]).difference(
-                    set(self.training_data_levels[c]),
+        for col in self.columns:
```
liamholmes31, line 638: have refactored this a bit to avoid repeated loops in multiple dict comprehensions

tubular/nominal.py
```diff
@@ -591,43 +534,38 @@ def fit(
...
 
-        total_weight_expr = nw.col(weights_column).sum()
+        if not self.unseen_levels_to_rare:
+            self.training_data_levels = {}
 
-        level_weight_perc_exprs = {
-            c: level_weights_exprs[c] / total_weight_expr for c in self.columns
-        }
+        for c in self.columns:
+            group = X.group_by(c).agg(nw.col(weights_column).sum())
```
liamholmes31, line 547: have edited this to go back to using grouped logic, I think this is generally going to be better with fit as will be working on large data, and reducing the size by grouping will be beneficial

tubular/nominal.py
```diff
@@ -462,63 +461,6 @@ def _check_str_like_columns(self, schema: nw.Schema) -> None:
             msg = f"{self.classname()}: transformer must run on str-like columns, but got non str-like {non_str_like_columns}"
             raise TypeError(msg)
 
-    @beartype
-    def _check_for_nulls(self, present_levels: dict[str, list[Any]]) -> None:
```
danielarifmurphy, line 466: Q: pattern here being to remove all stateful checks and allow native errors to arise? im a fan! makes codebase lighter too

tubular/nominal.py
```diff
@@ -572,14 +514,12 @@ def fit(
 
         super().fit(X, y)
 
-        backend = nw.get_native_namespace(X)
```
danielarifmurphy, line 575: Q: my understanding is `backend` was needed before due to `pandas` behaving differently, is this no longer the case?

tubular/nominal.py
```diff
@@ -592,43 +532,38 @@ def fit(
...
 
-        present_levels = {c: list(set(X.get_column(c).unique())) for c in self.columns}
-
-        self._check_for_nulls(present_levels)
-
-        # sort once nulls are removed
-        present_levels = {c: sorted(present_levels[c]) for c in self.columns}
-
         self.non_rare_levels = {}
+        self.rare_levels_record_ = {}
```
danielarifmurphy: Q/C: when do we decide to use trailing underscores for attrs?

tubular/nominal.py
```diff
@@ -592,43 +532,38 @@ def fit(
...
-        }
+            non_rare_levels_expr = (
+                nw.when(level_weights_expr >= self.cut_off_percent)
+                .then(nw.col(c))
+                .otherwise(None)
+                .alias(f"{c}_non_rare_levels")
+            )
 
-        results_dict = X.select(**non_rare_levels_exprs).to_dict(as_series=True)
+            results = group.select(non_rare_levels_expr, nw.col(c))
```
danielarifmurphy, line 556: P: nice! find this much easier to read/reason about than slinging dicts around throughout

tubular/nominal.py
```diff
@@ -688,14 +621,6 @@ def transform(self, X: DataFrame) -> DataFrame:
         │ rare_level ┆ z   │
         └────────────┴─────┘
 
-        >>> # erroring example (with nulls)
```
danielarifmurphy, line 691: S: do we reckon there's value in keeping this doctest example in but showing the native error rather than a custom one?

tubular/nominal.py
```diff
@@ -592,43 +532,38 @@ def fit(
...
 
-        total_weight_expr = nw.col(weights_column).sum()
+        if not self.unseen_levels_to_rare:
+            self.training_data_levels = {}
 
-        level_weight_perc_exprs = {
-            c: level_weights_exprs[c] / total_weight_expr for c in self.columns
-        }
+        for c in self.columns:
+            group = X.group_by(c).agg(nw.col(weights_column).sum())
```
danielarifmurphy, line 547: C: do we have strong feelings around indicating if a var is a design matrix e.g. using `X_group`? i dont think it's needed as it's clear youre performing df ops on it but just wondering

tubular/nominal.py
```diff
@@ -708,52 +633,47 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+                    nw.col(col).is_in(self.non_rare_levels[col])
+                    | ~nw.col(col).is_in(self.training_data_levels[col])
                 )
-                non_rare_levels[c].extend(unseen_vals)
+            )
 
-        transform_expressions = {
-            c: nw.col(c).cast(
-                nw.String,
+            transform_expressions[col] = (
```
danielarifmurphy: Q: is there a reason these are stored in dicts rather than lists? imagine there being some (v minor) overhead

originally thought maybe this gets sticky if we wanted to add more exprs for a column with an existing key, but guess i can't really see a case where you'd have multiple sets of exprs for a given column for a given transformer

tubular/mixins.py
```diff
@@ -158,61 +157,50 @@ def _create_unit_weights_column(
...
-            ),
-        )
+            if X.schema[unit_weights_column] not in NumericTypes:
+                error_msg = f"{unit_weights_column} is present in X and non-numeric, transformer logic requires this to be an all 1 value column."
+                raise TypeError(
+                    error_msg,
+                )
+
+            if verbose:
+                # error if column already exists but is not suitable
```
danielarifmurphy: S: can update this comment as error no longer raised

tubular/mixins.py
```diff
@@ -158,61 +157,50 @@ def _create_unit_weights_column(
...
+                warn_msg = f"column {unit_weights_column} is present in X, transformer logic will assume this column contains all 1 values."
```
danielarifmurphy, line 197: Q: would we ever want to overwrite the existing unit weight col with 1s on the user's behalf?

only so defensive we can be...

tubular/nominal.py
liamholmes31: yeah exactly :)

tubular/nominal.py
liamholmes31: it looks like backend was only fed to _create_unit_weights_column which previously created the new column like nw.new_series(..., backend=backend), but now uses nw.lit(1) instead which simplifies things

tubular/nominal.py
liamholmes31: this wouldn't actually error now, the behaviour would just be None in -> None out

tubular/mixins.py
liamholmes31: yeah it feels like that would be too much, because it locks us in to checking/overwriting (or just overwriting) with every .fit call - so think I'm happy to just assume here (the column is unlikely to exist often at all going in anyway)

tubular/nominal.py
liamholmes31: I can't see a good reason why this one used _, so have removed here https://github.com/azukds/tubular/pull/694/changes/d285879e9d404aabe103ab8c66f398ec69b1a0ac

think generally trailing _ is just to avoid nameclashes, e.g. type_

tubular/nominal.py
liamholmes31: was able to switch this fairly easily so have done here https://github.com/azukds/tubular/pull/694/changes/eacbfdc0a4e6742d34a01c1417b3beb2f6c687c8

tubular/mixins.py
liamholmes31: thanks removed here https://github.com/azukds/tubular/pull/694/changes/35d49ff4df37450a7f975ccc28d7d9831e161765

tubular/mixins.py
danielarifmurphy: yeah agree with this

tubular/nominal.py
danielarifmurphy: of course!

tubular/nominal.py
danielarifmurphy: changed my mind on this: think you'd only want the `X_` prefix if its something you would reasonably pass to a model - a groupby op is a bit different

PR #712, opened 2026-04-10
Feature/remove droporiginal

liamholmes31:
removed droporiginal logic to simplify repo, dropping columns can easily be left to users

to avoid a breaking change, BaseTransformer still accepts this argument and will just warn that it is unused (when verbose is on)

https://github.com/azukds/tubular/issues/557


tubular/base.py
```diff
@@ -186,12 +187,23 @@ def __init__(
...
+            deprecated argument, has no effect.
+
         """
         self.verbose = verbose
+        self.drop_original = drop_original
 
         if self.verbose:
             print("BaseTransformer.__init__() called")
 
+            if self.drop_original:
```
ozzyholemans: Q: are we happy with no warning being raised if verbose mode not on? I'm guessing most usage is with `verbose=False`, so users won't by default see that the `drop_original` argument is no longer doing anything

tubular/base.py
```diff
@@ -186,12 +187,23 @@ def __init__(
         return_native: bool, default = True
             Controls whether transformer returns narwhals or native pandas/polars type
 
+        drop_original: Optional[bool], default = None
```
ozzyholemans: N: as per numpy style guide, optional parameters should be formatted like: `drop_original : bool, optional`

tests/base_tests.py
```diff
@@ -136,29 +136,6 @@ def test_columns_non_string_or_list_error(
             uninitialized_transformers[self.transformer_name](**args)
 
 
-class DropOriginalInitMixinTests:
```
ozzyholemans, line 139: Q: we are making `drop_original` an accepted (but deprecated) parameter for `BaseTransformer` to ensure the change is not breaking. Do we need to retain a test to confirm that no error is thrown if a user supplies `drop_original`?

tests/nominal/test_OneHotEncodingTransformer.py
ozzyholemans: S: remove the `drop_original` arg here?

tests/nominal/test_MeanResponseTransformer.py
```diff
@@ -1250,18 +1250,21 @@ def test_expected_outputs(
...
                 nw.col(col).cast(getattr(nw, x.return_type)),
             )
-        expected = nw.to_native(expected)
 
         column_order = nw.from_native(df_transformed).columns
 
+        df = nw.from_native(df)
+        for col in columns:
+            if col not in expected:
+                expected = expected.with_columns(df[col] for col in columns)
```
ozzyholemans, line 1259: Q: what's the purpose of this and why wasn't it needed before?

liamholmes31: this is just reintroducing the original columns to the expected_df (that were previously dropped by drop_original), added a comment here https://github.com/azukds/tubular/pull/712/changes/c184cf8c17ab9f21ef3c473a13a815049b6e7a5b

tests/nominal/test_OneHotEncodingTransformer.py
liamholmes31: removed here https://github.com/azukds/tubular/pull/712/changes/69daae72408dd76d8755e161e3fc874fa55f1ca5

tests/base_tests.py
liamholmes31: good shout, added here https://github.com/azukds/tubular/pull/712/changes/833629529be47a42e49e66a388363cdc9d0506cc

tubular/base.py
liamholmes31: updated to match style guide here https://github.com/azukds/tubular/pull/712/changes/0f4d0cf68539dea1d08d703f32a2649a0cb25277

tubular/base.py
liamholmes31: very good point, I've now moved outside so that users can see this warning and remove the arg from their pipelines https://github.com/azukds/tubular/pull/712/changes/5493155d1b6eec9f6b1e1a707876803d402fa1cc

tests/dates/test_DateDiffLeapYearTransformer.py
ozzyholemans, line 106: F: Doing a quick ctrl-f across the repo finds a duplicate function to this in `tests/test_data.py` that doesn't appear to be used. Do we want to remove it?

ozzyholemans (approved):

tests/dates/test_DateDiffLeapYearTransformer.py
liamholmes31: is that meaning `expected_date_diff_df_2`? happy to do this in a quick followup PR, but couldn't spot the dupe on my end

PR #715, opened 2026-04-16
Added check is fitted for sklearn compatibility

### LINKED ISSUE #494: Upcoming changes to sklearn fitted status checking and underscore attributes
Saw this warning on a pipeline using non tubular transformers:

`FutureWarning: This Pipeline instance is not fitted yet. Call 'fit' with appropriate arguments before using other methods such as transform, predict, etc. This will raise an error in 1.8 instead of the current warning.`

I have not yet checked in tubular for the warning but I assume the same issue will pop up here.

Sklearn is starting to check if a pipeline is fitted or not with [check_is_fitted](https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html) which checks for attributes with a trailing underscore for the __sklearn_is_fitted__ attribute. 

To address this we should set __sklearn_is_fitted__ to True in BaseTransformer.fit().

ChristinaKatsiafa:
Added check is fitted for sklearn compatibility and also added python 3.14 to the automated tests.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: "ubuntu-22.04"
     strategy:
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
 
     steps:
 
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -20,6 +20,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        sklearn-version: ["1.7", "1.8"]
 
     steps:
 
@@ -36,6 +37,7 @@ jobs:
           uv venv -p  ${{ matrix.python-version }} ${{ matrix.python-version }}
           source ${{ matrix.python-version }}/bin/activate
           uv pip install .[dev]
+          uv pip install "scikit-learn==${{ matrix.sklearn-version }}.*"
 
       - name: Test with pytest
         if: always()
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -21,6 +21,9 @@ jobs:
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
         sklearn-version: ["1.7", "1.8"]
+        exclude:
+          - python-version: "3.10"
+            sklearn-version: "1.8"
 
     steps:
 
--- a/tests/base_tests.py
+++ b/tests/base_tests.py
@@ -1612,9 +1612,7 @@ class OtherBaseBehaviourTestsNumeric:
         del transformer.is_fitted_
 
         # Now transform should raise NotFittedError
-        with pytest.raises(
-            NotFittedError, match=re.escape("Pipeline is not fitted yet.")
-        ):
+        with pytest.raises(NotFittedError):
             pipeline.transform(df)
 
 
@@ -1643,9 +1641,7 @@ class OtherBaseBehaviourTestsString:
         del transformer.is_fitted_
 
         # Now transform should raise NotFittedError
-        with pytest.raises(
-            NotFittedError, match=re.escape("Pipeline is not fitted yet.")
-        ):
+        with pytest.raises(NotFittedError):
             pipeline.transform(df)
 
 
@@ -1674,9 +1670,7 @@ class OtherBaseBehaviourTestsDatetime:
         del transformer.is_fitted_
 
         # Now transform should raise NotFittedError
-        with pytest.raises(
-            NotFittedError, match=re.escape("Pipeline is not fitted yet.")
-        ):
+        with pytest.raises(NotFittedError):
             pipeline.transform(df)
 
 
@@ -1705,7 +1699,5 @@ class OtherBaseBehaviourTestsDates:
         del transformer.is_fitted_
 
         # Now transform should raise NotFittedError
-        with pytest.raises(
-            NotFittedError, match=re.escape("Pipeline is not fitted yet.")
-        ):
+        with pytest.raises(NotFittedError):
             pipeline.transform(df)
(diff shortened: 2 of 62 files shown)


ChristinaKatsiafa:
Replaced the .Schema with collect_schema() and .columns with collect_schema().names() as this is most efficient to use with lazyframes

tubular/capping.py
```diff
@@ -781,10 +782,11 @@ def fit(self, X: DataFrame, y: Series | None = None) -> CappingTransformer:
         ```
 
         """
+        self.is_fitted_ = False  # will be set to True at end of fit if no errors raised
```
liamholmes31, line 785: Q: is there a reason not to set this in init?

liamholmes31 (commented):
just one main comment - looks like we sometimes set `self.is_fitted_ = False` in a few different ways (class attr, init, start of fit), may be best to just pick one and standardise?

adamsardar:
Is there a ticket for this feature?

Piecing together from comments, the changes are needed for the latest sklearn versions (which? >=1.8.0?) otherwise an error will be raised.

The project `requirements-dev.txt`, and therefore CI, is still on [scikit-learn==1.7.2](https://github.com/azukds/tubular/blob/main/requirements-dev.txt#L78), so we don't actually test if the project runs with the newer scikit-learn version. Could we expand the CI build matrix to test both options? sklearn has been a repeat offender for breaking tubular functionality over the years (e.g. #308), so it feels prudent to test around package versions that we know can raise errors.

ChristinaKatsiafa:
Hello Adam,
Yes there is a ticket and it is liked to this PR as well. https://github.com/azukds/tubular/issues/494
This is something that is required in sklearn version 1.8.
Good point regarding the testing. I didnt notice that we have a pin for the sklearn package so I will expand the CI build matrix so this is properly tested.
Thanks

tubular/capping.py
ChristinaKatsiafa: Yes. So the idea here is that the is_fitted will be set to true if the transformer does not fit or when the fit function is fully run without errors. This is why I set it to False and then I set it to true at the end of the fit function.
If this is not something that we would like to have then this can very easily be set to true and the base transformer and be done with the whole process but I thought that it might worth going the long way.
What do you think?

ChristinaKatsiafa:
Checking the CI and it looks like CI is set to use the dev dependencies from the pyproject.toml instead of the requirements-dev.txt
image
I can also confirm that python 3.14 the sklearn version used is 1.8.0
image
It looks like that the fact that we didn't test for the sklearn pipeline compatibility with our transfomers was the reason that we didn't pick this up earlier. So I am currently adding extra tests to all transformers to make sure that this will thoroughly be tested.
Finally please let me know if this is enough for you or if you still want to add the sklearn version in the CI build matrix.

ChristinaKatsiafa:
All changes and extra tests are now done. What has been changed:
1. Added is_fitted_ to all transformers. This should be set to True when the transformer does not fit or at the end of the fit function to make sure that the fit process was successful.
2. Added check_is_fitted("is_fitted_") at the transform function. This will check if the is_fitted_ function exists and also this is now required for the sklearn version>=1.8. If those are not added then the sklearn pipeline will return an NotFitError in all versions >=1.8
3. Added tests to check that the transform will run without errors and also if the is_fitted_ is not set in the transformers then the pipeline will raise a NotFitError.
4. Some lz optimisation is also added by replacing .columns with collect_schema().names() and .Schema with collect_schema()

tubular/base.py
```diff
@@ -363,7 +364,7 @@ def fit(self, X: DataFrame, y: Series | None = None) -> BaseTransformer:
         y = _convert_series_to_narwhals(y)
 
         self.columns_check(X)
-
+        self.is_fitted_ = True
```
danielarifmurphy: Q: do we want this here at the end of `Base.fit`? this means for the capping transformer for example we have setting of `is_fitted_` as follows:

1. `False` (base init) --> 2. `True` (base fit) --> 3. `False` (capping fit start) --> 4. `True` (capping fit finish)

I think we can skip 2. and 3. here and just have:
- stateless transforms: set to `True` in `init`
- stateful transforms: set to `True` at the end of `fit` i.e. 4.

what do you reckon?

.github/workflows/python-package.yml
danielarifmurphy, line 42: S: if it's particular `sklearn` versions we wish to test against, we could add something as simple as:

```
matrix:
  python-version: ["3.10", "3.11", "3.12", "3.13"]
  sklearn-version: ["1.7", "1.8"] 
```
above, and then here:

```
uv pip install .[dev]
uv pip install "scikit-learn==${{ matrix.sklearn-version }}.*"
```
?

tests/base_tests.py
```diff
@@ -1582,3 +1585,119 @@ def test_is_serialisable(self, initialized_transformers, tmp_path):
...
+
+        # Transform should work
+        result = pipeline.transform(df)
+        assert result is not None
+
+        # Delete is_fitted_ from the transformer
+        del transformer.is_fitted_
+
+        # Now transform should raise NotFittedError
+        with pytest.raises(NotFittedError):
```
danielarifmurphy, line 1615: S: alternatively, we could have an `sklearn>=1.8` conditional here?

ChristinaKatsiafa: The new structure of adding is_fitted in all transformers and then check for this means that this will always return an error in all sklearn versions. So this is not needed.

.github/workflows/python-package.yml
ChristinaKatsiafa: Added [here](https://github.com/azukds/tubular/pull/715/commits/79ded3969a0b1af670612cc9302f98662ae2d368)

ChristinaKatsiafa: Also removed the combo of p3.10 and sklearn 1.8 as sklearn 1.8 works for python >=3.11
[here](https://github.com/azukds/tubular/pull/715/commits/be3e6d84aa10b719b34395611f5feed50a8ad20f)

tubular/base.py
ChristinaKatsiafa: That is a good point. Changed are added [here](https://github.com/azukds/tubular/pull/715/commits/0defc417b4f6f5ce8a2a255af869d10f2acea23d)
Thanks Dan!

danielarifmurphy (dismissed):

danielarifmurphy (approved):

PR #717, opened 2026-04-24
updated pytest

liamholmes31:
following dependabot recommendation

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -55,7 +55,7 @@ pygments==2.20.0
     # via
     #   tubular (pyproject.toml)
     #   pytest
-pytest==9.0.2
+pytest==9.0.3
     # via
     #   tubular (pyproject.toml)
     #   pytest-cov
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -55,7 +55,7 @@ pygments==2.20.0
     # via
     #   tubular (pyproject.toml)
     #   pytest
-pytest==9.0.2
+pytest==9.0.3
```
Ritunjai-Sharma, line 58: Q: Isn't this something that should be done in pyproject.toml and then run uv pip compile to update requirements-dev.txt?

liamholmes31: was worried about making the pin too restrictive, so have now done it with != here https://github.com/azukds/tubular/pull/717/changes/c4f5f75c76c36040177964ad7d4b6984b9aeff90

Ritunjai-Sharma (approved):

PR #721, opened 2026-04-27
deprecated OneDKMeansTransformer

liamholmes31:
deprecated transformer and moved to bottom of file


tubular/numeric.py
```diff
@@ -1858,3 +1535,346 @@ def transform(self, X: pd.DataFrame) -> pd.DataFrame:
             X[self.feature_names_out] = self.pca.transform(X[self.columns])
 
         return X
+
+
+@deprecated(
+    "This Transformer is deprecated, use DateDifferenceTransformer instead. "
```
lsumption: N: This needs updating

tubular/numeric.py
lsumption: I know this PR isn't looking at this transformer - so absolutely feel free to reject this for now, but this is a bit confusing and should be updated (at some point!)

PR #722, opened 2026-04-29
Feature/string contains

liamholmes31:
setup transformer which will search given columns for a given reference value (or column values), and flag the output with a boolean column

### DIFF
--- a/tests/strings/test_StringContainsTransformer.py
+++ b/tests/strings/test_StringContainsTransformer.py
@@ -1,0 +1,265 @@
+import narwhals as nw
+import pytest
+from beartype.roar import BeartypeCallHintParamViolation
+
+from tests.base_tests import (
+    ColumnStrListInitTests,
+    GenericTransformTests,
+    OtherBaseBehaviourTests,
+)
+from tests.utils import (
+    _check_if_skip_test,
+    _collect_frame,
+    _convert_to_lazy,
+    _handle_from_json,
+    assert_frame_equal_dispatch,
+    dataframe_init_dispatch,
+)
+from tubular.strings import StringContainsTransformer
+
+
+class TestStringContainsTransformerInit(
+    ColumnStrListInitTests,
+):
+    """Tests for StringContainsTransformer initialization."""
+
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "StringContainsTransformer"
+
+    @pytest.mark.parametrize("bad_reference", [[1, 2], 1, True])
+    def test_invalid_reference_error(
+        self, minimal_attribute_dict, uninitialized_transformers, bad_reference
+    ):
+        """Test that an error is raised for invalid characters arg."""
+
+        args = minimal_attribute_dict[self.transformer_name]
+        args["reference"] = bad_reference
+        with pytest.raises(BeartypeCallHintParamViolation):
+            uninitialized_transformers[self.transformer_name](**args)
+
+    @pytest.mark.parametrize("bad_reference_as_column", [[1, 2], 1, "a"])
+    def test_invalid_reference_as_column_error(
+        self,
+        minimal_attribute_dict,
+        uninitialized_transformers,
+        bad_reference_as_column,
+    ):
+        """Test that an error is raised for invalid characters arg."""
+
+        args = minimal_attribute_dict[self.transformer_name]
+        args["reference_as_column"] = bad_reference_as_column
+        with pytest.raises(BeartypeCallHintParamViolation):
+            uninitialized_transformers[self.transformer_name](**args)
+
+
+class TestStringContainsTransformerTransform(GenericTransformTests):
+    "tests for StringContainsTransformer.transform"
+
+    @classmethod
+    def setup_class(cls):
+        cls.transformer_name = "StringContainsTransformer"
+
+    @pytest.mark.parametrize("lazy", [True, False])
+    @pytest.mark.parametrize("library", ["polars"])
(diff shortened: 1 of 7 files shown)


tests/strings/test_StringContainsTransformer.py
```diff
@@ -0,0 +1,265 @@
...
+        if _check_if_skip_test(transformer, df, lazy=lazy, from_json=from_json):
+            return
+
+        transformer = _handle_from_json(transformer, from_json=from_json)
+
+        msg = "StringContainsTransformer: reference_as_column=True is only supported for polars backend"
+        with pytest.raises(TypeError, match=msg):
+            transformer.transform(_convert_to_lazy(df, lazy=lazy))
+
+    @pytest.mark.parametrize("lazy", [True, False])
```
ChloeCrisp, line 135: I don't think Pandas supports lazy frames so not sure if this is required here

tests/strings/test_StringContainsTransformer.py
```diff
@@ -0,0 +1,265 @@
...
+    @pytest.mark.parametrize(
+        ("input_values", "reference_value", "expected_output"),
+        [
+            (["a"], "a", [True]),
+            ([None], "a", [None]),
+            (["a"], "b", [False]),
+            (["a", "b", None], "b", [False, True, None]),
+        ],
+    )
+    def test_output_cases_as_value(
```
ChloeCrisp, line 147: Should there be a test for this where we check for when reference_value is None?

tubular/strings.py
```diff
@@ -1,583 +1,816 @@
...
+            characters to remove from specified columns.
+
+        **kwargs
+            Arbitrary keyword arguments passed onto BaseTransformer.init method.
+
+        """
+        super().__init__(columns=columns, **kwargs)
+
+        self.characters = characters
+        self.characters_formatted = r"[{}]".format("".join(self.characters))
```
ChloeCrisp, line 247: Does this accept raw regex fragment? If not should there be a separate regex parameter?

tubular/strings.py
```diff
@@ -1,583 +1,816 @@
...
+    """
+
+    polars_compatible = False
+
+    lazyframe_compatible = False
+
+    jsonable = False
+
+    deprecated = True
+
```
ChloeCrisp, line 395: This does not include `FITS = False` where other functions do include this. Should this be included here?

tubular/strings.py
```diff
@@ -1,583 +1,816 @@
...
+        Returns
+        -------
+        list[str]:
+            list of features modified/created by the transformer
+
+        Examples
+        --------
+        ```pycon
+        >>> transformer = transformer = StringContainsTransformer(columns=["a", "b"], reference="c")
```
ChloeCrisp, line 593: Should this be double assigned? 

`transformer = transformer =`

tubular/strings.py
```diff
@@ -1,583 +1,816 @@
...
+    """
+
+    polars_compatible = False
+
+    lazyframe_compatible = False
+
+    jsonable = False
+
+    deprecated = True
+
```
ChloeCrisp, line 762: Similar to above comment about FITS:

This does not include `FITS = False` where other functions do include this. Should this be included here?

tubular/strings.py
```diff
@@ -1,583 +1,816 @@
...
+    @beartype
+    def __init__(
+        self,
+        columns: str | ListOfStrs,
+        new_column_name: str = "new_column",
+        separator: str = " ",
+        **kwargs: bool,
```
ChloeCrisp, line 769: Should this be `**kwargs: bool | None` ?

ChloeCrisp (changes requested):

tubular/strings.py
```diff
@@ -1,583 +1,816 @@
...
+        return X
+
+
+@register
+class StringContainsTransformer(BaseTransformer):
+    r"""Transformer class to indicate if given columns contain reference values.
+
+    Attributes
+    ----------
+    reference: str
```
Boluwatife28, line 487: S: We should add descriptions for "reference" and "reference_as_column" to improve clarity

tubular/strings.py
```diff
@@ -1,583 +1,816 @@
...
+
+        Parameters
+        ----------
+        columns: Union[str, ListOfStrings]
+            columns to remove characters from.
+
+        reference: str
+            reference value to search for
+
+        reference_as_column: bool
```
Boluwatife28, line 570: S: Might be worth mentioning here also that this option is only supported for the Polars backend

tubular/strings.py
```diff
@@ -1,583 +1,816 @@
...
+
+        ```
+
+        """
+        X = _convert_dataframe_to_narwhals(X)
+
+        backend = nw.get_native_namespace(X).__name__
+
+        if backend == "pandas" and self.reference_as_column:
+            msg = f"{self.classname()}: reference_as_column=True is only supported for polars backend"
```
Boluwatife28, line 704: Q: Just curious, is the reason this isn't supported for pandas because the implementation relies on Narwhals/ Polars expressions?

tests/strings/test_StringContainsTransformer.py
```diff
@@ -0,0 +1,265 @@
...
+        self, lazy, library, from_json, input_values, reference_value, expected_output
+    ):
+        "test output cases for transformer when reference_column is True"
```
Boluwatife28, line 150: Q: Should the test description say "reference is False" here? The transformer is instantiated with "reference_as_column=False"

PR #724, opened 2026-04-30
Feature/string components

liamholmes31:
added ExtractStringComponentsTransformer, useful for e.g. pulling out email domains

### DIFF
--- a/tubular/functions/strings.py
+++ b/tubular/functions/strings.py
@@ -25,3 +25,20 @@ def remove_characters_from_string_columns(
 
     """
     return [nw.col(col).str.replace_all(characters_formatted, "") for col in columns]
+
+
+def extract_string_components(
+    columns: list[str], by: str, return_n_components: int
+) -> list[nw.Expr]:
+    """Get expression for extracting components from a str columns, split by provided character.
+
+    Returns
+    -------
+    list[nw.Expr]: expressions for transformation
+
+    """
+    return [
+        nw.col(col).str.split(by=by).list.get(i).alias(f"{col}_split_by_{by}_entry_{i}")
+        for col in columns
+        for i in range(return_n_components)
+    ]
--- a/tubular/functions/strings.py
+++ b/tubular/functions/strings.py
@@ -37,8 +37,13 @@ def extract_string_components(
     list[nw.Expr]: expressions for transformation
 
     """
+    split_exprs = {col: nw.col(col).str.split(by=by) for col in columns}
     return [
-        nw.col(col).str.split(by=by).list.get(i).alias(f"{col}_split_by_{by}_entry_{i}")
+        nw.when(split_exprs[col].list.len() > i)
+        .then(split_exprs[col])
+        .otherwise(None)
+        .list.get(i)
+        .alias(f"{col}_split_by_{by}_entry_{i}")
         for col in columns
         for i in range(return_n_components)
     ]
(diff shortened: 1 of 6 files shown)


tubular/functions/strings.py
```diff
@@ -25,3 +25,25 @@ def remove_characters_from_string_columns(
 
     """
     return [nw.col(col).str.replace_all(characters_formatted, "") for col in columns]
+
+
+def extract_string_components(
+    columns: list[str], by: str, return_n_components: int
```
Ritunjai-Sharma: Q: We currently have no validation checks for return_n_components. So it could be set to 0 or negative and no error would be raised. Do we want to add checks/guard for this or is it a user responsibility to choose sensible values?

tubular/strings.py
```diff
@@ -155,6 +156,228 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+        by: str,
+        return_n_components: int,
+        **kwargs: bool | None,
+    ) -> None:
+        """Initialise class instance.
+
+        Parameters
+        ----------
+        columns: Union[str, ListOfStrings]
+            columns to remove characters from.
```
Ritunjai-Sharma: Q: This seems outdated? Shouldn't it be something like: columns containing string values to split into components.

tubular/strings.py
```diff
@@ -155,6 +156,228 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+            return_n_components=self.return_n_components,
+        )
+
+    def transform(self, X: DataFrame) -> DataFrame:
+        r"""Extract components from string columns, split by given character.
+
+        Parameters
+        ----------
+        X : DataFrame
+            Data containing columns to strip.
```
Ritunjai-Sharma: Q: This seems outdated too!

tubular/strings.py
```diff
@@ -155,6 +156,228 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+
+        Returns
+        -------
+        X : DataFrame
+            Transformed input X with characters stripped from specified columns.
```
Ritunjai-Sharma: Q: This seems outdated too!

tubular/functions/strings.py
liamholmes31: very good point, have updated here https://github.com/azukds/tubular/pull/724/changes/206d189434467df5f06ccbaddb8c14841c64f8c7

tubular/strings.py
liamholmes31: thanks, updated here https://github.com/azukds/tubular/pull/724/changes/206d189434467df5f06ccbaddb8c14841c64f8c7#diff-da69eea07afacae563685991b3fdaa846a4b88fa29bf5b4131578cf8ddb3a4eb

tubular/strings.py
liamholmes31: thanks, updated here https://github.com/azukds/tubular/pull/724/changes/faa46f2a55c94c280d91bfde6be992de6908679a

tubular/strings.py
liamholmes31: thanks, updated here https://github.com/azukds/tubular/pull/724/changes/faa46f2a55c94c280d91bfde6be992de6908679a#diff-da69eea07afacae563685991b3fdaa846a4b88fa29bf5b4131578cf8ddb3a4eb

Ritunjai-Sharma (approved):

PR #728, opened 2026-05-08
added get_transform_exprs methods for easier classes

liamholmes31:
added get_transform_exprs methods to all transformers where this was simple/just involved moving code around, moved expression logic into new functions module so this can be used separately to transformer classes.

New functions are tested through the existing transformer tests, so proposal is to not test separately

### DIFF
--- a/tubular/comparison.py
+++ b/tubular/comparison.py
@@ -2,9 +2,7 @@
 
 from __future__ import annotations
 
-import operator
-from enum import Enum
-from typing import TYPE_CHECKING, Annotated, Any, ClassVar
+from typing import TYPE_CHECKING, Any
 
 import narwhals as nw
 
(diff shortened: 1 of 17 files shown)


tubular/comparison.py
```diff
@@ -439,33 +441,9 @@ def transform(self, X: DataFrame) -> DataFrame:
             )
             raise TypeError(message)
 
-        null_filter_expr = (
-            nw.col(self.columns[0]).is_null() | nw.col(self.columns[1]).is_null()
-        )
+        transform_expr = self.get_transform_exprs()
```
Boluwatife28, line 446: Q: Just me being curious, the previous implementation had backend specific handling for polars and Pandas to ensure the output column was cast to Boolean. Was this intentionally removed because Narwhals now handles the dtype consistently across backends?

tubular/functions/dates.py
```diff
@@ -0,0 +1,499 @@
...
+
+        The required ranges for each mapping are:
+            timeofday: 0-23
+            timeofmonth: 1-31
+            timeofyear: 1-12
+            dayofweek: 1-7
+
+        If an option is present in 'include' but no mappings are provided,
+        then default values from DEFAULT_MAPPINGS will be used for this
+        option.
```
Boluwatife28, line 289: P: really clear and detailed docstring :)

Boluwatife28 (commented):
Can we merge main into this branch to resolve the conflicts? :)

tubular/comparison.py
liamholmes31: in the past we sometimes hit issues with nullable bool columns in pandas being handled in weird ways (e.g. nulls being converted to False), i'm not completely sure what changed here but as an experiment I removed the casting and it seems to be OK (the tests do cover nullable bool cases so it should be covered)

liamholmes31:
have resolved conflicts :)

Boluwatife28 (approved):

Issue #648, opened 2025-12-12
Capping routines to use 'clip' method


adamsardar:
**What**

Our capping routines should make much greater use of the [clip](https://narwhals-dev.github.io/narwhals/api-reference/expr/#narwhals.Expr.clip) expression

**Why**

Currently we lean on when/then blocks. These are inefficient, complicated and less type-safe. They're also very difficult to extend because the logic can be very subtle.

Several eager checks that we make would be naturally covered by `clip`, which will help with processes such as #528 

**How**

I think that this could be quite a tricky task because there's complex behaviour expected.

Several methods that perform eager checks should be adapted to expressions.

liamholmes31:
Actioned as part of this one https://github.com/azukds/tubular/pull/729

PR #729, opened 2026-05-08
added get_transform_exprs method to capping transformers

liamholmes31:
- added get_transform_exprs method to capping transformers
- moved stateless parts of logic to function module
- converted OutOfRangeNullTransformer and CappingTransformer to use different/more specialised transform logic
    - this point relates to this one https://github.com/azukds/tubular/issues/648

### DIFF
--- a/tubular/capping.py
+++ b/tubular/capping.py
@@ -4,12 +4,10 @@ from __future__ import annotations
 
 import copy
 import warnings
-from typing import Annotated
 
 import narwhals as nw
 import numpy as np
 from beartype import beartype
-from beartype.vale import Is
 
 from tubular._stats import _weighted_quantile_expr
 from tubular._utils import (
@@ -19,24 +17,14 @@ from tubular._utils import (
     _return_narwhals_or_native_dataframe,
 )
 from tubular.base import block_from_json, register
+from tubular.functions.capping import (
+    CappingValues,
+    cap_columns,
+    set_out_of_range_to_none,
+)
 from tubular.mixins import WeightColumnMixin
 from tubular.numeric import BaseNumericTransformer
-from tubular.types import DataFrame, LazyFrame, Number, Series
-
-CappingValues = Annotated[
-    list[Number | None],
-    Is[
-        lambda list_arg: (
-            (len(list_arg) == 2)  # noqa: PLR2004
-            & (
-                all(
-                    (isinstance(value, (int, float)) or value is None)
-                    for value in list_arg
-                )
-            )
-        )
-    ],
-]
+from tubular.types import DataFrame, FloatTypeAnnotated, LazyFrame, Number, Series
 
 
 @register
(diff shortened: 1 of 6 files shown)


tubular/functions/capping.py
```diff
@@ -0,0 +1,106 @@
...
+) -> list[nw.Expr]:
+    """Get expression for capping columns within provided ranges.
+
+    Parameters
+    ----------
+    columns:
+        columns to cap
+
+    column_capping_ranges:
+        dict containing per column capping ranges
```
Ritunjai-Sharma: S: worth mentioning here and also in docstring of set_out_of_range_to_none that these are [lower_bound, upper_bound] pairs....

tubular/functions/capping.py
```diff
@@ -0,0 +1,106 @@
...
+        .cast(getattr(nw, dtype))
+        .alias(col)
+        if (column_capping_ranges[col][0] is not None)
+        else nw.when(nw.col(col) > column_capping_ranges[col][1])
+        .then(None)
+        .otherwise(nw.col(col))
+        .cast(getattr(nw, dtype))
+        .alias(col)
+        if (column_capping_ranges[col][1] is not None)
+        else nw.col(col).cast(getattr(nw, dtype))
```
Ritunjai-Sharma: Q: This fourth branch doesn't seem to be followed with .alias(col). So when neither bound exists, the column won't be renamed back to its original name? If so, worth adding .alias(col) here?

tubular/functions/capping.py
```diff
@@ -0,0 +1,106 @@
...
+            upper_bound=column_capping_ranges[col][1],
+        )
+        for col in columns
+    ]
+
+
+@beartype
+def set_out_of_range_to_none(
+    columns: list[str],
+    column_capping_ranges: dict[str, CappingValues] | None,
```
Ritunjai-Sharma: Q: Is this optional None intended here? Asking since for example, if a caller passes column_capping_ranges = None, then the code `column_capping_ranges[col][0] is not None` would fail immediately with TypeError: 'NoneType' object is not subscriptable before the bounds check can even run, because [None[col]] is invalid.

tubular/capping.py
Ritunjai-Sharma, line 236: Q: Not related to this PR but this seems duplicated!

tubular/functions/capping.py
liamholmes31: I think because the expression starts with `nw.col(col)` it just inherits the original name from that :)

tubular/functions/capping.py
liamholmes31: good shout, updated here https://github.com/azukds/tubular/pull/729/changes/1da644e9cfb0f3b41d22c453d9f0ea2f92040892

tubular/functions/capping.py
liamholmes31: ah thanks think this was left in by accident, removed here https://github.com/azukds/tubular/pull/729/changes/70d0d3d8c3842d419d1358011eabf31b769458ee

tubular/capping.py
liamholmes31: good spot, removed here https://github.com/azukds/tubular/pull/729/changes/5f245985924a102fbf03daee82316b843defc56e

Ritunjai-Sharma (approved):

PR #741, opened 2026-06-24
fixed bug for strings with nulls in ColumnDtypeStter

liamholmes31:
fixed error for casting columns with nulls to string, previously was being silently cast to string "None"


tests/misc/test_ColumnDtypeSetter.py
```diff
@@ -148,6 +148,85 @@ def test_expected_output(
                 df_expected_row.to_native(),
             )
 
+    @pytest.mark.parametrize(
+        ("a_values", "b_values", "expected_a_values", "expected_b_values"),
+        [
+            (["a", "b"], ["c", "d"], ["a", "b"], ["c", "d"]),
+            ([None, None], [None, None], [None, None], [None, None]),
```
Ritunjai-Sharma, line 156: S: Should we also add a mixed case here like ["a", None] / [["c", None]] for testing as it could be a common real world scenario?

liamholmes31: good shout, added here https://github.com/azukds/tubular/pull/741/changes/be17aa517e199d4c2eccf79202702c0a7c1b71f4

Ritunjai-Sharma (approved):

PR #743, opened 2026-07-06
edited transformers to sort dictionary attrs during to_json process, …

liamholmes31:
…to give consistent output

### DIFF
--- a/tests/utils/test_sort_nested_dict.py
+++ b/tests/utils/test_sort_nested_dict.py
@@ -1,0 +1,20 @@
+from tubular._utils import _sort_nested_dict
+
+
+def test_output():
+    "test output of function."
+
+    dictionary = {
+        "b": {"b": 1, "a": 2, None: 3, "z": 4},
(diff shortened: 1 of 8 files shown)


tests/utils/test_sort_nested_dict.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dictionary = {
+        "b": {"b": 1, "a": 2, None: 3, "z": 4},
+        None: {"z": "a", "t": "b"},
+        "c": {},
+    }
+
+    output = _sort_nested_dict(dictionary)
+
+    assert output == dictionary
```
Boluwatife28, line 15: Q: this assert statement doesn't test the sorting (dict == ignores order), is it just a sanity check to confirm the output values/keys are unchanged?

liamholmes31: yeah that's right, I test the order with the assert statements below this :)

Boluwatife28 (approved):

PR #746, opened 2026-07-09
Feature/refactor mapping

liamholmes31:
refactor in order to fix a couple of bugs with mappingtransformers null handling:

- all null mappings were erroring due to type being inferred as Null, which does not exist in narwhals
- mapping to String type columns where nulls/nans were present was causing issues like NaN->"nan", or erroring due to NaN values being incompatible with string type

Was easiest to fix by splitting out logic by type, this was on my todo list anyway and is a stepping stone towards having 'get_transform_exprs' methods for this module

### DIFF
--- a/tests/mapping/test_MappingTransformer.py
+++ b/tests/mapping/test_MappingTransformer.py
@@ -1,4 +1,5 @@
 import narwhals as nw
+import numpy as np
 import pytest
 
 import tests.test_data as d
@@ -89,6 +90,88 @@ class TestTransform(BaseMappingTransformerTransformTests, ReturnNativeTests):
         return_dtypes = {"a": "String", "b": "Int8"}
 
         transformer = MappingTransformer(mappings=mapping, return_dtypes=return_dtypes)
+
+        if _check_if_skip_test(transformer, df, lazy=lazy, from_json=False):
+            return
+
+        transformer = _handle_from_json(transformer, from_json)
+
+        df_transformed = transformer.transform(_convert_to_lazy(df, lazy=lazy))
+
+        assert_frame_equal_dispatch(_collect_frame(df_transformed, lazy=lazy), expected)
+
+        df = nw.from_native(df)
+        expected = nw.from_native(expected)
+
+        # also check single rows
+        for i in range(len(df)):
+            df_transformed_row = transformer.transform(
+                _convert_to_lazy(df[[i]].to_native(), lazy=lazy)
+            )
+            df_expected_row = expected[[i]].to_native()
+
+            assert_frame_equal_dispatch(
+                _collect_frame(df_transformed_row, lazy=lazy),
+                df_expected_row,
+            )
+
+    @pytest.mark.parametrize("lazy", [True, False])
+    @pytest.mark.parametrize("from_json", [True, False])
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_expected_output_null_type(self, library, from_json, lazy):
+        """Test that transform is giving the expected output when type of mappings is null."""
+
+        df_dict = {"a": [np.nan, None, 1]}
+        expected_df_dict = {"a": [None, None, 1]}
+        df = dataframe_init_dispatch(dataframe_dict=df_dict, library=library)
+        expected = dataframe_init_dispatch(
+            dataframe_dict=expected_df_dict, library=library
+        )
+
+        mapping = {"a": {np.nan: None}}
+
+        transformer = MappingTransformer(mappings=mapping)
+
+        if _check_if_skip_test(transformer, df, lazy=lazy, from_json=False):
+            return
+
+        transformer = _handle_from_json(transformer, from_json)
+
+        df_transformed = transformer.transform(_convert_to_lazy(df, lazy=lazy))
+
+        assert_frame_equal_dispatch(_collect_frame(df_transformed, lazy=lazy), expected)
+
+        df = nw.from_native(df)
+        expected = nw.from_native(expected)
+
+        # also check single rows
+        for i in range(len(df)):
+            df_transformed_row = transformer.transform(
+                _convert_to_lazy(df[[i]].to_native(), lazy=lazy)
+            )
+            df_expected_row = expected[[i]].to_native()
+
+            assert_frame_equal_dispatch(
+                _collect_frame(df_transformed_row, lazy=lazy),
+                df_expected_row,
+            )
+
+    @pytest.mark.parametrize("lazy", [True, False])
+    @pytest.mark.parametrize("from_json", [True, False])
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_null_handling_for_str_return_type(self, library, from_json, lazy):
+        """Test that transform is giving the expected output - for string type with nulls."""
+
+        df_dict = {"a": [np.nan, None, 1]}
+        expected_df_dict = {"a": [None, None, "b"]}
+        df = dataframe_init_dispatch(dataframe_dict=df_dict, library=library)
+        expected = dataframe_init_dispatch(
+            dataframe_dict=expected_df_dict, library=library
+        )
+
+        mapping = {"a": {1: "b"}}
+
+        transformer = MappingTransformer(mappings=mapping)
 
         if _check_if_skip_test(transformer, df, lazy=lazy, from_json=False):
             return
--- a/tests/mapping/test_MappingTransformer.py
+++ b/tests/mapping/test_MappingTransformer.py
@@ -421,6 +421,33 @@ class TestTransform(BaseMappingTransformerTransformTests, ReturnNativeTests):
 
         assert_frame_equal_dispatch(expected, _collect_frame(df_transformed, lazy=lazy))
 
+    @pytest.mark.parametrize("lazy", [True, False])
+    @pytest.mark.parametrize("from_json", [True, False])
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    def test_error_for_bad_type(self, library, from_json, lazy):
+        """Test expected error is raised for unexpected type."""
+        df_dict = {"a": ["01/02/2020", "20/03/1990"]}
+        df = dataframe_init_dispatch(dataframe_dict=df_dict, library=library)
+        df = nw.from_native(df)
+        df = df.with_columns(nw.col("a").str.to_datetime(format="%d/%m/%Y")).to_native()
+
+        mapping = {"a": {"a": "aaa"}}
+        return_dtypes = {"a": "Categorical"}
+
+        transformer = MappingTransformer(mappings=mapping, return_dtypes=return_dtypes)
+
+        if _check_if_skip_test(transformer, df, lazy=lazy, from_json=False):
+            return
+
+        transformer = _handle_from_json(transformer, from_json)
+
+        msg = r"MappingTransformer: The following columns have types which are not covered by the existing mapping logic \['a'\]"
+        with pytest.raises(
+            RuntimeError,
+            match=msg,
+        ):
+            transformer.transform(_convert_to_lazy(df, lazy=lazy))
+
 
 class TestOtherBaseBehaviour(
     OtherBaseBehaviourTests,
--- a/tubular/functions/mapping.py
+++ b/tubular/functions/mapping.py
@@ -1,0 +1,270 @@
+"""Stateless mapping transforms."""
+
+from typing import Any, Literal
+
+import narwhals as nw
+from beartype import beartype
+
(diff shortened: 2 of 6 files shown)


tubular/mapping.py
```diff
@@ -309,7 +313,7 @@ class attribute, indicates whether transformer works with lazyframes
     jsonable = False
 
     @beartype
-    def transform(
+    def transform(  # noqa: PLR0914, will fix in future refactor
```
Ritunjai-Sharma: N: Has this been noted down in a future refactor ticket?

tubular/mapping.py
```diff
@@ -206,13 +201,22 @@ def _infer_return_type(
...
         'Int64'
 
         ```
 
         """
-        return str(pl.Series(mappings[col].values()).dtype)
+        dtype = str(pl.Series(self.mappings[col].values()).dtype)
+
+        if dtype == "Null":
+            dtype = "Float64"
```
Ritunjai-Sharma, line 211: So _infer_return_type defaults to Float64 when all mapping values are None? If unmapped values remain in the column (from the otherwise branch), .cast(nw.Float64) will crash on non-numeric values?

For example: mappings={"status": {"active": None, "inactive": None}} would have dtype inferred as "Float64". If the column also contains "pending" (unmapped), it passes through the otherwise branch unchanged as a string, then .cast(nw.Float64) fails because "pending" can't be cast to float?

Or are we putting the responsibility on the user to set appropriate return_dtypes if mapping values to None?

tubular/functions/mapping.py
```diff
@@ -0,0 +1,270 @@
...
+            mappings_from_null=mappings_from_null,
+        ).cast(getattr(nw, return_dtypes[col]))
+        for col in cols
+    ]
+
+
+@beartype
+def map_generic_to_categorical(
+    cols: list[str],
+    return_dtypes: dict[str, RETURN_DTYPES],
```
Ritunjai-Sharma: Q: The return_dtypes parameter here accepts any RETURN_DTYPES value (including "Float64", "Int32", etc.), but the function is semantically intended for categorical targets right? While the upstream routing should only pass "Categorical", there's no guard here at the function level....

tubular/functions/mapping.py
```diff
@@ -0,0 +1,270 @@
...
+
+    Returns
+    -------
+    list[nw.Expr]: expressions for transformation
+
+    """
+    return [
+        _get_mapping_expr(
+            col,
+            pre_mapping_expr=nw.col(col).fill_nan(None),
```
Ritunjai-Sharma: Q: map_number_to_string has fill_nan(None) because NaN in numeric source → string target causes "nan" stringification. But what about numeric source → numeric target (e.g. Float64 column with NaN being mapped to Int32)?

For example:
```
# Source: [1.0, NaN, 2.0]
# Mapping: {1.0: 10, 2.0: 20}
# NaN is not in mapping → passes through via otherwise → still NaN
# .cast(nw.Int32) on NaN → ?
```

So NaN might survive and cause issues when casting to integer types. Do we know if narwhals/polars handles NaN → null automatically during integer casts, or if fill_nan(None) should also be applied here too?

tubular/mapping.py
```diff
@@ -78,19 +87,6 @@ class attribute, indicates whether transformer works with lazyframes
 
     jsonable = True
 
-    RETURN_DTYPES = Literal[
-        "String",
-        "Object",
```
Ritunjai-Sharma, line 80: The routing handles:

- "Categorical" → map_generic_to_categorical
- "String" → map_number_to_string or map_generic_to_str
- "Boolean" → map_generic_to_bool
- "Int8" through "Float64" → map_generic_to_number

But what about "Object"? The old RETURN_DTYPES included "Object" but was removed from the new module's RETURN_DTYPES. If someone passes return_dtypes={"col": "Object"} (or if legacy JSON configs contain it), the column would silently fall through every routing check and never get mapped?

Should we add validation that all columns were routed (assert len(mapping_exprs) == len(self.return_dtypes)) and add a catch-all/else that raises an error for unrecognised dtypes? Or should we be putting the responsibility on the user?

tests/mapping/test_MappingTransformer.py
```diff
@@ -1,4 +1,5 @@
 import narwhals as nw
```
Ritunjai-Sharma, line 3: I had a few suggestions for additional tests that may be worth adding:
1. **All-None mapping with unmapped string values**:
The existing test_expected_output_null_type has [np.nan, None, 1] where the unmapped value (1) is numeric, so .cast(nw.Float64) succeeds. How about we add a test with a string unmapped value (e.g. ["active", "inactive", "pending"] with mappings={"a": {"active": None, "inactive": None}}). Here, "pending" would pass through otherwise, then .cast(nw.Float64) would crash. This test would then exercise whether the Float64 default in _infer_return_type is actually safe.

2. **NaN in numeric source mapped to integer target**:
map_number_to_string has fill_nan(None) but map_generic_to_number doesn't. Should we add a test with e.g. [1.0, NaN, 2.0], mappings={1.0: 10, 2.0: 20}, return_dtypes={"a": "Int32"}. Since NaN would be unmapped, it would survive through otherwise, then .cast(nw.Int32) may fail. This test would confirm whether fill_nan(None) is also needed in the numeric→numeric path.

3. **Categorical source to non-categorical target**:
A column with categorical dtype source but "Int64" target gets routed to map_generic_to_categorical (due to the source-type check). Should we add a test confirming the full path works: pre-cast to String → key matching → mapping → .cast(nw.Int64)? Also worth testing with a partially unmapped categorical (e.g. ["low", "medium", "high", "extreme"] where "extreme" has no mapping).... The unmapped string might fail on .cast(nw.Int64)?

tubular/mapping.py
liamholmes31: should be covered by this one, where we'll split out into separate classes https://github.com/azukds/tubular/issues/753

tubular/mapping.py
liamholmes31: yeah that is a risk, I'm not sure there's a way around this though - the warning message does direct users to use the return_dtypes arg which might be the best we can do for this scenario, what do you think?

tubular/functions/mapping.py
liamholmes31: dug into this a bit, you are right this does currently cause errors. It would be possibly to fix this for polars by doing fill_nan(None) as you suggest, but this would still error for pandas which defaults to non-nullable int types. 

I think I would suggest just leaving this to error for now, as the message you get is quite informative:
narwhals.exceptions.InvalidOperationError: conversion from `f64` to `i8` failed in column 'a' for 1 out of 2 values: [NaN]

and we can revisit when we split this transformer up into separate classes - what do you think?

tubular/functions/mapping.py
liamholmes31: have updated the numeric function to have correct return dtypes

for the categorical one, realised it was actually badly setup, so have changed to `map_categorical_to_generic`. The type hint is now correct, and have edited the other logic to fit this new name

https://github.com/azukds/tubular/pull/746/changes/0578de33c393a4f944c7a5a6cc2eea67160bd306

tubular/mapping.py
liamholmes31: yeah narwhals doesn't support casting to object, I think because it's a pandas catch-all legacy type which has all kinds of problems. We are better moving away from this type generally, I don't think polars has an equivalent either.

I've added a check that all columns are processed properly here, to make sure things don't silently slip through https://github.com/azukds/tubular/pull/746/changes/0578de33c393a4f944c7a5a6cc2eea67160bd306

tests/mapping/test_MappingTransformer.py
liamholmes31: working on these!

for 2, this links to discussion above so will leave out for now

Update - added a test for this https://github.com/azukds/tubular/pull/746/changes/c69c0f01d09e2aa4f7473779ae28925928127664

liamholmes31: reading through 1, think this is also related to another comment thread so will leave for now

Update - this case is now blocked in the code by the need to provide a full mapping

tubular/functions/mapping.py
```diff
@@ -0,0 +1,501 @@
...
+
+    return (
+        mapping_expr.fill_null(mappings_from_null[col])
+        if mappings_from_null[col] is not None
+        else mapping_expr
+    )
+
+
+@beartype
+def _get_partial_mapping_expr(
```
liamholmes31, line 103: have updated to give options for both full or partial mappings, because e.g.:
- to map a str col to bool, you would need to map all values (as e.g. bool('hello') would error)
- to map a float to str, you can get away with a partial mapping as e.g. an unmapped 1.0 would just go to "1.0"

tubular/functions/mapping.py
```diff
@@ -0,0 +1,501 @@
...
+    )
+
+    return (
+        mapping_expr.fill_null(mappings_from_null[col])
+        if mappings_from_null[col] is not None
+        else mapping_expr
+    )
+
+
+def map_integer_columns(
```
liamholmes31, line 161: Have now split out the mapping functions by starting type. The alternative was to have a function for each start,end type pair, which felt like it would bloat the API too much and confuse users - hopefully this is a good middle ground!

tubular/functions/mapping.py
```diff
@@ -0,0 +1,501 @@
...
+        Controls verbosity of function.
+
+    Returns
+    -------
+    list[nw.Expr]: expressions for transformation
+
+    """
+    if verbose and "Boolean" in return_dtypes.values():
+        warnings.warn(
+            """map_integer_columns: Note if working in pandas and casting to Boolean,
```
liamholmes31, line 196: As these functions do not see data, the best we can do for pandas users is to throw a warning about bad types..

tests/nominal/test_OrdinalEncoderTransformer.py
```diff
@@ -1,249 +0,0 @@
-from typing import Literal
```
liamholmes31: Bolu is currently reviewing another PR which deletes this one, have just deleted here as well to avoid having to fix tests pointlessly

tubular/mapping.py
```diff
@@ -346,73 +356,91 @@ def transform(
...
-
-        mapping_exprs = {
-            col: nw.when(mappable_conditions[col])
-            .then(
-                # default here allows replace_strict to work, but the nulls are replaced
-                # in the otherwise section anyway
-                mapping_exprs[col].replace_strict(self.mappings[col], default=None)
+
+        if backend == "pandas":
+            bool_input = any(
```
liamholmes31, line 359: have blocked non-nullable pandas types where bools are involved, they introduce a lot of headaches that are hard to handle in the new functional approach

tubular/mapping.py
liamholmes31: have now updated this so that casts from str->float require a full mapping to be provided, so this will go through the error handling for expr.replace_strict()

tubular/functions/mapping.py
liamholmes31: this function has now been replaced, but I think the new functions have correct type hints for return_dtypes

tubular/functions/mapping.py
liamholmes31: have added a test case to verify that this now works for polars and pandas-nullable types https://github.com/azukds/tubular/pull/746/changes/c69c0f01d09e2aa4f7473779ae28925928127664

non-nullable pandas types will error as described above, but I think this is OK

tests/mapping/test_MappingTransformer.py
liamholmes31: for 3, have added the test function 'test_expected_output_for_categorical_input' - it's spread across a few commits, but can be found in test_MappingTransformer.py

Ritunjai-Sharma: For 3, has the following kind of scenario been tested:

- Categorical source (e.g., ["low", "medium", "high", "extreme"])
- Partial mappings (e.g., {"low": 1, "medium": 2, "high": 3} with no mapping for "extreme")
- Int64 target dtype

The unmapped string value ("extreme") might fail when trying to .cast(nw.Int64) during the mapping→casting pipeline?

liamholmes31: this would go through the _get_full_mapping_expr logic, so would be expected to hit the 'need to provide full mappings' error from expr.replace_strict :)

this error would be handled outside of tubular, so have left untested atm, but the narwhals/polars error handling looks informative enough for users. What do you think?

Ritunjai-Sharma (approved):

PR #747, opened 2026-07-14
Fix schema type membership checks and extend GroupRareLevelsTransformer to accept Enum columns

ozzyholemans:
Due to the implementation from narwhals of the `Enum` dtype, a type check written like: `schema[col] in {nw.Enum}` would return `False`, even when `col` is actually of Enum type.

This is because when doing a set lookup Python uses the hash, however `hash(nw.Enum(["a"])) != hash(nw.Enum)`, leading to Python not finding the Enum dtype in the set.

This replaces all affected areas with usage of `isinstance` instead. It also extends the definition of "string-like" columns in `GroupRareLevelsTransformer` to include Enum-type columns, as the `transform` method is written to expect.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -32,6 +32,7 @@ Each individual change should have a link to the pull request after the descript
 
 Changed
 ^^^^^^^
+- bugfix: use isinstance for schema type membership checks and allow enum columns for GroupRareLevelsTransformer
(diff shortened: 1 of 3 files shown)


liamholmes31:
looks good! would it be possible to also add an Enum test to MappingTransformer/GroupRareLevelsTransformer? that will prevent this from regressing in the future

PR #759, opened 2026-08-04
chore: refactored OtherBaseBehaviourTests to improve maintainability

liamholmes31:
made some simplifying changes to make upcoming imputer PRs easier - basically changed tests to utilise minimal_dataframe_lookup more rather than having type specific e.g. 'OtherBaseBehaviourTestsNumeric' classes

### DIFF
--- a/tests/imputers/test_BaseImputer.py
+++ b/tests/imputers/test_BaseImputer.py
@@ -589,6 +589,7 @@ class TestOtherBaseBehaviour(OtherBaseBehaviourTests):
         )
 
     # overload this test, as does not make sense for test of independent base class
+    @pytest.mark.skip
     def test_pipeline_raises_not_fitted_error_when_unfitted(self):
         return
 
--- a/tests/imputers/test_BaseImputer.py
+++ b/tests/imputers/test_BaseImputer.py
@@ -588,6 +588,10 @@ class TestOtherBaseBehaviour(OtherBaseBehaviourTests):
             f"{x.classname()}: get_feature_names_out does not agree with output of .transform, expected {expected_new_columns} but got {output_columns}"
         )
 
+    # overload this test, as does not make sense for test of independent base class
+    def test_pipeline_raises_not_fitted_error_when_unfitted(self):
+        return
+
     @classmethod
     def setup_class(cls):
         cls.transformer_name = "BaseImputer"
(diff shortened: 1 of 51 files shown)


tests/test_data.py
```diff
@@ -30,24 +30,55 @@ def create_numeric_df_2(library="pandas"):
     df_dict = {
         "a": [2, 3, 2, 1, 4, 1],
         "b": [None, None, 1, 3, 3, 4],
-        "c": [1, 1, 2, 3, 3, None],
+        "c": [1, 1, 2, 3, 3, 3],
```
Boluwatife28, line 33: Q: Just curious, was the None removed from c because its now used as the target?

tests/imputers/test_BaseImputer.py
```diff
@@ -588,6 +588,10 @@ def test_get_feature_names_out_matches_new_features(
             f"{x.classname()}: get_feature_names_out does not agree with output of .transform, expected {expected_new_columns} but got {output_columns}"
         )
 
+    # overload this test, as does not make sense for test of independent base class
```
Boluwatife28, line 591: Q: is the intention here to disable the inherited test for the base class? If so, would pytest.skip() better communicate that intent than returning immediately?

tests/mapping/test_BaseMappingTransformMixin.py
```diff
@@ -532,6 +532,10 @@ def test_get_feature_names_out_matches_new_features(
             f"{x.classname()}: get_feature_names_out does not agree with output of .transform, expected {expected_new_columns} but got {output_columns}"
         )
 
+    # overload this test, as does not make sense for test of independent mixin
+    def test_pipeline_raises_not_fitted_error_when_unfitted(self):
+        return
```
Boluwatife28, line 538: Q: same question as above :)

tests/test_data.py
liamholmes31: yes exactly!

tests/imputers/test_BaseImputer.py
liamholmes31: good point, added here https://github.com/azukds/tubular/pull/759/changes/e4ed1668be694830b489ef57a19abf9109763804

tests/mapping/test_BaseMappingTransformMixin.py
liamholmes31: good point, added here https://github.com/azukds/tubular/pull/759/changes/e4ed1668be694830b489ef57a19abf9109763804

Boluwatife28 (approved):

Issue #752, opened 2026-07-22
`get_transform_exprs` methods for ArbitraryImputer


liamholmes31:
What?
we want to be able to extract transformations offline for ArbitraryImputer, using a `get_transform_exprs` method

Why?
pave the way for working directly with exprs in transform

How? 
the logic for this class depends on the incoming data schema, so to pave the way for this issue we have split ArbitraryImputer into type specific subclasses under the hood. For this final part, we will probably need to:

- deprecate ArbitraryImputer
- remove `_` from ArbitraryImputer subclasses to make public, add @register decorator and expose in tubular.init
- move around ArbitraryImputer tests into appropriate subclass test files
- in the short term, I suggest just duplicating tests, and we will eventually delete the old ArbitraryImputer tests when we remove the class
- add `get_transform_exprs` methods to each of the new subclasses, which can hopefully make use of existing functions in functions.imputers

PR #761, opened 2026-08-04
Feature/numberimputer

liamholmes31:
added tests for NumberImputer, and exposed as public class 

For this one, mainly focused on replicating ArbitraryImputers test coverage for numeric columns, so can the reviewer please also spend some time reading through the old tests and double checking that we haven't lost any cases in the move. (I think ArbitraryImputer tests contained some duplicated cases, so I haven't copied over each test, but tried to condense down)

### DIFF
--- a/tubular/imputers.py
+++ b/tubular/imputers.py
@@ -261,7 +261,7 @@ class BaseImputer(BaseTransformer):
 
 
 class NumberImputer(BaseImputer):
-    """Private subclass to handle arbitrary number imputation.
+    """Class to handle arbitrary number imputation.
 
     Attributes
     ----------
--- a/tubular/imputers.py
+++ b/tubular/imputers.py
@@ -24,7 +24,10 @@ from tubular._utils import (
     block_from_json,
 )
 from tubular.base import BaseTransformer, register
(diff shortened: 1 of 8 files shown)


tubular/imputers.py
```diff
@@ -257,7 +260,7 @@ def transform(
         return _return_narwhals_or_native_dataframe(X, return_native)
 
 
-class _NumberImputer(BaseImputer):
+class NumberImputer(BaseImputer):
     """Private subclass to handle arbitrary number imputation.
```
lsumption: N: Should we update the docstring as it's no longer private?

tubular/imputers.py
lsumption: N: Update the name in the error message

tubular/imputers.py
liamholmes31: updated here :) https://github.com/azukds/tubular/pull/761/changes/f93222ab6c9380c01584b7b08194a8b764b47f5c

tubular/imputers.py
liamholmes31: updated here :) https://github.com/azukds/tubular/pull/761/changes/26194567d2d0aa2ccaac7d41f1cc2251c893a577

lsumption (approved):

PR #763, opened 2026-08-05
Type specific transform functions for GroupRareLevelsTransformer

### LINKED ISSUE #745: How to give GroupRareLevelsTransformer a 'get_transform_exprs' method
What?
The transform method of GroupRareLevelsTransformer has different logic paths for str columns and category/enum columns, but we want to implement a `get_transform_exprs` method which would have no knowledge of the data. What is the best way for us to achieve this? one option would be to have separate transformer classes for str/cat/enum columns, but this would introduce some complexity to the code (and to a user in knowing which class to bring to their data).

I think it's worth spending some time brainstorming on potential approaches before we proceed with this one.

Why?
make sure we take best path to implementing 'get_transform_exprs' methods

How?
spend some time reading through the transformer and considering design options, probably also worth discussing with chatgpt!

lsumption:
Split out transform functionality into type specific functions that are used within the main transform function.

Will act as a step towards having 'get_transform_exprs' methods for this transformer

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,7 +33,7 @@ Changed
 ^^^^^^^
 - added get_transform_exprs methods to NullIndicator transformer `#750 <https://github.com/azukds/tubular/issues/750>_`
 - added get_transform_exprs methods to BaseImputer `#751 <https://github.com/azukds/tubular/issues/751>_`
-- placeholder
+- chore: refactored GroupRareLevelsTransformer to pave way for splitting into type based classes with `get_transform_exprs` methods.
 
 3.9.0 (17/07/2026)
 ------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,8 +31,8 @@ Each individual change should have a link to the pull request after the descript
 -------------------
 Changed
 ^^^^^^^
-- added get_transform_exprs methods to NullIndicator transformer `#750 <https://github.com/azukds/tubular/issues/750>_`
-- added get_transform_exprs methods to BaseImputer `#751 <https://github.com/azukds/tubular/issues/751>_`
+- feat: added get_transform_exprs methods to NullIndicator transformer `#750 <https://github.com/azukds/tubular/issues/750>_`
+- feat: added get_transform_exprs methods to BaseImputer `#751 <https://github.com/azukds/tubular/issues/751>_`
 - chore: refactored GroupRareLevelsTransformer to pave way for splitting into type based classes with `get_transform_exprs` methods.
 
 3.9.0 (17/07/2026)
--- a/tests/nominal/test_GroupRareLevelsTransformer.py
+++ b/tests/nominal/test_GroupRareLevelsTransformer.py
@@ -369,6 +369,7 @@ class TestTransform(GenericNominalTransformTests):
         # set the mappging dict directly rather than fitting x on df so test works with decorators
         transformer.non_rare_levels = {"b": ["a"], "c": ["e", "c", "a"]}
         transformer.rare_levels_record = {}
+        transformer.training_data_levels = {}
         transformer = _handle_from_json(transformer, from_json)
         df_transformed = transformer.transform(_convert_to_lazy(df, lazy=lazy))
 
@@ -402,6 +403,7 @@ class TestTransform(GenericNominalTransformTests):
         # set the mapping dict directly rather than fitting x on df so test works with decorators
         transformer.non_rare_levels = {"b": ["a"]}
         transformer.rare_levels_record = {}
+        transformer.training_data_levels = {}
         transformer = _handle_from_json(transformer, from_json)
         df_transformed = transformer.transform(_convert_to_lazy(df, lazy=lazy))
 
--- a/tubular/functions/nominal.py
+++ b/tubular/functions/nominal.py
@@ -6,6 +6,201 @@ import narwhals as nw
 import numpy as np
 from beartype import beartype
 
+from tubular.types import ListOfStrs
+
+
+@beartype
+def _get_non_rare_condition_expression(
+    col: str,
+    non_rare_levels: dict[str, ListOfStrs],
+    unseen_levels_to_rare: bool,
+    training_data_levels: dict[str, ListOfStrs] | None,
+) -> nw.Expr:
+    """Get expression for non rare condition.
+
+    Parameters
+    ----------
+    col:
+        column to map
+
+    non_rare_levels :
+        dict of non rare levels per column
+
+    unseen_levels_to_rare:
+        whether to map unseen levels to rare
+
+    training_data_levels:
+        dict of training data levels per column
+
+    Returns
+    -------
+    nw.Expr: expression for non rare condition
+
+    """
+    return (
+        nw.col(col).is_in(non_rare_levels[col])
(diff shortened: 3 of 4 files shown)


tubular/nominal.py
```diff
@@ -468,47 +477,41 @@ def transform(self, X: DataFrame) -> DataFrame:
...
-                # levels (unseen)
-                else (
-                    nw.col(col).is_in(self.non_rare_levels[col])
-                    | ~nw.col(col).is_in(self.training_data_levels[col])
-                )
-            )
+        str_cols = [
+            col
+            for col in self.columns
+            if ((schema[col] == "String") or (isinstance(schema[col], nw.String)))
```
liamholmes31: S: think this would work with just the `isinstance` comparison

tubular/nominal.py
```diff
@@ -468,47 +477,41 @@ def transform(self, X: DataFrame) -> DataFrame:
...
+        ]
+        cat_enum_cols = [col for col in self.columns if col not in str_cols]
```
liamholmes31: S: would be more thorough to check `isinstance(schema[col], (nw.Categorical, nw.Enum)`

tubular/functions/nominal.py
```diff
@@ -6,6 +6,201 @@
...
+                rare_level_name=rare_level_name,
+                non_rare_condition_expression=non_rare_condition_expression,
+                non_rare_levels=non_rare_levels,
+                str_col=False,
+            )
+        )
+    return grouped_expressions
+
+
+def group_str_cat(
```
liamholmes31: N: to me something like `rare_encode_str_columns` would be a more descriptive name?

tubular/functions/nominal.py
```diff
@@ -6,6 +6,201 @@
...
+    if not str_col:
+        transform_expression = transform_expression.cast(
+            nw.Enum(non_rare_levels[col] + [rare_level_name]),
+        )
+
+    return transform_expression
+
+
+@beartype
+def group_rare_cat_enum(
```
liamholmes31: N: to me something like `rare_encode_categorical_or_enum_columns` would be a more descriptive name

tubular/nominal.py
```diff
@@ -229,6 +234,8 @@ def to_json(self) -> dict[str, dict[str, Any]]:
             json_dict["fit"]["training_data_levels"] = _sort_dict(
                 self.training_data_levels
             )
+        else:
+            json_dict["fit"]["training_data_levels"] = None
```
liamholmes31, line 238: Q: was this erroring somewhere?

tubular/functions/nominal.py
```diff
@@ -6,6 +6,201 @@
...
+
+    str_col:
+        whether the column is a string column
+
+    Returns
+    -------
+    list[nw.Expr]: expressions for transformation
+
+    """
+    pre_grouping_expr = nw.col(col) if str_col else nw.col(col).cast(nw.String)
```
liamholmes31: S: we could avoid having a branch here by passing in an expression instead of col/str_col, so e.g. categorical/enum would pass an arg like col_expr=nw.col(col).cast(nw.String) and string would pass nw.col(col)

tubular/functions/nominal.py
```diff
@@ -6,6 +6,201 @@
...
+
+    transform_expression = (
+        nw.when(non_rare_condition_expression | nw.col(col).is_null())
+        .then(pre_grouping_expr)
+        .otherwise(nw.lit(rare_level_name))
+    )
+
+    if not str_col:
```
liamholmes31: S: would move this cast into group_rare_cat_enum to avoid a branch

CHANGELOG.rst
```diff
@@ -31,9 +31,10 @@ Each individual change should have a link to the pull request after the descript
 -------------------
 Changed
 ^^^^^^^
-- feat: added get_transform_exprs methods to NullIndicator transformer `#750 <https://github.com/azukds/tubular/issues/750>_`
-- feat: added get_transform_exprs methods to BaseImputer `#751 <https://github.com/azukds/tubular/issues/751>_`
+- added get_transform_exprs methods to NullIndicator transformer `#750 <https://github.com/azukds/tubular/issues/750>_`
```
liamholmes31: N: did you mean to remove the `feat:` tags here?

tests/nominal/test_GroupRareLevelsTransformer.py
```diff
@@ -368,6 +368,7 @@ def test_expected_output_no_weight(self, library, from_json, lazy):
         # set the mappging dict directly rather than fitting x on df so test works with decorators
         transformer.non_rare_levels = {"b": ["a"], "c": ["e", "c", "a"]}
         transformer.rare_levels_record = {}
+        transformer.training_data_levels = {}
```
liamholmes31, line 371: Q: did this start erroring somewhere?

tubular/nominal.py
lsumption: Good point, have removed the redundant check [here](https://github.com/azukds/tubular/pull/763/changes/540d30cd14fb1e10b68c8715a2c2714e7abd3605)

tubular/nominal.py
lsumption: Have also added this check [here](https://github.com/azukds/tubular/pull/763/changes/540d30cd14fb1e10b68c8715a2c2714e7abd3605)

tubular/functions/nominal.py
lsumption: I did have this before but ran into the too many function arg problem, but now we've simplified it with the below comment I've updated it to an expr [here](https://github.com/azukds/tubular/pull/763/changes/a0be364b4523798f299961712cfd0c6a25b2ff82)

tubular/functions/nominal.py
lsumption: Moved in [this commit](https://github.com/azukds/tubular/pull/763/changes/a0be364b4523798f299961712cfd0c6a25b2ff82)

tubular/functions/nominal.py
lsumption: Updated name [here](https://github.com/azukds/tubular/pull/763/changes/bfc0362115b78a9704504f499a42e6026e6f6a26)

tubular/functions/nominal.py
lsumption: Updated name [here](https://github.com/azukds/tubular/pull/763/changes/bfc0362115b78a9704504f499a42e6026e6f6a26)

CHANGELOG.rst
lsumption: Oops, something went funny with the merge conflict! Sorted [here](https://github.com/azukds/tubular/pull/763/changes/ee055d853f4a0505dab1383e9dcdbfe11da61149)

tests/nominal/test_GroupRareLevelsTransformer.py
lsumption: Yep as before we were only defining the training_data_levels attribute only when unseen_levels_to_rare was defined, but we need to set the attribute in order to pass it to the type specific functions.

tubular/nominal.py
lsumption: Yep, this is from the same issue we saw in the comment below with training_data_levels not always being set

liamholmes31 (approved):

PR #764, opened 2026-08-06
Feature/booleanimputer

liamholmes31:
made booleanimputer class public and added tests. Tests are adapted from existing ArbitraryImputer tests and have attempted to remove duplications, but is worth reviewer comparing new tests against old ArbitraryImputer tests to double check that we have not lost any coverage 

### DIFF
--- a/tests/imputers/test_BooleanImputer.py
+++ b/tests/imputers/test_BooleanImputer.py
@@ -1,0 +1,316 @@
+import re
+
+import narwhals as nw
+import pandas as pd
+import polars as pl
+import pytest
+
+from tests.base_tests import (
+    ColumnStrListInitTests,
+    EmptyColumnsFitTransformPassTests,
+    GenericFitTests,
+    GenericTransformTests,
+    OtherBaseBehaviourTests,
+    ReturnNativeTests,
+)
+from tests.utils import (
+    _check_if_skip_test,
+    _collect_frame,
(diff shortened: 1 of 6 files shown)


tests/imputers/test_BooleanImputer.py
```diff
@@ -0,0 +1,316 @@
...
+
+    @pytest.mark.parametrize("from_json", [True, False])
+    @pytest.mark.parametrize(
+        ("input_col", "impute_value", "expected_values"),
+        [
+            ([True, False, None], True, [True, False, True]),
+            ([True, False, None], False, [True, False, False]),
+        ],
+    )
+    def test_output_for_pandas_object_type(
```
lsumption, line 201: Is there a reason we don't check full df against expectation and check outcomes for single rows like we do in the other output tests here?

liamholmes31: good spot, added that here https://github.com/azukds/tubular/pull/764/changes/c281819d8ff61a5aed5b500bd90a1225bd1bd38a

lsumption: Thanks, what about the full df against expectation too?

lsumption: ^that was already done, so happy to approve

lsumption (approved):

PR #766, opened 2026-08-12
Feature/stringimputer

liamholmes31:
made stringimputer class public and added tests. Tests are adapted from existing ArbitraryImputer tests and have attempted to remove duplications, but is worth reviewer comparing new tests against old ArbitraryImputer tests to double check that we have not lost any coverage

Also added CategoricalImputer class with tests, and private _EnumImputer class to keep ArbitraryImputer working


tubular/imputers.py
```diff
@@ -261,7 +265,7 @@ def transform(
 
 
 class NumberImputer(BaseImputer):
-    """Class to handle arbitrary number imputation.
+    """Private subclass to handle arbitrary number imputation.
```
lsumption: Sorry super nitpicky as this review isn't for this transformer - but can we update this doc string

tubular/imputers.py
lsumption, line 496: Why are checking if col is not a str/cat/enum/unknown, but then the error msg says we can only handle str/unknown types?

tubular/imputers.py
```diff
@@ -747,22 +751,152 @@ class attribute, indicates whether transform requires fit to be run first
...
+
+        bad_types = [
+            schema[col]
+            for col in self.columns
+            if not isinstance(schema[col], (nw.Categorical, nw.Unknown))
+        ]
+
+        if bad_types:
+            msg = f"""
+                ArbitraryImputer: transformer can only handle Categorical/Unknown type columns
```
lsumption: Does this need updating?

tubular/imputers.py
```diff
@@ -816,54 +972,64 @@ def transform(self, X: DataFrame) -> DataFrame:
...
-                return_native=self.return_native,
+        bad_types = [
+            schema[col]
+            for col in self.columns
+            if not isinstance(schema[col], (nw.Enum, nw.Unknown))
+        ]
+
+        if bad_types:
+            msg = f"""
+                ArbitraryImputer: transformer can only handle Enum/Unknown type columns
```
lsumption: Same here, should this be EnumImputer?

tubular/imputers.py
```diff
@@ -747,22 +751,152 @@ class attribute, indicates whether transform requires fit to be run first
...
+
+        transform_exprs = self.get_transform_exprs()
+
+        X = X.with_columns(*transform_exprs) if transform_exprs else X
+
+        return _return_narwhals_or_native_dataframe(X, self.return_native)
+
+
+@register
+class _EnumImputer(BaseImputer):
```
lsumption, line 816: Is the plan to make this a public class in the future? Otherwise, how are we planning on handling enum imputation?

tubular/imputers.py
```diff
@@ -747,22 +751,152 @@ class attribute, indicates whether transform requires fit to be run first
...
+    """Class to handle arbitrary imputation into enum columns.
```
lsumption: If the plan is to keep this as private, can we update the docstring to say that

tubular/imputers.py
lsumption: Maybe worth double checking all the error messages to make sure it's the right transformer

tubular/imputers.py
lsumption: Probably a silly question, but why are we using this method only on the str and boolean transformers?

tests/conftest.py
```diff
@@ -417,7 +423,8 @@ def minimal_dataframe_lookup(request) -> dict[str, pd.DataFrame]:
     min_df_dict["BaseAggregationTransformer"] = agg_df
     min_df_dict["AggregateRowsOverColumnTransformer"] = agg_df
     min_df_dict["WhenThenOtherwiseTransformer"] = when_then_df
-    min_df_dict["_StringImputer"] = object_df
+    min_df_dict["StringImputer"] = object_df
+    min_df_dict["CategoricalImputer"] = categorical_df
     min_df_dict["_BooleanImputer"] = bool_df
```
lsumption, line 335: should we be adding tests for enum imputer too? (happy to be a follow up PR if we want)

tests/imputers/test_ArbitraryImputer.py
```diff
@@ -393,7 +436,7 @@ def test_expected_output_with_object_and_categorical_columns(
         df2 = d.create_df_2(library=library)
 
         args = minimal_attribute_dict[self.transformer_name]
-        args["columns"] = ["b", "c"]
+        args["columns"] = ["c"]
```
lsumption: Why have we dropped b here (and below)?

tests/imputers/test_CategoricalImputer.py
```diff
@@ -0,0 +1,282 @@
...
+
+    @pytest.mark.parametrize(
+        "lazy",
+        [True, False],
+    )
+    @pytest.mark.parametrize("from_json", [True, False])
+    @pytest.mark.parametrize("library", ["pandas", "polars"])
+    @pytest.mark.parametrize(
+        ("input_col", "impute_value", "expected_values"),
+        [
```
lsumption, line 114: Can we also add a test for imputing with an existing level?

lsumption:
Also looks like there's some conflicts that need resolving

tests/conftest.py
liamholmes31: my thinking was because this one is private/mainly exists atm to support ArbitraryImputer, that we can just rely on ArbitraryImputers tests. What do you think?

(If we ever made it public/delete arbitrary imputer then we would need to add it's own tests ofc)

lsumption: I guess it links to the comment I made above, how are we expecting people to handle arbitrary enum imputation?

tests/imputers/test_CategoricalImputer.py
liamholmes31: good idea, added here https://github.com/azukds/tubular/pull/766/changes/542946518d7ef62680dd6a8298c403aecaac33c6

tests/imputers/test_ArbitraryImputer.py
liamholmes31: so I removed as in the new setup you can't mix/match types in an ArbitraryImputer instance, but you're right that this did reduce test coverage so have readded here
https://github.com/azukds/tubular/pull/766/changes/3565841bc6e9fb5fc00e53b4376563c172b7c03b

for the second case, I don't think it's harmful to coverage to remove so have left that one as is

tubular/imputers.py
liamholmes31: not silly at all, looks like this PR wasn't as ready as I thought - sorry should have self reviewed first. I've now updated so that the imputer classes are using the new functions https://github.com/azukds/tubular/pull/766/changes/b7084185138a665c07bed6744c8468f620f22d93

tubular/imputers.py
liamholmes31: updated here https://github.com/azukds/tubular/pull/766/changes/57bc5f92a5d35a1776a15eeb8b40bb267cdcef9a

tubular/imputers.py
liamholmes31: ah thanks, have gone through all of these here https://github.com/azukds/tubular/pull/766/changes/0042ff6ccd0cdaf0ea12dc55a4a40b047e031e13

tubular/imputers.py
liamholmes31: thanks, fixed here https://github.com/azukds/tubular/pull/766/changes/0042ff6ccd0cdaf0ea12dc55a4a40b047e031e13

tubular/imputers.py
liamholmes31: good spot, fixed here https://github.com/azukds/tubular/pull/766/changes/b7084185138a665c07bed6744c8468f620f22d93#diff-49429640bd11a61fef0ea2bcb010bfb24ce43e4804c5b9d1f09f744c7e7ae48c

tubular/imputers.py
liamholmes31: I think this resolved when I merged in main :)

tubular/imputers.py
liamholmes31: hopefully in the future, have set up a spinout issue for this here https://github.com/azukds/tubular/issues/767

tldr; working with enums complicates our API, as you need to have knowledge of enum categories, so instead of, `self.get_transform_exprs()` you would need to do something like `self.get_transform_exprs(categories)`, which would break the format of our API

for the time being, I think users would need to either work with ArbitraryImputer, or convert to categorical type and work with the new CategoricalImputer

tests/conftest.py
liamholmes31: will pick this up in the other thread :)

tubular/imputers.py
lsumption: That makes sense, thanks for updating the docstring to cover this too!

lsumption (approved):

